PR #11, opened 2016-10-31
looping on rows columns functionality

NaySoftware:
I've added a looping functionality on rows and columns, this will ease when scanning over a table for ex.

let the code talk :+1: 

```go
package main

import (
	"fmt"
	"github.com/NaySoftware/excelize"
	"os"
)

func main() {
	xlsx, err := excelize.OpenFile("WorkBook.xlsx")
	if err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	rows, err := xlsx.GetRows("Sheet1")
	if err != nil {
		fmt.Println(err)
	}

	for _, row := range rows {
		for _, colCell := range row.C {
			val, _ := colCell.GetValueFrom(xlsx)
			fmt.Print(val, "\t\t")
		} // cols
		fmt.Println()
	} // rows

	fmt.Println()
} // main


```


Luxurioust:
Thanks for your PR. I found that some code will cause `golint` tips. I will update them later. I also found u have push some `.xlsx` files after tested, these excel file is useless and will decrease coverage. I will delete them.

Luxurioust:
Thanks for your PR. I found that some code will cause `golint` tips. I will update them later. I also found u have push some `.xlsx` files after tested, these excel file is useless and will decrease coverage. I will delete them.

NaySoftware:
I will update the code and make a new pull request

PR #15, opened 2017-01-11
Setter function to set raw values into a cell

Sharsie:
I did not want to meddle much with your code, but I recently needed to insert a float number into the sheet and wasn't able to use SetCellStr, because it escapes the value with an appostrophe. I needed to do some calculations with the floats

The PR adds a new function SetCellDefault which is exaclty what SetCellInt does, but accepts a string instead
It also modifies the behaviour of SetCellValue for float32 and float64 values and converts the float number to a string before calling SetCellDefault


Luxurioust:
A nice addition. Thanks Sharsie.

Sharsie:
Thank **you** for the library :)

PR #107, opened 2017-08-15
move 'format' to own package to allow struct usage and prevent name pollution

plandem:


plandem:
Travis will fail because of usage of "github.com/xuri/excelize/format" package.

xuri:
@plandem great work. I might to evaluate and fix failing test, and then consider whether to merge eventually.

PR #153, opened 2017-11-16
Rename import path to github.com/360EntSecGroup-Skylar/excelize

jinzhu:
Seems like we need to change import path to `github.com/360EntSecGroup-Skylar/excelize`?


javasgl:
import path `"github.com/360EntSecGroup-Skylar/excelize/v2"` was a little bit ugly.

SHOULD BE **lowercase-without-dash** witch is recommend by golang

PR #155, opened 2017-11-17
CopySheet() using reflect instead of encoding/gob

dolmen:
Use [github.com/mohae/deepcopy](https://godoc.org/github.com/mohae/deepcopy#Copy) (license MIT) to deep copy worksheets with package `reflect` instead of the internal [`deepcopy` function](https://github.com/360EntSecGroup-Skylar/excelize/blob/88e48e079a91191e05f32a232f8dec4454b25238/lib.go#L114L123) that was using `encoding/gob` serialization and
deserialization.

### DIFF
--- a/excelize_test.go
+++ b/excelize_test.go
@@ -88,7 +88,6 @@ func TestOpenFile(t *testing.T) {
 	xlsx.SetCellValue("Sheet2", "F16", true)
 	xlsx.SetCellValue("Sheet2", "F17", complex64(5+10i))
 	t.Log(letterOnlyMapF('x'))
-	t.Log(deepCopy(nil, nil))
 	shiftJulianToNoon(1, -0.6)
 	timeFromExcelTime(61, true)
 	timeFromExcelTime(62, true)
--- a/lib.go
+++ b/lib.go
@@ -3,7 +3,6 @@ package excelize
 import (
 	"archive/zip"
 	"bytes"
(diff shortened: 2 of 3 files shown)


dolmen:
Could this PR get a review?

xuri:
Sorry for my late reply, gob is based on reflect mechanism, I think it'is not necessary to introduce this third-party package to instead of gob.

dolmen:
1/ using `encoding/gob` is much slower than [`mohae/deepcopy`](https://github.com/mohae/deepcopy/)
2/ When building an application this implementation of `deepcopy` drags the `encoding/gob` package into the binary which is otherwise not linked. And this package is much bigger than `mohae/deepcopy` (which only depends on `time` and `reflect`).

```
$ LC_ALL=C stat -f "%6z %N" $(go env GOPATH)/pkg/$(go env GOOS)_$(go env GOARCH)/github.com/mohae/deepcopy.a $(go env GOROOT)/pkg/$(go env GOOS)_$(go env GOARCH)/encoding/gob.a
 10508 .../pkg/darwin_amd64/github.com/mohae/deepcopy.a
541818 .../pkg/darwin_amd64/encoding/gob.a
```

xuri:
Thanks for your pull request. As you said, I tested the deepcopy is 10x faster than gob.

dolmen:
Thanks for merging!

PR #156, opened 2017-11-17
SheetViewOptionPtr: document that it is a superset of SheetViewOption

dolmen:
Document in type system (not just in text for humans) that all SheetViewOptionPtr are also SheetViewOption (well, if not `nil`).
This improves documentation visible with godoc but this simple change also allows more flexibility to manipulate SheetViewOption programatically such as saving and restoring values as a `SheetViewOptionPtr` can now be used both with `SetSheetViewOption` as well as with `GetSheetView`:

```go
var opt excelize.ShowFormulas
opt = new(excelize.ShowFormulas)
_ = xl.GetSheetViewOptions(sheet, -1, opt)
_ = xl.SetSheetViewOptions(sheet, -1, opt)
```

### DIFF
--- a/sheetview.go
+++ b/sheetview.go
@@ -9,6 +9,7 @@ type SheetViewOption interface {
 
 // SheetViewOptionPtr is a writable SheetViewOption. See GetSheetViewOptions().
 type SheetViewOptionPtr interface {
+	SheetViewOption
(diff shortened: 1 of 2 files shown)


dolmen:
Thanks for merging! :fireworks:

Issue #182, opened 2018-01-25
Unable to write a bool type to a cell


pvelder:
When creating a sheet using MS Excel putting bool values in cells A1 and A2, we see following column values in the underlying sheet xml: 

    # for TRUE: 
    <row r="1"><c r="A1" s="7" t="b"><v>1</v></c></row>

    # for FALSE: 
    <row r="2"><c r="A2" s="7" t="b"><v>0</v></c></row>

When creating the sheet using excelize: 

    xlsx.SetCellValue("Sheet2", "A1", true)
    xlsx.SetCellValue("Sheet2", "A2", false)

we get strings, not boolean type values:

    <row r="1"><c r="A1" t="str"><v>true</v></c></row>
    <row r="2"><c r="A2" t="str"><v>false</v></c></row>

I will create a pull request today or tomorrow.

PR #183, opened 2018-01-25
xlsx.SetCellValue() now supports bool value

pvelder:
Here's a pull request addressing issue #182


pvelder:
No problem.  Thanks for this great library!

PR #184, opened 2018-01-29
Speedup setcellstyle more

martinal:
The recent improvement to `SetCellStyle` still loops over all the rows, this is more concise and faster.

Since the area check is now done by the loop, the inner loop check `checkCellInArea` is no longer needed.


xuri:
Thanks for your PR. Could you squash the commits and refine the commit message, you can copy the description of the PR to the squashed commit in this case.

martinal:
Squashed on top of latest master.

PR #196, opened 2018-03-15
Prevent corrupting the excel file by recreating the default sheet

cemremengu:


xuri:
Thanks for your PR.  I can't accept this patch, because the user may have changed the default worksheet name by the function [`SetSheetName()`](https://godoc.org/github.com/360EntSecGroup-Skylar/excelize#File.SetSheetName).

 I will add an exception return value such as `0` for `NewSheet()` when already exists a worksheet of the same name.

Issue #208, opened 2018-04-09
Excel files with inlineStr cell types read empty value


schbook:
[**Description**]

Excel files with inlineStr cell types read empty value, here is the sample file
[output.xlsx](https://github.com/360EntSecGroup-Skylar/excelize/files/1889512/output.xlsx)

**Steps to reproduce the issue:**
1.
download the sample file and read it with sample code as belong
2.
       xlsx, err := excelize.OpenFile("./output.xlsx")
	if err != nil {
		fmt.Println(err)
		return
	}
	rows := xlsx.GetRows("test")
	for _, row := range rows {
		for _, colCell := range row {
			fmt.Print(colCell, "\t")
		}
		fmt.Println()
	}

**Describe the results you received:**
			
1	2	3	
4	5	6	
7	8	9	

**Describe the results you expected:**

col1	col2	col3	
1	2	3	
4	5	6	
7	8	9		

**Output of `go version`:**

any

**Excelize version or commit ID:**

Excelize version: 1.2

**Environment details (OS, Microsoft Excel™ version, physical, etc.):**

any

xuri:
Thanks for your issue, I have added read `inlineStr` type cell value support. I also saw your pull request after I fixed it, thanks for you do that. I will close the PR.

schbook:
Thanks for your support, really fast and terrific.

PR #209, opened 2018-04-09
#208

schbook:
Support the "inlineStr" cell type


xuri (approved):
LGTM, I add unit test for formula calculation engine based on this branch.

Issue #203, opened 2018-03-29
Publicize style structs (formatFont, formatStyle and all internal structs separatly)


therox:
I think it'll be very comfortable and will minimize errors not to generate style json by hands, but (at least) marshal it for default struct.

xuri:
Thanks for your advice. I will make changes to assessment the APIs effect.

mlh758:
I would also like to see this change. I had to dig into the source code to figure out what style properties to set in order to align font for example. If the struct was pubic it would have been a lot easier to figure out and it would also help prevent typos.

trende-jp:
I also think this would be a big improvement.  There is nothing to lose by making it available and, in fact, I think it will make discovering how to interact with this library much simpler.

trende-jp:
https://github.com/360EntSecGroup-Skylar/excelize/pull/470

xuri:
I exported Style structs to allow create the style for cells by given JSON or structure. Ref: [9e2318c](https://github.com/360EntSecGroup-Skylar/excelize/commit/9e2318cefa4ebaa7bf6b1dbc95b30ad7a32366b1)

PR #219, opened 2018-05-06
Added helper functions to set the chart size.

jdevelop:
Just a nice addition to set the chart size upon inserting. 

It's rather too small and insignificant to create an enhancement issue for it first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ func main() {
     for k, v := range values {
         xlsx.SetCellValue("Sheet1", k, v)
     }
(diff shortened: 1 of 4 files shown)


xuri:
Thanks for your job. I think the size parameter should be in the style string, it is not recommended to add parameters for the `AddChart()` function.

xuri:
Size attributes are part of the style, the chart also contains many other attributes. If strip out the size parameters from the style parameters will bring a sense of split.

jdevelop:
Sounds like a good thing to have. Do you mind sharing which properties we can extract from the current style and use in the new way? I assume that it implies renaming of `chartOptions` in sort of `chartStyle` and providing more properties that could be extracted from the style itself.

jdevelop:
Moved the chart dimension into the format string, pls take a look.

xuri:
Thank you. Actually, I am considering to make style structs public, reference issue #203.

PR #233, opened 2018-06-11
Update documentation and add sample file

sewi-cpan:
Documentation:
* Add information about return value of NewSheet()
* Minor documentation language fixes
Samples:
* Added sample go file for dumping a XLSX file to the console


xuri:
Thanks for your pull request. I suggest that use [Godoc examples](https://blog.golang.org/examples) instead of the single sample file, that can be displayed as package documentation and that are verified by running them as tests. They can also be run by a user visiting the godoc web page for the package and clicking the associated "Run" button.

sewi-cpan:
@xuri: The sample requires an input XLSX file. That's why it can't be run on godoc.

PR #236, opened 2018-06-19
Add support for format 176 and fix excel time precision

lunny:


xuri:
Thanks for your PR. But the index number `176` is currently occupied by this lib's [built-in number format code](https://godoc.org/github.com/360EntSecGroup-Skylar/excelize#File.NewStyle): `$ English (U.S. Virgin Islands)`. Are you sure this number is a definite code in the [ECMA-376 specification](http://www.ecma-international.org/publications/standards/Ecma-376.htm)?

PR #237, opened 2018-06-19
Ability to parse dates further in future

liepumartins:
Golangs time.Duration uses nanoseconds, thus it is limited to approximately 290 years.


xuri:
Thanks for your pull request, could you add test cases for these changes?

xuri:
Could you squash the commits and refine the commit message, you can copy the description of the PR to the squashed commit in this case.

PR #267, opened 2018-09-01
Valiationdata

rentiansheng:
# PR Details

Support excel data data valiation 

## Description

You can use data validation to restrict the type of data or the values that users enter into a cell. One of the most common data validation uses is to create a drop-down list.
But  there is no way to get the format of the existing unit data validtion.
eg:
Digital range range error

image

Input prompt
image

Down list

image

## How Has This Been Tested

I wrote a unit test with restrictions on the range of numbers and implementation drop-down lists. Enter prompts and test tips

### DIFF
--- a/datavalidation.go
+++ b/datavalidation.go
@@ -1,0 +1,196 @@
+package excelize
+
+import (
+	"fmt"
+	"strings"
+)
+
+type DataValidationType int
+
+// Data validation types
+const (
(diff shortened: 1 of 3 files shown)


xuri:
Thanks for your PR. I have added `godoc` comments for the function, adjust `DataValidation` struct and fix `golint` errors based on your commits.

PR #275, opened 2018-10-08
Add Support for WriteTo Buffer

Zheaoli:
# PR Details

export a new function that people can get byte.buffer with all of file content

## Description

In some situation, we have to save the excel to a cloud storage service such as the Ali OSS. Some of the SDK that is supported by the cloud doesn't support write action. So we need to give it a reader that can read all of the excel file's content.

## Types of changes

## Checklist

### DIFF
--- a/excelize_test.go
+++ b/excelize_test.go
@@ -135,6 +135,10 @@ func TestOpenFile(t *testing.T) {
 	if err != nil {
 		t.Log(err)
 	}
+	_, err = xlsx.WriteToBuffer()
+	if err != nil {
+		t.Error(err)
+	}
 }
 
 func TestAddPicture(t *testing.T) {
--- a/file.go
+++ b/file.go
@@ -76,6 +76,15 @@ func (f *File) Write(w io.Writer) error {
 
 // WriteTo implements io.WriterTo to write the file.
 func (f *File) WriteTo(w io.Writer) (int64, error) {
+	buf, err := f.WriteToBuffer()
+	if err != nil {
+		return 0, err
+	}
+	return buf.WriteTo(w)
+}
+
+// WriteToBuffer provides a function to get bytes.Buffer from the saved file.
+func (f *File) WriteToBuffer() (*bytes.Buffer, error) {
 	buf := new(bytes.Buffer)
 	zw := zip.NewWriter(buf)
 	f.contentTypesWriter()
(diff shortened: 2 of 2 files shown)


Zheaoli:
@xuri already fix the review idea.

Here's another question: How to test std lib error such as `
```go
zw := zip.NewWriter(buf)
fi, err := zw.Create(path)
```

file.go
```diff
@@ -100,3 +97,33 @@ func (f *File) WriteTo(w io.Writer) (int64, error) {
...
+func (f *File) writeContent() {
+	f.contentTypesWriter()
+	f.workbookWriter()
+	f.workbookRelsWriter()
+	f.worksheetWriter()
+	f.styleSheetWriter()
+}
+
+// GenerateReader provides a function to generate a Reader reading from the file
+func (f *File) GenerateReader() (io.Reader, error) {
```
xuri: I suggest that create a new function `GetBuffer()` instead of `GenerateReader()` like:

```go
// WriteToBuffer provides a function to get bytes.Buffer from the saved file.
func (f *File) WriteToBuffer() (*bytes.Buffer, error) {
	buf := new(bytes.Buffer)
	zw := zip.NewWriter(buf)
	f.contentTypesWriter()
	f.workbookWriter()
	f.workbookRelsWriter()
	f.worksheetWriter()
	f.styleSheetWriter()
	for path, content := range f.XLSX {
		fi, err := zw.Create(path)
		if err != nil {
			return buf, err
		}
		_, err = fi.Write(content)
		if err != nil {
			return buf, err
		}
	}
	return buf, zw.Close()
}
```

and the user can get `io.Reader` like this:

```go
buf, _ := f.WriteToBuffer()
strings.NewReader(buf.String())
```

file.go
```diff
@@ -78,11 +79,7 @@ func (f *File) Write(w io.Writer) error {
 func (f *File) WriteTo(w io.Writer) (int64, error) {
 	buf := new(bytes.Buffer)
 	zw := zip.NewWriter(buf)
-	f.contentTypesWriter()
-	f.workbookWriter()
-	f.workbookRelsWriter()
-	f.worksheetWriter()
-	f.styleSheetWriter()
+	f.writeContent()
```
xuri: We can refactor the function `WriteTo()` like this:

```go
// WriteTo implements io.WriterTo to write the file.
func (f *File) WriteTo(w io.Writer) (int64, error) {
	buf, err := f.WriteToBuffer()
	if err != nil {
		return 0, err
	}
	return buf.WriteTo(w)
}
```

excelize_test.go
```diff
@@ -135,6 +135,10 @@ func TestOpenFile(t *testing.T) {
 	if err != nil {
 		t.Log(err)
 	}
+	_, err = xlsx.GenerateReader()
```
xuri: No additional unit testing is required if we have the function `WriteToBuffer`.

Zheaoli:
@xuri 
thx for review, that's a good idea.

xuri:
@Zheaoli Could you squash the commits and refine the commit message, you can copy the description of the PR to the squashed commit in this case.

Zheaoli:
@xuri down

PR #316, opened 2018-12-25
Add RegSearchSheet

r-uchino:
Add a regular expression to the search method

Since the return value is returned by MAP,
 it is convenient when you want to search by regular expression and obtain multiple cell coordinates

PS:
I am using translation, I'm sorry in dirty English :(

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -713,14 +713,13 @@ func (f *File) SearchSheet(sheet, value string) []string {
 // For a merged cell, get the coordinates
 // of the upper left corner of the merge area.
 // :example)
-// Search the coordinates where the numerical value in the range of "0 - 9" of Sheet 1 is described:
+// Search the coordinates where the numerical value in the range of "0-9" of Sheet 1 is described:
 //
 //    xlsx.RegSearchSheet("Sheet1", "[0-9]")
 //
-// return map[CELL:result CELL:result ...]
-func (f *File) RegSearchSheet(sheet, value string) map[string]string {
+func (f *File) RegSearchSheet(sheet, value string) []string {
 	xlsx := f.workSheetReader(sheet)
-	result := map[string]string{}
+	result := []string{}
 	name, ok := f.sheetMap[trimSheetName(sheet)]
 	if !ok {
 		return result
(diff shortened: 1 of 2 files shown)


xuri:
Hi @r-uchino, thanks for your PR. Please read [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) and add unit tests for your changes. I've left some comments.

sheet.go
```diff
@@ -707,6 +708,59 @@ func (f *File) SearchSheet(sheet, value string) []string {
...
+// with the given worksheet name and regular expression
+// For a merged cell, get the coordinates
+// of the upper left corner of the merge area.
+// :example)
+// Search the coordinates where the numerical value in the range of "0 - 9" of Sheet 1 is described:
+//
+//    xlsx.RegSearchSheet("Sheet1", "[0-9]")
+//
+// return map[CELL:result CELL:result ...]
+func (f *File) RegSearchSheet(sheet, value string) map[string]string {
```
xuri: I suggest that the return value of the function `RegSearchSheet` use type `[][]string` instead of `map[string]string` will be better, for example: `[][]string{"B3", "cell value"}`.

sheet.go
```diff
@@ -707,6 +708,59 @@ func (f *File) SearchSheet(sheet, value string) []string {
...
+		}
+		switch startElement := token.(type) {
+		case xml.StartElement:
+			inElement = startElement.Name.Local
+			if inElement == "row" {
+				r = xlsxRow{}
+				_ = decoder.DecodeElement(&r, &startElement)
+				for _, colCell := range r.C {
+					val, _ := colCell.getValueFrom(f, d)
+					_reg := regexp.MustCompile(value)
```
xuri: The contents of functions `SearchSheet` and `RegSearchSheet` are mostly the same, we can refactor function `SearchSheet` to add a variable length parameter `reg` instead of add the new function `RegSearchSheet`, and change return type of function `SearchSheet` to `[][]string`.

sheet.go
```diff
@@ -707,6 +708,59 @@ func (f *File) SearchSheet(sheet, value string) []string {
...
+					if !_reg.MatchString(val) {
+						continue
+					}
+					key := fmt.Sprintf("%s%d", strings.Map(letterOnlyMapF, colCell.R), r.R)
+					result[key] = _reg.FindString(val)
```
xuri: It'is not recommended to name variables with `_` at the start.

r-uchino:
# PR Details

`RegSearchSheet(sheet, value string)` method added

## Description

`RegSearchSheet(sheet, value string) ` Method added,It is similar to `SearchSheet(sheet, value string)`, but uses regular expressions for the search. Returns a cell that matches a regular expression


## Related Issue


## Motivation and Context

It is necessary to describe a certain tag in the Excel file and to identify and change the content from the program (for example, you can get the cell marked as {tag: thisPrice} Replace with variable values)
I implemented it for that

## How Has This Been Tested

https://github.com/360EntSecGroup-Skylar/excelize/pull/316/commits/6fba208b2cee4dcb3b3021f86255dc608e0d5f74

We added unit tests as well as existing methods

## Types of changes

## Checklist




r-uchino:
I did not grasp accurately, I am sorry, I sent the change as described above, I am pleased if you can confirm, Thank you for taking care of polite

Issue #357, opened 2019-03-18
Add SetCellFloat


mlh758:
**Description**

Floats are currently saved to the excel document as a string. Excel will show a warning for this and prompt users to convert to an actual number. It would be nice to be able to just write the float directly.

It looks like there would be two changes:

1. Create a `SetCellFloat` function
2. Use that function [here](https://github.com/360EntSecGroup-Skylar/excelize/blob/0072bb731043f89ce978778b9d7fdc6160e29de0/cell.go#L70)

**Steps to reproduce the issue:**
1. Use `SetCellValue` to write a float to the document
2. Open the document and observe the little warning triangle

![image](https://user-images.githubusercontent.com/4331955/54542908-9d8aef80-496a-11e9-9dc5-506c3d6b028a.png)

**Output of `go version`:**

```text
go version go1.12 darwin/amd64
```

I intend to submit a PR for this.

mlh758:
This isn't needed.

mlh758:
Actually this would still be helpful. Rounding to a certain decimal width is a pain in Go. Being able specify the number of places after the decimal with `SetCellFloat` while `SetCellValue` just defaults to `-1` in `FormatFloat` would be really useful.

PR #358, opened 2019-03-18
SetCellFloat to avoid numeric warnings in Excel

mlh758:
# PR Details

Excel will warn you if a string field holds a numeric value, this
allows the developer to avoid that warning.

Also removed duplicate type assertions in SetCellValue and
setCellIntValue.

## Related Issue

#357 

## Motivation and Context

My users were complaining about the little warning Excel gives if you put numeric values into a cell
as strings. This allowed me to properly set floating point values. Also, removing the duplicate type assertions should help performance.

## How Has This Been Tested

I added a unit test to make sure the value could still be retrieved using `GetCellValue`. I also integrated this change into my application using `gohack` and ensured it generated a valid xlsx document. I unzipped the document to make sure the xml looked like I would expect.

## Types of changes

## Checklist

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -70,23 +70,23 @@ func (f *File) mergeCellsParser(xlsx *xlsxWorksheet, axis string) string {
 func (f *File) SetCellValue(sheet, axis string, value interface{}) {
 	switch t := value.(type) {
 	case float32:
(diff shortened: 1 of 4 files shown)


mlh758:
Going to rebase off master.

mlh758:
I don't think this was actually necessary.

Issue #362, opened 2019-03-20
Set outline level range from startRowIndex & endRowIndex


suryateja619:
**Description**

Having an option for specifying the range of row numbers to be grouped in the same outline level would be a huge benefit. 

**Describe the results you received:**

For instance, if you have 10 rows in your excel and want to group the first five & remaining to the same level you would need to call the `GetRowOutlineLevel` for each row.

```go
xlsx.GetRowOutlineLevel("Sheet1" , 1, 1)
xlsx.GetRowOutlineLevel("Sheet1" , 2, 1)
...
```

**Describe the results you expected:**

We can create another function where we specify the start & end index numbers and it would set the outline range for us.

```go
xlsx.GetRowOutlineRangeLevel("Sheet1" , 1, 5, 1)
```

**Output of `go version`:**

```text
go version go1.11.5 darwin/amd64
```

PR #363, opened 2019-03-20
Added SetRowOutlineRangeLevel for range of rows

suryateja619:
# PR Details

## Description

Having an option for specifying the range of row numbers to be grouped in the same outline level.

## Related Issue

#362 

## Motivation and Context

It would allow users to specify the range of row numbers for the outline level to be applied instead of specifying it for every row.

## How Has This Been Tested

Tested it out locally.

## Types of changes

## Checklist


xuri:
Hi @suryateja619, thanks for your PR. But I don't think this is a very necessary method. In the design principle of the class library, there is only one way to implement a feature. So let the user handle this with simple functions.

Issue #359, opened 2019-03-19
Saving Duplicate Images


mlh758:
**Description**

When adding an image, it would be useful to detect if the image is a duplicate and instead of adding the image again, return a reference to the existing media.

We have a header that gets added in some large workbooks to each sheet, and this substantially increases the file size. If we open the file in Excel and save it again, Excel detects the duplicate images and consolidates them on save, reducing the file size.

I think it would be possible to change `addMedia` in `picture.go` to handle this behavior. `AddPictureFromBytes` would change slightly as well:

1. `addMedia` steps through all the saved media and looks for byte slices that have the same length as the file we are trying to save. Compute the hash for both (we can obviously re-use the hash of the file being saved for future comparisons)
2. If the hash matches, return the media path for the existing image. If no matches are found, save the new media and return its path
3. `AddPictureFromBytes` calls `addMedia` before calling `addDrawingRelationships` and uses the media path provided by `addMedia` in the call to `addDrawingRelationships`.

Since checking the length of a slice is a constant time operation, and very few slices should have the _exact_ same length without actually being the same media there shouldn't be much of a performance impact from the hashing. However, it would be worth adding a benchmark to ensure this doesn't cause a regression for #274. It would also be useful to incorporate a benchmark for actually saving the xlsx file since it's likely that the performance impact of this check would be offset by not having to write as many files.

I will wait for feedback on this one since while I think it may be useful but I understand not wanting to risk the performance impact.

xuri:
Thanks for the insight @mlh758. That's a useful feature. I'll certainly accept that patch if somebody did that. we can store the hash of the image in the `File` object for reducing the impact on performance.

mlh758:
A hash may actually be overkill, the bytes package has has an [Equal](https://godoc.org/bytes#Equal) function that already handles equality and would probably be more efficient.

PR #371, opened 2019-03-25
Do not save duplicate images

mlh758:
# PR Details

## Description

Adding the same image should create a drawing referencing the
already stored copy of the image.

## Related Issue

#359 

## Motivation and Context

This reduces the overall file size if an image is reused. For example, a header might be used frequently in a number of sheets and if the same logo is used this change will make it so only one copy of that logo is saved.

## How Has This Been Tested

Added a unit test, verified there was no performance regression in the benchmark, and generated
some documents with several repeated images.

## Types of changes

## Checklist


xuri:
LGTM, thanks @mlh758.

PR #391, opened 2019-04-25
Add the ability to change the default font

### LINKED ISSUE #390: Change Default Font
I would like to be able to change the default font of the document away from Calibri. This should just be a matter of changing the first font and setting the default named style to custom. I have a working prototype already so I'll get a PR up for this.

mlh758:
## Description

This allows users to change the default font of the workbook. I also removed the `scheme` stuff since the library doesn't have an API for themes in it yet and without that they don't appear to impact anything.

In the template I removed the theme reference from the default font and changed the default style name to `Normal` although I don't think that name change was necessary if you preferred the old name.

## Related Issue

Closes #390 

## Motivation and Context

My users want something other than Calibri so I figured I would add the change back to the main repo.

## How Has This Been Tested

I added some unit tests and generated some documents with this branch.

## Types of changes

## Checklist


mlh758:
The new line I added that I'm not hitting in the tests is the panic after the `xml.Unmarshal` in the new `readDefaultFont` function. That should never happen, but the panic is there to make it clear where this went wrong. I used panic instead of an error value because this is a truly exceptional case that should never be encountered. Even if you set invalid xml as the font like `<invalid>font<sjas` that will just end up being an invalid font in the workbook but it will save just fine.

mlh758:
Any feedback on this PR?

xuri:
Hi @mlh758, sorry for my late reply, the struct `font` has been removed for uniform, replace with `xlsxFont`. Could you update this patch? thank you.

mlh758:
Yes I will update that.

mlh758:
I rebased and cleaned it up. A lot of my change was no longer necessary since you refactored the xml.

mlh758:
I verified it worked still with a new default font and all the tests are passing. There were a fair amount of merge conflicts though so it might be good to try it out yourself as well.

xuri:
LGTM, thanks @mlh758.

Issue #439, opened 2019-07-19
I found excelize(golang) is 2~3 times slower than js-xlsx


xtutu:
Platform: Windows 10  
go version go1.12.4 windows/amd64  
node version v8.11.1  

-----

![image](https://user-images.githubusercontent.com/12950709/61506039-65c4af00-aa13-11e9-8fe8-f0caa606abba.png)

-------

Maybe there is some problem with my usage?

code is here:

```golang
package main
import (
	"fmt"
	"time"
    "github.com/360EntSecGroup-Skylar/excelize"
)
func main() {
	excelPath := "data.xlsx"

	timeStart := time.Now().UTC().UnixNano() / 1e6
	times := 5
	for i :=0 ; i < times; i++ {
		f, _ := excelize.OpenFile(excelPath)
		_, _ = f.GetRows("main")
	}

	timeFinished := time.Now().UTC().UnixNano() / 1e6
	fmt.Println(timeFinished - timeStart)
}
```

```js
import * as XLSX from 'xlsx';
function run() {
  let currentMilliSec = Date.now()
  let times = 5

  let excelPath = "data.xlsx"
  for (let i = 0; i < times; i++) {
    var workbook = XLSX.readFile(excelPath);
    let sheet = workbook.Sheets["main"];

    // let ref: string = sheet['!ref'];
    // console.log(ref);
    // let xx = ref.split(':')
    // console.log(xx[0], xx[1]);
    // let start = XLSX.utils.decode_cell(xx[0]);
    // console.log(start.c, start.r, XLSX.utils.encode_cell(start));
    // let end = XLSX.utils.decode_cell(xx[1]);
    // console.log(end.c, end.r, XLSX.utils.encode_cell(end));

    // let starRowIdx = 0
    // for (let r = 0; r <= end.r; r++) {
    //   let cellName = XLSX.utils.encode_cell({ c: 0, r: r });
    //   let cell = sheet[cellName]
    // }
  }

  let after = Date.now()
  console.log("js cost: ", after - currentMilliSec, "ms")
}

run()
```

mlh758:
Does `worksbook.Sheets["main"]` actually parse and read each row? `GetRows` does, it looks like you have the actual reading code commented out on the JS tests there.

xtutu:
> Does `worksbook.Sheets["main"]` actually parse and read each row?
Yes.

Actually 
```js
var workbook = XLSX.readFile(excelPath); //  this code is already finished parse.
[...]

PR #456, opened 2019-08-04
Fix 439

mlh758:
## Description

We were parsing the entire sheet twice which was leading to extra memory consumption and parsing time when reading an excel sheet with `Rows()` and `GetRows()`

## Related Issue

#439 

## Motivation and Context

This effectively halves the read time for spreadsheets.

## How Has This Been Tested

All tests continue to pass, and I added a benchmark to validate the performance improvement.
Before:
BenchmarkRows-8   	     500	   2434130 ns/op	  888380 B/op	    8612 allocs/op
OnlyOnce:
BenchmarkRows-8   	    1000	   2123277 ns/op	  811150 B/op	    7173 allocs/op

When run against the data sheet used in #439 the read time went from ~6.4 seconds on my machine to ~3.8 seconds.

## Important!

The one thing I did change is that if you declare a blank Rows struct and immediately call Columns on it you will get an index out of bounds error since you'll be trying to access nothing. However, this didn't seem realistic since users should be getting the rows struct by calling `file.Rows("someSheet")` and it works correctly under those conditions. I deleted the part of the test that looked for this behavior. I can add a bounds check to `Columns()` if you prefer but it seems unnecessary to me.

## Types of changes

## Checklist

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -10,10 +10,8 @@
 package excelize
 
 import (
-	"bytes"
 	"encoding/xml"
 	"fmt"
-	"io"
 	"math"
 	"strconv"
 )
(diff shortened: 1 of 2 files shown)


mlh758:
Squashed and re-signed.

mlh758:
Failing for code coverage on the patch, but this is mostly deleting code.

Also, is [this](https://github.com/360EntSecGroup-Skylar/excelize/blob/master/rows.go#L179) block necessary? It's replacing a namespace header for compatibility with Office 2007, but this is only a read operation within the library so Office 2007 isn't a concern at this step.

If I remove it, the tests all still pass, the test files I have still read properly, and it shaves another 30% off the read time of a file.

If other code in the library is relying on this side effect then I'll hold off on removing it for a later PR and spend some time investigating how to do it. If I can just delete it without any side effects then let me know and I'll remove it in this PR.

xuri:
Excellent!

PR #461, opened 2019-08-06
Further improve read performance

mlh758:
## Description

Instead of re-encoding the full sheet to change the namespaces
in the encoded bytes, read the sheet again and do the byte
replacements there.

In this case, file access ends up being more performant than
marshaling the sheet back to XML.

In the `SharedStrings` test, ensure the strings are actually read.

I actually think this entire operation is unnecessary since we shouldn't be modifying anything on a read but doing the change this way ensures my PR is fully passive in case anyone was relying on the behavior.

## Related Issue

 #439

## Motivation and Context

This further improves the read performance.

Before
BenchmarkRows-8   	    1000	   2177328 ns/op	  811432 B/op	    7173 allocs/op
After
BenchmarkRows-8   	    1000	   2035770 ns/op	  768284 B/op	    6883 allocs/op

On the test file I was using before this reduced the read time from ~3.8 seconds to ~2.5 seconds.

## How Has This Been Tested

I ran (and updated) all the unit tests and ran it against the same test files as before.

## Types of changes

## Checklist

### DIFF
--- a/excelize_test.go
+++ b/excelize_test.go
@@ -1003,7 +1003,11 @@ func TestSharedStrings(t *testing.T) {
 	if !assert.NoError(t, err) {
 		t.FailNow()
 	}
-	f.GetRows("Sheet1")
+	rows, err := f.GetRows("Sheet1")
+	if !assert.NoError(t, err) {
+		t.FailNow()
+	}
+	assert.Equal(t, "A", rows[0][0])
 }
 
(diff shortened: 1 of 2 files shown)


mlh758:
I think I was calling `readXML` incorrectly before, switched to using the name from `sheetMap` instead of the passed in sheet string directly.

Issue #462, opened 2019-08-07
Can't print cell contents


AngangGuo:
The following program can't print the contents of the last two cells:

go.mod
```
module report

go 1.13

require github.com/360EntSecGroup-Skylar/excelize/v2 v2.0.2-0.20190731155349-9279c86d85ab
```

```
package main

import (
	"fmt"
	"github.com/360EntSecGroup-Skylar/excelize/v2"
)

func main() {
	var fileName = "User Metrics.xlsx"
	var sheetName = "User Metrics"

	f, _ := excelize.OpenFile(fileName)
	contents, _ := f.GetRows(sheetName)

	for i,j:=range contents[2]{
		fmt.Println(i,j)
	}
}
```
Output:
```
0 Employee Name
1 Clean & Pack
2 Data Erasure
3 FG (Client)
4
5 FG(RL-LIQ)
6 Rebox
7 Repair Without Parts
8 Sort
9 Test
10
11
// missing the following contents:
// 10 Sort Weight (Kgs)
// 11 Test Weight (Kgs)
```

[User Metrics.zip](https://github.com/360EntSecGroup-Skylar/excelize/files/3474778/User.Metrics.zip)

mlh758:
Here is the XML for those cells:

```xml
<c s="3" t="inlineStr" r="K3">
  <is>
    <r>
      <rPr>
        <b></b>
        <i val="0"></i>
        <strike val="0"></strike>
        <u val="none"></u>
        <sz val="9"></sz>
        <color rgb="FFFFFFFF"></color>
        <rFont val="Tahoma"></rFont>
      </rPr>
      <t xml:space="preserve">Sort Weight (</t>
    </r>
    <r>
      <rPr>
        <b></b>
        <i val="0"></i>
        <strike val="0"></strike>
        <u val="none"></u>
        <sz val="9"></sz>
        <color rgb="FFFFFFFF"></color>
        <rFont val="Tahoma"></rFont>
      </rPr>
      <t xml:space="preserve">Kgs</t>
    </r>
    <r>
      <rPr>
        <b></b>
        <i val="0"></i>
        <strike val="0"></strike>
        <u val="none"></u>
        <sz val="9"></sz>
        <color rgb="FFFFFFFF"></color>
        <rFont val="Tahoma"></rFont>
      </rPr>
      <t xml:space="preserve">)</t>
    </r>
  </is>
</c>
```

[...]

PR #464, opened 2019-08-07
Handle multi row inline strings

mlh758:
This is based on #461 because I added to that unit test so either merge that one first and then this one, or just merge this one and close the other.

## Description

Handle multi row inline strings, simplify the row combination logic and use a
string builder.

## Related Issue

#462 

## Motivation and Context

Inline strings, not just shared strings, can have the `r` fields. Since the structs matched I went ahead and consolidated them.

## How Has This Been Tested

Added unit test and verified against the sample file provided in the issue.

## Types of changes

## Checklist


mlh758:
It looks like the tests go back to Go 1.9 and strings.Builder is unavailable. Do we still support Go 1.9? If so I can change that back to string concatenation.

xuri:
The new version can no longer support Go 1.9. This change will be included in the next version 2.0.2. I suggest simply break compatibility and update README notes.

mlh758:
Okay I updated the readme to note that Go 1.10 or higher is required.

PR #470, opened 2019-08-09
Exporting FormatStyle structs

trende-jp:
# PR Details

## Description

Exporting FormatStyles.

## Motivation and Context

Much friendlier for any user of the library.

## How Has This Been Tested

Using existing tests.

## Types of changes

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/360EntSecGroup-Skylar/excelize/v2
+module github.com/trende-jp/excelize
 
 go 1.12
 
--- a/styles.go
+++ b/styles.go
@@ -1885,7 +1885,15 @@ func parseFormatStyleSet(style string) (*FormatStyle, error) {
 //
 // Cell Sheet1!A6 in the Excel Application: martes, 04 de Julio de 2017
 //
-func (f *File) NewStyle(fs *FormatStyle) (int, error) {
+func (f *File) NewStyle(style string) (int, error) {
+	fs := &FormatStyle{}
+	if err := json.Unmarshal([]byte(style), fs); err != nil {
+		return 0, err
+	}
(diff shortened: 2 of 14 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/360EntSecGroup-Skylar/excelize/v2
```
mlh758: When you are developing locally with go modules you can use `replace` to change where the code is pulled from to test your local changes instead of renaming the module.

styles.go
```diff
@@ -1885,13 +1885,9 @@ func parseFormatStyleSet(style string) (*formatStyle, error) {
 //
 // Cell Sheet1!A6 in the Excel Application: martes, 04 de Julio de 2017
 //
-func (f *File) NewStyle(style string) (int, error) {
+func (f *File) NewStyle(fs *FormatStyle) (int, error) {
```
mlh758: I think you should add a new function that takes the newly exposed style struct instead of changing the signature of NewStyle.

As it stands, we would have to release v3 of the library since you are making a breaking change to how people would consume it. If you add a new function to take the struct like `NewStyleStruct` or `AddStyle` (or something...) then this can go out as v2.1.0.

All the private functions you changed seem fine since consumers can't see those.

mlh758:
You could use [iota](https://github.com/golang/go/wiki/Iota) to give names to the numbers used in border styles and such which would further help users consume this.

trende-jp:
Reverted to something api compatible.  As for iota, I will leave that alone for now.

trende-jp:
Is it possible to get this approved?  I am running into performance issues, but my code is based on the templates being passed objects rather than strings.

xuri:
I exported Style structs to allow create the style for cells by given JSON or structure.

Issue #485, opened 2019-09-18
 Call GetSheetName can not get correct sheetname


virteman:
**Description**

GetSheetName can not get correct sheet name just return empty string

**Steps to reproduce the issue:**
1. my xlsx file's first sheet named "basic". so I call GetSheetIndex("basic") get return value is 15
2. call GetSheetName(15) return value is "" (empty string)

**Describe the results you received:**
I get an empty string as sheetname

**Describe the results you expected:**
I expect the return string should be "basic"

**Output of `go version`:**

```text
go version go1.11 linux/amd64
```

**Excelize version or commit ID:**

```text
eef232f09ecd41b1f8fc199906ce0be64865802e
```

**Environment details (OS, Microsoft Excel™ version, physical, etc.):**
```text
Ubuntu 18.04, wps 2019( can open and eidt excel 2007 xlsx file), thinkpad x1 carbon
```

xuri:
Hi @virteman, thanks for your feedback. Could you provide testing file attachment?

mlh758:
This is likely related to the changes I made [here](https://github.com/360EntSecGroup-Skylar/excelize/commit/faaaa52cb862499454a7f893b92e8430d00172a5) since the old method of using SheetID was causing the bug in #457 

It looks like we need to update [GetSheetIndex](https://github.com/360EntSecGroup-Skylar/excelize/blob/master/sheet.go#L298) to actually return the index rather than the SheetID since SheetID isn't a reliable index.

mariaefi29:
Dear developers,

Do you have any updates on the issue? 

I have just moved to your package from another xlsx parser and when started testing on real data, it turned out it doesn't work for a lot of files.

For instance, I have a file with the only one sheet. I would like to get that sheet and I use file.GetSheetName(1), assuming that the sheet's name will return to me. Unfortunately, it returns an empty string. I dig deeper and I realised that the only one sheet I have in that file has index 2 (it's SheetID is 2). That's weird. As you said, for some reason it seems that SheetID isn't reliable index, but it has 'rId1' real id.

Could you please clarify that and/or give us some ETA on the fix? I used the latest version v2.1.0.

You can get the file here (I filled it up with the fake data).
[file to download](https://drive.google.com/file/d/1oJBcC7JPgWc_KWwmJR8LEn3_DIo_CoGt/view?usp=sharing)

We would be grateful you could look into that.

xuri:
[...]

PR #463, opened 2019-08-07
Get sheet names based on index

mlh758:
## Description

SheetID only seems to indicate the file name for the sheet.
Check the sheets list based on index instead. Reordering sheets in Excel changes the order they appear in that list.

## Related Issue

Fixes #457

## Motivation and Context

Using SheetID is not a reliable way to get a sheet's name based on index.

## How Has This Been Tested

Verified with the test file given in #457, and against the existing files in the test directory of this project. I added unit tests to ensure the expected matches.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #457, opened 2019-08-06
GetSheetName doesn't work on this excel


GZLHZ:
**Description**

GetSheetName doesn't work on this excel  (other excels are no problem)
[GetSheetName.xlsx](https://github.com/360EntSecGroup-Skylar/excelize/files/3469994/GetSheetName.xlsx)

```jsx
func ReadExcelFile(){
	f, err := excelize.OpenFile("./GetSheetName.xlsx")
	if err != nil {
		Println(err)
		return
	}

	fmt.Println(f.GetSheetName(1))
}
```

**Steps to reproduce the issue:**
1.Open the excel and read it and you can find it can't get the result "Sheet1",it got a empty string

**Output of `go version`:**
1.12.7

**Excelize version or commit ID:**
2019.08.04 update github

**Environment details (OS, Microsoft Excel™ version, physical, etc.):**
win10 1903
office 2019(but i don't know who create this excel and his excel version, it's on our svn)

mlh758:
This opens fine for me on MacOS. `"./GetSheetName.xlsx"` doesn't look like valid path syntax on Windows.

GZLHZ:
you can test it on windows, other excels are no problem, but i don't know why this excel cannot get the sheet name.

GZLHZ:
> This opens fine for me on MacOS. "./GetSheetName.xlsx" doesn't look like valid path syntax on Windows.
you can test it on windows, other excels are no problem, but i don't know why this excel cannot get the sheet name.

mlh758:
Okay yeah I can reproduce this. `GetSheetName` looks for a sheet with an ID that matches the index. This presumably assumes that the IDs on the sheets are in order. The only sheet in this excel document has an ID of 5.

mlh758:
It looks like changing the sheet order changes the order sheets appear within the `<sheets>` tag within `workbook.xml`. @xuri I think `GetSheetName` can be simplified to just check the index of the sheet against the `Sheets` field in a workbook.

mlh758:
Go ahead and pull down the code from [this PR](https://github.com/360EntSecGroup-Skylar/excelize/pull/463) and let me know if that fixes your issue.

GZLHZ:
@mlh758  yes, it can slove the problem.thank you.

PR #486, opened 2019-09-18
fix sheepID not be related with sheet array's index

virteman:
# PR Details
Make the sheepID to be related with sheet array's index

## Description
When some excel sheep's sheepID maybe any number instead of start 1. for example I have a xslx file only one sheet, I GetSheetIndex('firstsheet') is 15,  then I call GetSheetName(15) cannot get any sheet name just a empty string return.

## Related Issue
 #485 

## Motivation and Context

even if the sheetID not start 1, GetSheetName can get the correct name.

## How Has This Been Tested

My excel sheet copy operator can correct call, and the result is what I expect.

## Types of changes

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -85,6 +85,10 @@ func (f *File) workbookReader() *xlsxWorkbook {
 		var content xlsxWorkbook
 		_ = xml.Unmarshal(namespaceStrictToTransitional(f.readXML("xl/workbook.xml")), &content)
 		f.WorkBook = &content
+		//rebuild sheetID
+		for i, _ := range f.WorkBook.Sheets.Sheet {
+			f.WorkBook.Sheets.Sheet[i].SheetID = i + 1
+		}
 	}
 	return f.WorkBook
 }


sheet.go
```diff
@@ -85,6 +85,10 @@ func (f *File) workbookReader() *xlsxWorkbook {
 		var content xlsxWorkbook
 		_ = xml.Unmarshal(namespaceStrictToTransitional(f.readXML("xl/workbook.xml")), &content)
 		f.WorkBook = &content
+		//rebuild sheetID
```
xuri: If the ID associated with it is not updated together，this modification will lead to the corruption of the spreadsheet in some cases. Ref. [ECMA-376 5th edition Part 1 §18.11.1.23](https://www.ecma-international.org/publications/standards/Ecma-376.htm)

xuri:
Hi @virteman, thanks for your PR, I have left some comments.

mlh758:
Is GetSheetIndex returning the wrong value? The code [here](https://github.com/360EntSecGroup-Skylar/excelize/blob/e7581ebf3e14f096b6e2d56ed34d381b4af6d310/sheet.go#L303) appears to be returning the SheetID instead of the index so it's definitely possible for them to get out of sync.

#463 and #457 were related to using ID instead of index in other places.

xuri:
Hi @virteman, I fixed workbook's rels ID calc error in the commit [4e4a5b9](https://github.com/360EntSecGroup-Skylar/excelize/commit/4e4a5b9b3e052d1694442515492792fb1aa74c5a#diff-ec4216644fd4919ab5c2a68efb70d453R209-R216), this fix may be work for your case, I will close this PR, any feedback is welcome.

mlh758:
Does that change still work if a sheet is deleted? That seems to basically revert #463

xuri:
I have tested with a lot of files to ensure its correctness, any feedback on special circumstances is welcome.

mlh758:
https://gist.github.com/mlh758/b486e4fe39badf5b2becf59062a00b82

I ran that against master and the output is:
```
1 => First Sheet
2 => Second Sheet
3 =>
Index: 1, name: First Sheet
Index: 2, name: Second Sheet
Index: 0, name:
```

Against version 2.0.2 the output is:
```
1 => First Sheet
2 => Second Sheet
3 => Fourth Sheet
Index: 1, name: First Sheet
Index: 2, name: Second Sheet
Index: 4, name: Fourth Sheet
```
which is more correct but that's what I was getting at with my comment earlier that GetSheetIndex is returning the wrong value.

xuri:
Use `GetSheetMap` instead of handle sheet index manuly:
```go
for idx, name := range xlsx.GetSheetMap() {
    println("Index:", idx, "GetSheetIndex:", xlsx.GetSheetIndex(name), "name:", name)
}
```

mlh758:
The sheet index would be the order of the sheets:
`[Sheet1, Sheet2, Sheet3]`

This library mostly uses a 1 based index since Excel does the same so a user of the library would expect those to be at indexes 1, 2, and 3 respectively. These are unrelated to the SheetID attributes.

If I had created 5 sheets programmatically, with some deleted and added along the way and I then called `GetSheetName(4)` I would expect the name of the 4th sheet, not a blank reply.

If you no longer intend to support `GetSheetName` and `GetSheetIndex` then release a v3 and make those functions private.

olegkrutikov:
Here is what I had to do to be able to parse my file which only consists of one sheet which has (gets from f.workbookReader()?) SheetID of 2 for some reason:

```
func (f *File) GetSheetName(index int) string {
	wb := f.workbookReader()
	if wb == nil || index < 1 {
		return ""
	}
	for i, sheet := range wb.Sheets.Sheet {
	//	if index == sheet.SheetID {
		if index == i+1 {
			return sheet.Name
		}
	}
	return ""
}
```

And here is what is being parsed (debugger info)

```
Sheets = {github.com/360EntSecGroup-Skylar/excelize.xlsxSheets} 
 Sheet = {[]github.com/360EntSecGroup-Skylar/excelize.xlsxSheet} len:1, cap:1
  0 = {github.com/360EntSecGroup-Skylar/excelize.xlsxSheet} 
   Name = {string} "Прайс"
   SheetID = {int} 2
   ID = {string} "rId1"
   State = {string} ""
```

mariaefi29:
@olegkrutikov  hi!

I have the same problem :(

Issue #487, opened 2019-09-24
Improve write performance


ducquangkstn:
Currently, excelize writing approach is:
- marshal obj with format xml to []byte (XLSX field)
- write XLSX to bytes.Buffer
- write buffer to file
However, this approach will consume a lot of memory when the amount of data is large.
I suggest writing to file as stream with xml.Decoder
I would like to create a PR if u think this proposal is necessary

xuri:
Hi @ducquangkstn, welcome to contribute. Streaming read and write support will be an excellent feature, we need to the API design first. This is some of my thoughts on [performance improvement](https://github.com/360EntSecGroup-Skylar/excelize/wiki#optimization).

ducquangkstn:
Some sample of my change would be:
`
func (f File) calcChainWriter(zw *zip.Writer) error {
	if f.CalcChain != nil && f.CalcChain.C != nil {
		return writeXML(zw, "xl/calcChain.xml", f.CalcChain)
	}
	return nil
}

func writeXML(zw *zip.Writer, name string, data interface{}) error {
	w, err := zw.Create(name)
	if err != nil {
		return err
	}
	_, err = w.Write([]byte(XMLHeader))
	if err != nil {
		return err
	}
	encoder := xml.NewEncoder(w)
	err = encoder.Encode(data)
	if err != nil {
		return err
	}
	return nil
}
`
However, there are some **hack code** like `replaceRelationshipsBytes(replaceRelationshipsNameSpaceBytes(output)))`
I would replace them by custom marshalXML function
What do you think

xuri:
Two years passed, the streaming writing API has been added, the performance has been an improvement, I'll close this issue, but still, welcome any optimization or suggestions for the performance.

PR #489, opened 2019-09-27
fix-save-to-file-as-stream

ducquangkstn:
# PR Details
- Change the write method from writing to a bytes.Buffer to writing a stream to io.Writer

## Description
- create a zip.Writer from io.Writer
- for each components, create a new file then using a xml.Encoder() and write to io.Writer
- using custom unmarshalXML to avoid using this block of code
```

func replaceRelationshipsBytes(content []byte) []byte {
	oldXmlns := []byte(`xmlns:relationships="http://schemas.openxmlformats.org/officeDocument/2006/relationships" relationships`)
	newXmlns := []byte("r")
	return bytes.Replace(content, oldXmlns, newXmlns, -1)
}
```

## Related Issue
https://github.com/360EntSecGroup-Skylar/excelize/issues/487

## Motivation and Context
improve performance and reduce memory for storing XML data

## How Has This Been Tested
- writing a unit test for comparing custom unmarshal function and bytes.Replace result

## Types of changes

- [] Bug fix (non-breaking change which fixes an issue)

## Checklist


xmlWorksheet.go
```diff
@@ -43,11 +45,114 @@ type xlsxWorksheet struct {
 	ExtLst                *xlsxExtLst                  `xml:"extLst"`
 }
 
+// MarshalXML implements xml.Marshaler
```
xuri: Invalid attribute and incorrect namespace order in workbooks, worksheets, and style, which will cause the file to be corrupted when creating a new file

ducquangkstn: - these attributes and namespace are replace this function replaceRelationshipsBytes(replaceWorkSheetsRelationshipsNameSpaceBytes(output))
in https://github.com/360EntSecGroup-Skylar/excelize/blob/master/sheet.go#L110
- I test this code by create a new file then calculate the check sum. Using two branch mine and master get the same results, which means the output files are identical

ducquangkstn: also this is a better way to replace the function `replaceRelationshipsBytes`

ducquangkstn: @xuri pls re-review the PR or give me more comments

xuri: A lot of code in this PR, I need some time to review.

file.go
```diff
@@ -76,44 +75,64 @@ func (f *File) SaveAs(name string) error {
...
 		}
 		_, err = fi.Write(content)
 		if err != nil {
-			zw.Close()
-			return buf, err
+			return err
 		}
 	}
-	return buf, zw.Close()
+	err = f.calcChainWriter(zw)
```
mlh758: Since these are all functions on File, you could either:

1. Put a `writeError` field or something on File that these functions check before doing anything so that we can just call all these functions in sequence and then return `writeError` at the end.
2. Collapse these into the shorter syntax like:
```
if err := f.commentsWriter(zw); err != nil {
  return err
}
```
like you did above in some of the other functions

xmlStyles.go
```diff
@@ -29,6 +29,67 @@ type xlsxStyleSheet struct {
 	ExtLst       *xlsxExtLst       `xml:"extLst"`
 }
 
+// MarshalXML implements xml.Marshaler
```
mlh758: A more thorough comment here explaining why you are copying a bunch of fields to an anonymous struct would help future contributors. I assume this is to get rid of the byte replacements?

xmlStyles_test.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func TestXlsxStyleSheet_MarshalXML(t *testing.T) {
+	relationships := xlsxStyleSheet{}
+
+	b, err := xml.Marshal(relationships)
+	require.NoError(t, err)
+
+	b2 := replaceStyleRelationshipsNameSpaceBytes(b)
+	t.Log(string(b))
```
mlh758: Do you still need the logs here?

lib.go
```diff
@@ -52,6 +53,30 @@ func (f *File) saveFileList(name string, content []byte) {
...
+	}
+	_, err = w.Write([]byte(XMLHeader))
+	if err != nil {
+		return err
+	}
+	encoder := xml.NewEncoder(w)
+	return encoder.Encode(data)
+}
+
+// writeStringToZipWriter writes string to zip,Writer
```
mlh758: Extra comma in the comment here

file.go
```diff
@@ -76,44 +75,64 @@ func (f *File) SaveAs(name string) error {
 
 // Write provides a function to write to an io.Writer.
 func (f *File) Write(w io.Writer) error {
-	_, err := f.WriteTo(w)
-	return err
-}
-
-// WriteTo implements io.WriterTo to write the file.
-func (f *File) WriteTo(w io.Writer) (int64, error) {
```
mlh758: I don't see anything replacing the lost `WriteTo` method. Does this mean that excel.File no longer satisfies io.WriteTo?

xmlWorkbook.go
```diff
@@ -44,6 +44,70 @@ type xlsxWorkbook struct {
 	FileRecoveryPr      *xlsxFileRecoveryPr      `xml:"fileRecoveryPr"`
 }
 
+// MarshalXML implements xml.Marshaler
+func (x xlsxWorkbook) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
+	x2 := struct {
```
mlh758: Would it be too costly to add these fields used in the custom marshal functions to the normal structs we already have and then populate the extra fields during a write? It seems like this exposes us to potential bugs where we might add other fields to things like `xlsxWorkbook` and forget to add them to these structs here and be very confused as to why they aren't appearing in the final output.

xmlWorkbook.go
```diff
@@ -155,6 +219,25 @@ type xlsxSheet struct {
 	State   string `xml:"state,attr,omitempty"`
 }
 
+// tmpXlsxSheet is struct for custom marshal xlsxSheet
+type tmpXlsxSheet struct {
```
mlh758: You used anonymous structs inside the function in other places but not here. It would be good to stay consistent.

xmlWorkbook.go
```diff
@@ -179,6 +262,21 @@ type xlsxPivotCache struct {
...
+// tmpXlsxPivotCache is struct for custom marshal xlsxPivotCache
+type tmpXlsxPivotCache struct {
+	XMLName xml.Name `xml:"pivotCache"`
+	RID     string   `xml:"r:id,attr,omitempty"`
+	CacheID int      `xml:"cacheId,attr"`
+}
+
+// MarshalXML implements xml.Marshaler
+func (x *xlsxPivotCache) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
+	return e.Encode(tmpXlsxPivotCache{
```
mlh758: Would this also need to copy `XMLName`?

mlh758:
Since this is primarily a performance PR, it would be helpful if you included some benchmarks in the test files to show the actual difference and prevent future regressions as changes are made.

file.go
ducquangkstn: WriteTo returns int64 and err. 
In prev version, excelize write from bytes buffer to a file, so File object can get the length of total bytes writen `https://golang.org/pkg/bytes/#Buffer.WriteTo`
Currently, this PR is using zip.Writer so this method is nolonger support.

mlh758: The problem is that `WriteTo` is part of the public interface of this library, we can't remove it without releasing this under version 3, and I don't really see a good reason to remove the function. For the sake of this PR, you could even use the old behavior of copying it through a buffer. It would be less efficient due to the copy but a comment saying to favor using Write until WriteTo can be improved would be enough for consumers of the library to know how to make an efficient choice without breaking existing code.

mlh758:
I created a benchmark [here](https://gist.github.com/mlh758/75a569faa798339b40859a3f594780b8) and ran it against master and this branch.

Master
BenchmarkWrite-8   	       2	 574756323 ns/op	207278252 B/op	 1893776 allocs/op

PR
BenchmarkWrite-8   	       2	 548271059 ns/op	164345284 B/op	 1893817 allocs/op

I also ran it against my branch [here](https://github.com/mlh758/excelize/tree/fix-494-write-allocations) that I brought up in #494 which has a narrower scope of changes.

BenchmarkWrite-8   	       2	 517713714 ns/op	127115252 B/op	 1833746 allocs/op

We can probably combine both changes but there is clearly some benefit to targeting the most allocation heavy functions when going after memory issues.

mlh758:
Would we achieve even more memory gains by replacing the File.XLSX map of `map[string][]byte` with a struct that wraps a zip file? Since it implements logic for reading and writing we could pass files in and out of it and let the zip object inside re-compress them when we're done with the files.

The struct could provide a similar interface to a map making the updates easier throughout the library. That should avoid the current issue of essentially storing an additional copy of the file in memory.

We could even provide some configuration options to flush the file contents to a temp file on disk while we are working with the file so we don't have to hold that in memory the whole time either.

ducquangkstn:
@mlh758 Currently, excelize is storing data at both map[string][]byte and zip.Writer and zip.Writer is built-in lib and optimize (I think)
I have test the memory using this code
```
func TestFile_Save(t *testing.T) {
	//t.Skip()
	done := make(chan struct{})
	go func() {
		for {
			select {
			case <-done:
				return
			case <-time.Tick(time.Millisecond * 500):
				PrintMemUsage()
			}
		}
	}()

	f := excelize.NewFile()
	for s := 0; s < 5; s++ {
		sheetName := fmt.Sprint("sheet", s+1)
		f.NewSheet(sheetName)
		for _, col := range []string{"A", "B", "C", "D"} {
			for i := 0; i < 100000; i++ {
				require.NoError(t, f.SetCellValue(sheetName, fmt.Sprint(col, i+1), i))
			}
		}

	}
	require.NoError(t, f.SaveAs("test.xlsx"))
	PrintMemUsage()

	done <- struct{}{}
}

// PrintMemUsage outputs the current, total and OS memory being used. As well as the number
// of garage collection cycles completed.
func PrintMemUsage() {
	var m runtime.MemStats
	runtime.ReadMemStats(&m)
	// For info on each, see: https://golang.org/pkg/runtime/#MemStats
	fmt.Printf("Alloc = %v MiB", bToMb(m.Alloc))
	fmt.Printf("\tTotalAlloc = %v MiB", bToMb(m.TotalAlloc))
	fmt.Printf("\tSys = %v MiB", bToMb(m.Sys))
	fmt.Printf("\tNumGC = %v\n", m.NumGC)
}

func bToMb(b uint64) uint64 {
	return b / 1024 / 1024
}
```

the result is the PR can save about 15% memory.

mlh758:
You can use Go's built in tools for profiling and memory statistics:

`go test -benchmem -run='github.com/360EntSecGroup-Skylar/excelize/v2' -bench BenchmarkWrite -memprofile memprofile.out`

`go tool pprof -nodefraction=0.1 -png memprofile.out > prchange.png`

That runs the benchmark I posted in my [comment above](https://gist.github.com/mlh758/75a569faa798339b40859a3f594780b8) and profiles the memory usage. The pprof command trims out smaller functions to clean up the visual and creates an image showing consumption. Here is what your PR looks like, if you're curious:

prchange

What I was getting at in my comment is that most of the memory consumed is in smaller operations repeated many times that perform excessive allocations. The copying at write is definitely a good target for optimization, and I like the idea of getting rid of the `replaceRelationshipsNameSpaceBytes` type functions because that will let us stream files more effectively, but when you're going after memory issues it's a good idea to profile first and then go after the low hanging fruit. As it stands right now this PR reduces total memory consumed but increases allocations and CPU time. I suspect some of that is due to the anonymous structs and having to copy data across to them.

Could you add the xml fields to the base structs we already have, and change the fields before we write instead? That would keep us from having to copy, and also likely keep us from needing custom Marshal functions which I suspect is the big culprit of the slow down. Custom marshal functions can [have some surprising side effects](https://philpearl.github.io/post/badgo_json/).

mlh758:
Also I've been tinkering with trying to use _just_ a zip archive for storage of serialized files for the lifetime of the excel object but I'm running into a lot of issues with modifying the archive. You can write the same file multiple times into a zip archive and removing a file means you have to copy the whole archive except for that one file.

Checking for the existence of a file being added first (and avoiding the copy most of the time a file is written to the archive) leads to some solid performance gains but it is also leading to some subtle bugs where the write buffer is not a valid zip archive and can't be read.

If I could figure out the write buffer issue it would be nice to combine it with the work you're doing in this PR to pass everything around as streams all the time.

lib.go
chaudharisuresh997: checking grammer in comment :p

ducquangkstn:
@mlh758 : Recently, I found the way to pass any default xml to a struct like `xmlWorkbook`
https://github.com/360EntSecGroup-Skylar/excelize/pull/489/commits/5dc36b72b4b92881a2d2ee4459c670cd40984f8f#diff-50b2cf86b4bb003458a82d3e89d0036a
PTAL

lib.go
mlh758: I know it's a little trivial but clean comments help everyone who uses the code

ducquangkstn: comment updated

jimsmart:
What is the status of this please? 

We have quite an interest in specifically reducing the maximum total memory consumed when saving a file. 

We would prefer overall memory saving over speed gains: our app runs on a server, where, for us, the tasks are not time critical but overall memory usage is.

xuri:
Hi @jimsmart, I have added a [stream writer](https://pkg.go.dev/github.com/360EntSecGroup-Skylar/excelize/v2?tab=doc#File.NewStreamWriter) for generating a new worksheet with huge amounts of data. This PR contains a lot of code and I need some time to review.

jimsmart:
Ok. Thanks for the info.

ducquangkstn:
@xuri: There are other methods to improve performance like change from string field to Stringer field
https://godoc.org/golang.org/x/tools/cmd/stringer  like xlsxC.T and xlsxC.V.
But the change might be big. Not sure I should create a PR

PR #530, opened 2019-12-10
Stream to Excel table

chowey:
# PR Details

This slightly changes and enhances the StreamWriter API.

## Description

This PR handles the following points:

* Use any data type in the StreamWriter API while writing new rows
* Provide styles for values in the StreamWriter API
* Stream to an Excel table with the StreamWriter API

## Related Issue

#523

## Motivation and Context

This morning I ran into an issue where very large Excel files were causing `excelize` to use all of my computer's RAM. I checked if there were any streaming APIs and sure enough, the StreamWriter API was merged this morning.

This handled my RAM issue but the new StreamWriter API has the above deficiencies.

I realize I am slightly changing the StreamWriter API, but I think it is still early enough to do this.

Now I can efficiently create large Excel tables with 160 columns and 200000 rows, including dates and formats, using less than 200MB of RAM. Before it was taking more than 20GB of RAM and then crashing.

## How Has This Been Tested

I ran the automated tests and tested against some production use cases.

## Types of changes

## Checklist


xuri:
Hi @chowey, thanks for your PR. This PR contains a lot of code and I need some time to review.

chowey:
No worries, I understand.

ys-d:
use  StreamWriter api  ，how time published？

xuri:
> use StreamWriter api ，what time published？ @xuri

Stream writing will be in the next [release v2.0.3](https://github.com/360EntSecGroup-Skylar/excelize/projects/5), but i'm not sure if include this PR code.

chowey:
Is there something I can do that will help with the review?

This API is slightly incompatible with 08d1a86c3a1bffdf431dba6a3d5a3b369ef740a7, since I allow setting styles while adding rows. I would like to fix it before v2.0.3 so there is no breaking API changes moving forward.

stream.go
```diff
@@ -66,45 +72,182 @@ func (f *File) NewStreamWriter(sheet string) (*StreamWriter, error) {
...
 	rsw.SheetData.WriteString("<sheetData>")
 	return rsw, nil
 }
 
-// SetRow writes an array to streaming row by given worksheet name, starting
-// coordinate and a pointer to array type 'slice'. Note that, cell settings
-// with styles are not supported currently and after set rows, you must call the
-// 'Flush' method to end the streaming writing process. The following
-// shows the supported data types:
+// SetTable creates an Excel table for the StreamWriter. This function must be
```
xuri: Please add the using example in the doc.

stream.go
```diff
@@ -66,45 +72,182 @@ func (f *File) NewStreamWriter(sheet string) (*StreamWriter, error) {
...
+// called before the first SetRow.
 //
-//    int
-//    string
-//
-func (sw *StreamWriter) SetRow(axis string, slice interface{}) error {
+// The first SetRow will create the column names. These Column names must be
+// strings or an error will be returned.
+func (sw *StreamWriter) SetTable(format string) error {
```
xuri: I propose to change the function name to `AddTable`. Also, It seems that there is no table range in the function parameters. Table range should be similar to [`AddTable`](https://github.com/360EntSecGroup-Skylar/excelize/blob/7358dca436f6ca5948a3f2865b14e828863d86a9/table.go#L53)

stream.go
```diff
@@ -66,45 +72,182 @@ func (f *File) NewStreamWriter(sheet string) (*StreamWriter, error) {
...
+	}
+	return nil
+}
+
+// SetRow writes an array to stream rows by giving a worksheet name, starting
+// coordinate and a pointer to an array of values. If styles is non-nil, then
+// the styles must be the same size as the values and will be applied to each
+// corresponding cell. Note that you must call the 'Flush' method to end the
+// streaming writing process.
+func (sw *StreamWriter) SetRow(axis string, values []interface{}, styles []int) error {
```
xuri: The cyclomatic complexities of functions too high and it can be divided into multiple functions.

xuri: Create a new exported type `Cell` that can be wrapped by the interface, type `Cell` include integer type StyleID filed, instead of the third parameter `styles`.

stream.go
```diff
@@ -66,45 +72,182 @@ func (f *File) NewStreamWriter(sheet string) (*StreamWriter, error) {
...
+
 	sw.SheetData.WriteString(fmt.Sprintf(`<row r="%d">`, row))
-	for i := 0; i < v.Len(); i++ {
+	for i, val := range values {
 		axis, err := CoordinatesToCellName(col+i, row)
 		if err != nil {
 			return err
 		}
-		switch val := v.Index(i).Interface().(type) {
+		c := xlsxC{R: axis, S: styles[i]}
```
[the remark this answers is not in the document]
chowey: You are correct, this does reduce memory usage. It also increases the speed by quite a bit.

Using the included benchmark, the results improved a lot on my computer:

    BenchmarkStreamWriter-4-old   514  2576155 ns/op  454918 B/op  6592 allocs/op
    BenchmarkStreamWriter-4-new  1614   777480 ns/op  147608 B/op  5570 allocs/op

I have added this commit so I am no longer using `xml.Encoder`.

stream.go
chowey: I have completely re-worked tables. I added an example in the new `AddTable` function.

stream.go
chowey: `SetTable` must be done before `SetRow`. I used this approach because it is difficult to read the table headers once they are written to a temp file.

I agree it is a poor approach and the code is complicated.

Therefore I changed my approach. I now have an `AddTable` function as you suggested. See 34fca084014aa0ed4300791a177a70dc583766cc.

This was a major re-work. But I made a mechanism to read the temp file so we can get the table headers.

I now do a much better job of creating the final XML during `Flush`. We do exactly one allocation to fit the entire file contents, and we don't do any more processing. I found this dropped my total memory usage significantly.

stream.go
chowey: Agreed. I have simplified this with 34fca084014aa0ed4300791a177a70dc583766cc.

chowey: This is now done with 443da043d89438707cde812c7dcb3efcc7bc0058.

chowey:
Some of the fixes required a lot of code changes. I am sorry for that. But I think the updated code is much better and I've seen less memory usage.

xuri:
@chowey, thanks very much for your contribution!

Astronomican:
When will this feature be released？

xuri:
Hi @Astronomican, the next version 2.1.0 of Excelize will be released on February 10th, 2020.

Issue #539, opened 2019-12-18
Errors skip in this library!


monoflash:
The library is written with a error skip at all.
Because of this, excel files do not open without error messages.

I was faced with the fact that the xml crashes with an error when the openned excel file in an encoding different from UTF-8, but the library ignores this error not reporting the error in any way.

After looking at the code, I found error skipping almost everywhere, and as a result, at the first not significant error everything stops working.

To my regret, I have to abandon the use of this library, but I really hope that all the error skips will be processed and the library can be usable again.

xuri:
Hi @monoflash, thanks for your issue. Please provide .xlsx file attachment that created in encoding different from UTF-8 if you can.

monoflash:
This is a financial report provided by our contractors. When I delete all confidential information from it and save it, it is transcoded to UTF-8. And along with the financial information, I can’t provide the file.
I will try later to prepare an example file that can be provided to you.

monoflash:
I manually edited the XML file contained in xlsx.
file: xl/worksheets/sheet1.xml

If you look at it, then its code page is not UTF-8, for such XML files, the XML library requires a transcoder.

The fact that XML file is not encoded in UTF-8 is normal, Microsoft Excel accepts this well.

The only thing I did was manually delete the data from XML, so when open this xlsx, there will be a message about data corruption.
I think this is enough for an example.

[m2.xlsx](https://github.com/360EntSecGroup-Skylar/excelize/files/3980870/m2.xlsx)

PR #540, opened 2019-12-18
Fixed issue #539 Fixed error opening excel file created in encoding d…

monoflash:
Fixed issue #539 Fixed error opening excel file created in encoding different from UTF-8, added logging of possible errors when decoding xml, if the function does not provide exit with error


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/360EntSecGroup-Skylar/excelize/v2
+module github.com/monoflash/excelize/v2
```
xuri: Need to keep it as `module github.com/360EntSecGroup-Skylar/excelize/v2`.

picture.go
```diff
@@ -16,6 +16,7 @@ import (
 	"errors"
 	"image"
 	"io/ioutil"
+	"log"
```
xuri: I propose to return an error to the caller of this function instead of log errors.

sheet_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/360EntSecGroup-Skylar/excelize/v2"
+	"github.com/monoflash/excelize/v2"
```
xuri: Need to keep it as `github.com/360EntSecGroup-Skylar/excelize/v2`.

sheetpr_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/mohae/deepcopy"
 	"github.com/stretchr/testify/assert"
 
-	"github.com/360EntSecGroup-Skylar/excelize/v2"
+	"github.com/monoflash/excelize/v2"
```
xuri: Same as mentioned above.

sheetview_test.go
```diff
@@ -6,7 +6,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
-	"github.com/360EntSecGroup-Skylar/excelize/v2"
+	"github.com/monoflash/excelize/v2"
```
xuri: Same as mentioned above.

calcchain.go
```diff
@@ -9,16 +9,31 @@
 
 package excelize
 
-import "encoding/xml"
+import (
+	"bytes"
+	"encoding/xml"
+	"io"
+	"log"
```
xuri: I propose to return an error to the caller of this function instead of log errors.

charset_reader.go
```diff
@@ -0,0 +1,38 @@
+package excelize
+
+//import "gopkg.in/webnice/debug.v1"
```
xuri: Remove invalid packages references.

charset_reader.go
```diff
@@ -0,0 +1,38 @@
...
+	"io"
+	"regexp"
+	"strings"
+
+	"golang.org/x/text/encoding"
+	"golang.org/x/text/encoding/charmap"
+)
+
+// CharsetReader Decoder from all codepages to UTF-8
+func CharsetReader(charset string, input io.Reader) (rdr io.Reader, err error) {
```
xuri: Please add the tests to cover your changes.

monoflash:
Oops.. I didn’t think that the commits following the pool-request would be added to the pool-request.

picture.go
monoflash: In your library design, a lot of functions do not return an error, but an error is possible inside the function. There is an option to use panic, but it is much worse than sending an error to the log.

Everywhere where it is possible to return an error from a function, an error is returned, where the design does not provide for an error, I had to insert a log.

charset_reader.go
monoflash: ok

charset_reader.go
monoflash: ok

monoflash:
I fixed all remarks you wrote about.

charset_reader.go
```diff
@@ -0,0 +1,38 @@
...
```
xuri: Can we use [charset.NewReaderLabel](https://godoc.org/golang.org/x/net/html/charset#NewReaderLabel) instead of the `CharsetReader` function?

Just like this:

```go
decoder.CharsetReader = func(c string, i io.Reader) (io.Reader, error) {
    return charset.NewReaderLabel(c, i)
}
```

.gitignore
```diff
@@ -1,4 +1,5 @@
 ~$*.xlsx
 test/Test*.xlsx
 *.out
-*.test
\ No newline at end of file
+*.test
+.idea
```
xuri: remove .idea

picture.go
```diff
@@ -471,39 +472,55 @@ func (f *File) GetPicture(sheet, cell string) (string, []byte, error) {
...
+		XlsxRelationship    *xlsxRelationship
+		DecodeTwoCellAnchor *decodeTwoCellAnchor
+	)
+
+	wsDr, _ = f.drawingParser(drawingXML)
+	for _, anchor = range wsDr.TwoCellAnchor {
 		if anchor.From != nil && anchor.Pic != nil {
 			if anchor.From.Col == col && anchor.From.Row == row {
-				xlsxRelationship := f.getDrawingRelationships(drawingRelationships,
+				XlsxRelationship = f.getDrawingRelationships(drawingRelationships,
```
xuri: I propose to keep non-exportable variables, lowercase first letter.

go.sum
```diff
@@ -11,4 +11,7 @@ github.com/stretchr/testify v1.3.0 h1:TivCn/peBQ7UY8ooIcPgZFpTNSz0Q2U6UrFlUfqbe0
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
 golang.org/x/image v0.0.0-20190910094157-69e4b8554b2a h1:gHevYm0pO4QUbwy8Dmdr01R5r1BuKtfYqRqF0h/Cbh0=
 golang.org/x/image v0.0.0-20190910094157-69e4b8554b2a/go.mod h1:FeLwcggjj3mMvU+oOTbSwawSJRM1uh48EjtB4UJZlP0=
+golang.org/x/text v0.3.0 h1:g61tztE5qeGQ89tm6NTjjM9VPIm088od1l6aSorWRWg=
 golang.org/x/text v0.3.0/go.mod h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ=
+gopkg.in/webnice/debug.v1 v1.0.0 h1:vslbCq7qIWgCwW2Ryxfefb5NuM3dwY1zuRx2a220U2s=
```
xuri: Remove unused dependencies

charset_reader_test.go
```diff
@@ -0,0 +1,112 @@
+package excelize
+
+//import "gopkg.in/webnice/log.v2"
```
xuri: Remove unused package

picture.go
monoflash: Lowercase variables match structure names.
This overrides the name of the structure inside the function and no longer makes it possible to use the structure. The local variable inside the function is not exportable.

.gitignore
monoflash: This is GoLand. What for?

charset_reader.go
monoflash: I did not check the correctness of the search for encodings used in NewReaderLabel without name conversion. With a probability of 99%, no encoding will be found and thus the functionality will be useless. Moreover, some of the names of the encodings used in the charset library do not correspond to the canonical names in the RFC/ISO. I even thought to make synonyms for the correct search for the encoding by name.
Your proposed option
1. Creates a lot of redundant code due to the need to use an anonymous function - a wrapper
2. Will stop working for 90% of encodings

monoflash: If you do not like the presence of CharsetReader inside the library, then this can be fixed by making it possible to injects CharsetReader from the outside.

Adding type:
```
type CharsetTranscoderFn func(charset string, input io.Reader) (rdr io.Reader, err error)
```

Adding function:
```
func OpenReader(fn CharsetTranscoderFn) *File
```

And use this:
```
obj, err := excelize.CharsetTranscoder(fn CharsetTranscoderFn).OpenFile(filename)
```

monoflash:
I added the function that you requested, and deleted my custom function, also fixed some skipped returned error and fix tests.

xuri:
Thanks @monoflash. I will add error return values to functions later, and unit tests for this PR.

PR #561, opened 2020-01-19
Add SetColsVisibleFrom() to fix #559

### LINKED ISSUE #559: Hide all columns or rows, starting from one
Hello, I'd like to hide all rows or columns from one.

At this point, a function using the current implementation of changing the visibility of a row makes the file too large (and it takes time to do so..).

Amodio:
# PR Details

We need a function to hide both rows or columns from one, this PR deals only with columns.

## Description

Fixed some typos. Creation of SetColsVisibleFrom(), which is tested.

## Related Issue

#559 

## Types of changes

### DIFF
--- a/col.go
+++ b/col.go
@@ -26,7 +26,7 @@ const (
 // worksheet name and column name. For example, get visible state of column D
 // in Sheet1:
 //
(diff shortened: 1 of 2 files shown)


Amodio:
Bug

PR #562, opened 2020-01-19
Add SetColsVisibleFrom() to fix #559

### LINKED ISSUE #559: Hide all columns or rows, starting from one

Amodio:
# PR Details

We need a function to hide both rows or columns from one, this PR deals only with columns.

## Description

Fixed some typos. Creation of SetColsVisibleFrom(), which is tested.

## Related Issue

#559 

## Types of changes

### DIFF
--- a/col.go
+++ b/col.go
@@ -26,7 +26,7 @@ const (
 // worksheet name and column name. For example, get visible state of column D
 // in Sheet1:
 //
-//    visiable, err := f.GetColVisible("Sheet1", "D")
+//    visible, err := f.GetColVisible("Sheet1", "D")
 //
 func (f *File) GetColVisible(sheet, col string) (bool, error) {
 	visible := true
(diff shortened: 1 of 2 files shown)


col.go
```diff
@@ -91,6 +91,46 @@ func (f *File) SetColVisible(sheet, col string, visible bool) error {
 	return err
 }
 
+// SetColsVisibleFrom provides a function to set visible of all the columns
+// after a given one in a worksheet. For example, hide column D in Sheet1:
+//
+//    err := f.SetColsVisibleFrom("Sheet1", "D", false)
+//
+func (f *File) SetColsVisibleFrom(sheet, col string, visible bool) error {
```
xuri: I propose that modify the exist function `SetColVisible` like `SetColStyle`:

```go
func SetColVisible(sheet, columns string, visible bool) error
```

to make it accept the specified column range:

```go
SetColVisible("Sheet1", "F", false)
SetColVisible("Sheet1", "F:XFD", false)
```

Amodio: Ok, i'll post that soon.

Amodio: Please see PR #565

Amodio:
Reposted: PR #565

PR #589, opened 2020-02-28
Reduce redundant memory copy in getRowHeight

xxb-at-julichina:
# PR Details

## Motivation and Context

https://github.com/360EntSecGroup-Skylar/excelize/blob/821a5d86725eb80b3f9e806d91eca5859497c2fa/rows.go#L241
v is copy by value, which cause a lot call to runtime.duffcopy, lead to slow performance.
Will this patch, generate a workbook with 8320 AddPicture calls take only 29s instead of 127s.

## Types of changes

## Checklist


xuri:
LGTM, thanks @xxb-at-julichina.

PR #606, opened 2020-04-06
Fix wrong example in README

linzhengen:
# PR Details
Hi,
I found wrong example in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,14 +73,14 @@ func main() {
         return
     }
     // Get value from cell by given worksheet name and axis.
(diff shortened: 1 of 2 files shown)


xuri:
Hi @linzhengen, thanks for your PR. This PR duplicate with #583, since [v2.0.0](https://github.com/360EntSecGroup-Skylar/excelize/releases/tag/v2.0.0), we need to receive the error of [`GetCellValue`](https://github.com/360EntSecGroup-Skylar/excelize/blob/66d0272f6af59b5f0c97a304379a795420a43e8b/cell.go#L37) function, so I will close this PR.

PR #633, opened 2020-05-11
Add set rows visibility method

### LINKED ISSUE #626: Allow multiple Rows to be Hidden
Hi I´m using this to restrict a sheet, so, I don't have any problem with hiding colums but there's no option to do the same with rows, it has to be one by one and takes much more time.

Could you tell me if there's another way to do it not using "SetRowVisible" but something that allows to hide more rows in one instruction?

elbuki:
# PR Details

## Description

This should create a method that allows the developer to apply visibility to multiple rows.

## Related Issue

Resolves #626 

## Motivation and Context

This could work as an enhancement to allow the developer to use a single statement to apply the visibility to multiple rows.

## How Has This Been Tested

I've added test cases to cover the new method with its possible error.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist

- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -26,3 +26,7 @@ func newInvalidCellNameError(cell string) error {
 func newInvalidExcelDateError(dateValue float64) error {
 	return fmt.Errorf("invalid date value %f, negative values are not supported supported", dateValue)
 }
+
+func newInvalidRowNumbersError() error {
+	return fmt.Errorf("invalid rows value, no rows to process")
+}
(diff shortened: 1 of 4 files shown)


xuri:
Thanks for your PR. Sorry, after evaluation and follow the principle of least availability, I don't purpose to add this API. The hidden attributes of multiple hidden columns can be stored in an element (Ref: ECMA-376 Part 1 §18.3.1.13), but hidden rows will set hidden attributes on each row (§18.3.1.73), `SetRowsVisibility` is same as calling `SetRowVisible` multiple times. So I'll close PR and related issues.

hidden columns

```xml
<cols>
    <col min="5" max="8" width="0" hidden="1" customWidth="1"/>
</cols>
```

hidden rows

```xml
<sheetData>
    <row r="10" hidden="1"/>
    <row r="11" hidden="1"/>
</sheetData>
```

PR #662, opened 2020-07-08
Resolve #661 Add Logarithmic scale option support on Y axis

### LINKED ISSUE #661: Add Logarithmic scale option support on Y axis
**Description**

Provide option to enable the Logarithmic scale option in Y axis for charts.

1. Add a line-type chart

No option to enable the Logarithmic scale option in Y axis

Being able to enable the Logarithmic scale option in Y axis

**Output of `go version`:**

```text
go version go1.14.4 windows/amd64
```

```text
f7bd0729c65fc82305328f7ac8fbaf329d1075c0
```

Windows 10 x64 2004
Excel 2016

huybuidev:
# PR Details

Provide support for "Logarithmic scale option" on Y axis when drawing a chart.

## Description

1. Add member `LogBase` to the struct `cScaling` so that it will be generated in the final XML as follow:
```xml
<c:scaling>
    <c:logBase val="10"/>
    <c:orientation val="minMax"/>
</c:scaling>
```
2. Add `Scaling.LogBase` option in the `formatChartAxis` struct for the format string

## Related Issue

Add Logarithmic scale option support on Y axis #661

## Motivation and Context

Implement new feature

## How Has This Been Tested

Add the following option into the format string when using AddChart to add a 2D Line chart:
```json
"y_axis":{"scaling":{"logbase":10}}
```

## Types of changes

## Checklist

### DIFF
--- a/xmlChart.go
+++ b/xmlChart.go
@@ -380,6 +380,7 @@ type cChartLines struct {
 // cScaling directly maps the scaling element. This element contains
 // additional axis settings.
 type cScaling struct {
+	LogBase     *attrValString `xml:"logBase"`
 	Orientation *attrValString `xml:"orientation"`
 	Max         *attrValFloat  `xml:"max"`
 	Min         *attrValFloat  `xml:"min"`
@@ -544,6 +545,9 @@ type formatChartAxis struct {
 		Italic    bool   `json:"italic"`
 		Underline bool   `json:"underline"`
 	} `json:"num_font"`
+	Scaling struct {
+		LogBase string `json:"logbase"`
+	} `json:"scaling"`
 	NameLayout formatLayout `json:"name_layout"`
 }
 
--- a/xmlChart.go
+++ b/xmlChart.go
@@ -545,9 +545,7 @@ type formatChartAxis struct {
 		Italic    bool   `json:"italic"`
 		Underline bool   `json:"underline"`
 	} `json:"num_font"`
-	Scaling struct {
-		LogBase float64 `json:"logbase"`
-	} `json:"scaling"`
+	LogBase    float64      `json:"logbase"`
 	NameLayout formatLayout `json:"name_layout"`
 }
 
--- a/xmlChart.go
+++ b/xmlChart.go
@@ -380,7 +380,7 @@ type cChartLines struct {
 // cScaling directly maps the scaling element. This element contains
 // additional axis settings.
 type cScaling struct {
-	LogBase     *attrValString `xml:"logBase"`
+	LogBase     *attrValFloat  `xml:"logBase"`
 	Orientation *attrValString `xml:"orientation"`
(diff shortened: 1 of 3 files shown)


xmlChart.go
```diff
@@ -380,6 +380,7 @@ type cChartLines struct {
 // cScaling directly maps the scaling element. This element contains
 // additional axis settings.
 type cScaling struct {
+	LogBase     *attrValString `xml:"logBase"`
```
xuri, line 383: The data type of `logBase` should be an integer, by [OOXMLSchemas](https://github.com/sc34wg4/OOXMLSchemas/blob/2b074ca2c5df38b18ac118646b329b508b5bdecc/Part1/OfficeOpenXML-XMLSchema-Strict/dml-chart.xsd#L1142-L1147), I propose that change its type to `attrValInt `, and check its min and max value (2-1000). in addition, could you add the unit testing case for the changes?

huybuidev: Thank you very much for the suggestion, I couldn't find quickly where to look for the standard definition so just refer from a real XLSX file without knowing the type or value range.

I have updated the code accordingly and added test cases as well.

xmlChart.go
```diff
@@ -380,6 +380,7 @@ type cChartLines struct {
 // cScaling directly maps the scaling element. This element contains
 // additional axis settings.
 type cScaling struct {
+	LogBase     *attrValFloat  `xml:"logBase"`
```
xuri, line 383: I think the data type of `logBase` in the Go language should be int at here.

xmlChart.go
```diff
@@ -544,6 +545,9 @@ type formatChartAxis struct {
 		Italic    bool   `json:"italic"`
 		Underline bool   `json:"underline"`
 	} `json:"num_font"`
+	Scaling struct {
+		LogBase float64 `json:"logbase"`
```
xuri, line 549: I suggest `LogBase` as a field of the struct `formatChartAxis`, instead of `Scaling`’s

xmlChart.go
huybuidev: But the according to [OOXMLSchemas](https://github.com/sc34wg4/OOXMLSchemas/blob/2b074ca2c5df38b18ac118646b329b508b5bdecc/Part1/OfficeOpenXML-XMLSchema-Strict/dml-chart.xsd#L1142-L1147), it's `xsd:double` and Excel does support the log base with a float number (you can see one of my test case with logbase 10.5 and the output XLSX file), so I think attrVarFloat should be used here.

xmlChart.go
huybuidev: I used `Scaling` because from what I saw in the `charts*.xml` file, the `LogBase` option was put under `Scaling`:
```xml
```

xuri:
LGTM, thanks @huybuidev.

Issue #467, opened 2019-08-08
SetColStyle not honoring given style


SteveTalt:
**Description**
SetColStyle seems to be corrupting spreadsheets.    It doesn't seem to matter what data is in the column or which column or sheet you use.

I created this simple code to re-produce it:

```
package main

import (
	"github.com/excelize"
)

// define globals

var xlfile = "testlocal.xlsx"
var xlsheetname = "test"

// main
func main() {
	var (
		file *excelize.File
		err  error
	)
	// open the file
	file, err = excelize.OpenFile(xlfile)
	errorhandler(err)

	err = file.SetColStyle(xlsheetname, "C", 7)
	errorhandler(err)
	file.Save()

}

func errorhandler(e error) {
	if e != nil {
		panic(e)
	}
}
```

**Steps to reproduce the issue:**
1.   Create a spreadsheet named localtest.xlsx and a sheet named test
2.   run this test code against it
3.   open the spreadsheet and see it needs repair

**Describe the results you received:**
![image](https://user-images.githubusercontent.com/12849870/62726322-d97f2980-b9cb-11e9-81fe-ab3a47d6daed.png)
Repair Log:
`
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<recoveryLog xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main"><logFileName>Repair Result to testlocal5.xml</logFileName><summary>Errors were detected in file '/Users/stalt/Documents/go/src/setcoltest/testlocal.xlsx'</summary><repairedRecords summary="Following is a list of repairs:"><repairedRecord>Repaired Records: Column information from /xl/worksheets/sheet1.xml part</repairedRecord></repairedRecords></recoveryLog>

`

**Describe the results you expected:**
I expect the column to be formated with 

**Output of `go version`:**

```text
$ go version
go version go1.10 darwin/amd64
```

**Excelize version or commit ID:**

```text
commit id: 448f552
```

**Environment details (OS, Microsoft Excel™ version, physical, etc.):**
macOS Mojave 10.14.5
Excel for Mac 16.27 (19071500)

SteveTalt:
Weirdly, if i do this:
	
`style, err = file.NewStyle(`{"number_format": 7}`)
err = file.SetColStyle(xlsheetname, "C", style)`

the value of style is 2?? and it formatting to 2 style and doesn't corrupt it.   It may be that format 7 doesn't exist.

[...]

PR #694, opened 2020-08-19
Fix SetColStyle

remoremorali:
# PR Details

SetColStyle was returning with no action when no columns were added
to the worksheet before.

## Description

This commit uses the same approach as the other col.go functions and
adds the columns to the worksheet if there are none.

## Related Issue

#467 

## Motivation and Context

Fixes 467
Now if styles are added before any column it's present, they are not ignored.

## How Has This Been Tested

Unittests.
The test was failing before the fix, and it's passing now

## Types of changes

## Checklist

### DIFF
--- a/col.go
+++ b/col.go
@@ -407,15 +407,19 @@ func (f *File) SetColStyle(sheet, columns string, styleID int) error {
 	if max < min {
 		min, max = max, min
 	}
-	if xlsx.Cols == nil {
-		xlsx.Cols = &xlsxCols{}
-	}
-	xlsx.Cols.Col = flatCols(xlsxCol{
+	col := xlsxCol{
 		Min:   min,
 		Max:   max,
 		Width: 9,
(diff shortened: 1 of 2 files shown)


xuri:
@remoremorali Thanks for your PR. I have tested setting column style on the master branch code, and it works well. So I think maybe needn't fix that. After reading related issues, I still don't know the problem clearly. Could you provide the complete code and attachment to reproduce the issue?

remoremorali:
You're right!
I had a problem with applying a date format which worked with SetCellStyle, hence I looked at the code and I thought that I had spotted the issue, and indeed from my tests the style was not there; but it did not influence the styling and it works, it turns out, _except_ when a date value is set before applying the formatting; which is a different bug that this PR won't fix.
You can discard this PR, thank you. I'll check if a bug is already there for the date issue and open one otherwise.

Issue #73, opened 2017-06-29
Read date/time from the cell


vchugreev:
How to correctly read the date/time from the cell?

vchugreev:
I will explain the problem:

In the Excel file there is a cell:
E7 => 29.06.2017 9:28:48
Format: "14-Mar-12"

I'm trying to get it...

`e7 := xlsx.GetCellValue("Sheet1", "e7")`
`println(e7)`

Output: "42915.395"

Luxurioust:
Do you mean `Sheet1!E7` is string type value `29.06.2017 9:28:48`? If so, string type value can't be formatted as a date format. Please provide file attachments If you can.

vchugreev:
xlsx?
[1_2017-06-29.xlsx](https://github.com/Luxurioust/excelize/files/1112114/1_2017-06-29.xlsx)

Luxurioust:
This library just support Excel's [built-in formats](https://godoc.org/github.com/Luxurioust/excelize#File.SetCellStyle) currently. The built-in formats 
 index (`numFmtId = 164`) in the attachment is out of this range. Same as the issue #50.

vchugreev:
Now it is clear. Thank you!

akatsukle:
I am not sure I understand how SetCellStyle works.
On a comment above the function there is the following:
```
//     Index | Format String
//    -------+----------------------------------------------------
//     0     | General
//     1     | 0
//     2     | 0.00
//     3     | #,##0
//     4     | #,##0.00
//     5     | ($#,##0_);($#,##0)
//     6     | ($#,##0_);[Red]($#,##0)
//     7     | ($#,##0.00_);($#,##0.00)
//     8     | ($#,##0.00_);[Red]($#,##0.00)
//     9     | 0%
//     10    | 0.00%
//     11    | 0.00E+00
//     12    | # ?/?
//     13    | # ??/??
//     14    | m/d/yy
//     15    | d-mmm-yy
//     16    | d-mmm
//     17    | mmm-yy
//     18    | h:mm AM/PM
//     19    | h:mm:ss AM/PM
//     20    | h:mm
//     21    | h:mm:ss
//     22    | m/d/yy h:mm
```

With this in mind the following makes perfect sense
```
xlsx.NewStyle(`{"number_format":15}`) // 29-Jun-17
```

However, when I tried for `m/d/yy`, I get the following:
```
xlsx.NewStyle(`{"number_format":14}`) // 11-15-21
[...]

PR #703, opened 2020-09-10
extend cell value load to support custom datetime format

artiz:
# PR Details

Add support for custom number formats to support all date-time variants parsing

## Description

We are using excelize in our ETL project tp parse variety XLSX files from our customers and had found that lots of required dates are parsed incorrecty as raw numbers. So we decided to extend dates support to use custom formats

## Related Issue
PR provides fix for https://github.com/360EntSecGroup-Skylar/excelize/issues/73

## Motivation and Context
PR provides support of custom Excel datetime formats

## How Has This Been Tested
New Unit Test added

## Types of changes

## Checklist


templates.go
```diff
@@ -29,7 +29,7 @@ const templateContentTypes = `<Types xmlns="http://schemas.openxmlformats.org/pa
...
...org/markup-compatibility/2006" mc:Ignorable="x14ac x16r2" xmlns:x14ac="http://schemas.microsoft.com/office/spreadsheetml/2009/9/ac" xmlns:x16r2="http://schemas.microsoft.com/office/spreadsheetml/2015/02/main"><numFmts count="2"><numFmt formatCode="General" numFmtId="100"/><numFmt formatCode="[$-409]MM/DD/YYYY" numFmtId="101"/></numFmts><fonts count="1" x14ac:knownFonts="1"><font><sz val="11"/><color theme="1"/><name val="Calibri"/><family val="2"/></font></fonts><fills count="2"><fill><patternFill patternType="none"/></fill><fill><patternFill patternType="gray125"/></fill></fills><borders count="1"><border><left/><right/><top/><bottom/><diagonal/></border></borders><cellStyleXfs count="1"><xf numFmtId="0" fontId="0" fillId="0" borderId="0"/></cellStyleXfs><cellXfs count="3"><xf numFmtId="0" fontId="0" fillId="0" borderId="0" xfId="0"/><xf numFmtId="100" fontId="0" fillId="0" borderId="0" xfId="0"/><xf numFmtId="101" fontId="0" fillId="0" borderId="0" xfId="0"/></cellXfs><cellStyles count="1"><cellStyle name="Normal" xfId="0" builtinId="0"/></cellStyles><dxfs count="0"/><tableStyles count="0" defaultTableStyle="TableStyleMedium2" defaultPivotStyle="PivotStyleLight16"/></styleSheet>`
```
xuri, line 32: That's unnecessary to add number format code 100 and 101 in the template.

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/360EntSecGroup-Skylar/excelize/v2
+module github.com/artiz/excelize
```
xuri, line 1: the incorrect package name in the go module.

xuri (changes requested):

go.mod
artiz: yep, package with default name in go.mod https://github.com/artiz/excelize/tree/extended-parse-time could not be installed by `go mod` so it was temporary fix, will revert it

go.sum
```diff
@@ -1,35 +1,50 @@
+github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ33E=
```
xuri, line 1: Remove unused dependencies version checksum info

xuri (changes requested):
In addition, I suggest removing personal info from the doc properties of the test spreadsheet files, and could you add some unit tests for the non covered codes?

templates.go
artiz: Yep, I did it to cover new logic that I had added, without it PR coverage check raises red flag.
https://github.com/artiz/excelize/commit/b1c578c0ba723285b6adba7eae9c6e9023f30322 - actually in this commit

artiz:
> In addition, I suggest removing personal info from the doc properties of the test spreadsheet files, and could you add some unit tests for the non covered codes?

Got it, will do

xuri (changes requested):
I have tested by the spreadsheet attachment `1_2017-06-29.xlsx` in the issue #73 uploaded by `vchugreev`, It seems like the value `29\ June\ 2017\ \г\.;@` after applying number format code contains escape characters,

go.sum
artiz: done

artiz:
> I have tested by the spreadsheet attachment `1_2017-06-29.xlsx` in the issue #73 uploaded by `vchugreev`, It seems like the value `29\ June\ 2017\ \г\.;@` after applying number format code contains escape characters.

Fixed in https://github.com/artiz/excelize/commit/6e66d1dab42abf6bfce59685825e1ccd46bba2ad

omaralsoudanii:
Hi @artiz @xuri 

Thank you for the amazing work you are doing, The library is pretty great when reading XLSX files, however I am facing the same problem when parsing dates. 

Some times the date value is parsed as string e.g: 25/03/2016, some times it's parsed as a float like 43345.5 and some times as an integer.

I believe this PR would probably solve it? if so do you have any estimate when this will be merged?

Thanks!

xuri (approved):

xuri:
@artiz Thanks for your PR. @omaralsoudanii I have merged this PR.

PR #729, opened 2020-11-14
Fix row duplicate mechanism

EugeneAndrosovPaser:
# Fix row duplication mechanism

## Description
When you want to copy row - formulas copied as a reference from the previous row. If you want to change formulas in the current row they also would be changed in the original row.

## Motivation and Context
This fix makes a deep copy of the original row to prevent its behavior.

## How Has This Been Tested
It was tested manually and using it in my project.

## Types of changes

## Checklist


xuri:
LGTM, thanks for your PR.

PR #739, opened 2020-12-09
support range  validation for decimal

jacentsao:
# PR Details

support range  validation for decimal

## Description

support range  validation for decimal

## Motivation and Context

support range  validation for decimal not only integer


xuri:
LGTM, thanks @jacentsao.

Issue #787, opened 2021-02-20
SetCellRichText is called multiple times on the same cell will increase the file size


tonny-zhang:
**I found the result excel file size increased after I called multiple times method SetCellRichText on same cell**

**Steps to reproduce the issue:**
1. one time
```go
package excelize

import (
	"testing"
)

func TestGetCellRichText(t *testing.T) {
	f := NewFile()

	// for i := 0; i < 10000; i++ {
	f.SetCellRichText("Sheet1", "F10", []RichTextRun{
		{
			Text: "a1：\n",
		},
		{
			Text: "      b1   ",
			Font: &Font{
				Underline: "single",
				Color:     "#123",
				Bold:      true,
				Italic:    true,
				Family:    "宋体",
				Size:      100,
				Strike:    true,
			},
		},
	})
	// }

	f.SaveAs("1.xlsx")
}
```
2. multiple times
```go
package excelize

import (
	"testing"
)

func TestGetCellRichText(t *testing.T) {
	f := NewFile()

	for i := 0; i < 10000; i++ {
		f.SetCellRichText("Sheet1", "F10", []RichTextRun{
			{
				Text: "a1：\n",
			},
			{
				Text: "      b1   ",
				Font: &Font{
					Underline: "single",
					Color:     "#123",
					Bold:      true,
					Italic:    true,
					Family:    "宋体",
					Size:      100,
					Strike:    true,
				},
			},
		})
	}

	f.SaveAs("2.xlsx")
}

```
3.  `ls -las *.xlsx`

**Describe the results you received:**
```
16 -rw-r--r--@ 1 tonny  staff   6164  2 20 16:32 1.xlsx
32 -rw-r--r--@ 1 tonny  staff  15472  2 20 16:33 2.xlsx
```
**Describe the results you expected:**
```
16 -rw-r--r--@ 1 tonny  staff   6164  2 20 16:32 1.xlsx
32 -rw-r--r--@ 1 tonny  staff  6164  2 20 16:33 2.xlsx
```
**Output of `go version`:**

```text
go version go1.15.6 darwin/amd64
```

**Excelize version or commit ID:**

```text
(paste here)
```

**Environment details (OS, Microsoft Excel™ version, physical, etc.):**

tonny-zhang:
**I fixed this bug**

cell.go -> SetCellRichText
```
func (f *File) SetCellRichText(sheet, cell string, runs []RichTextRun) error {
[...]

PR #788, opened 2021-02-20
fixed SetCellRichText bug

tonny-zhang:
# fixed SetCellRichText bug 

## Related Issue
https://github.com/360EntSecGroup-Skylar/excelize/issues/787

## Types of changes

## Checklist

### DIFF
--- a/xmlWorksheet.go
+++ b/xmlWorksheet.go
@@ -462,9 +462,10 @@ type xlsxC struct {
 	R        string   `xml:"r,attr,omitempty"` // Cell ID, e.g. A1
 	S        int      `xml:"s,attr,omitempty"` // Style reference.
 	// Str string `xml:"str,attr,omitempty"` // Style reference.
-	T  string  `xml:"t,attr,omitempty"` // Type.
-	F  *xlsxF  `xml:"f,omitempty"`      // Formula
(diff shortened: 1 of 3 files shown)


xmlWorksheet.go
```diff
@@ -462,9 +462,10 @@ type xlsxC struct {
 	R        string   `xml:"r,attr,omitempty"` // Cell ID, e.g. A1
 	S        int      `xml:"s,attr,omitempty"` // Style reference.
 	// Str string `xml:"str,attr,omitempty"` // Style reference.
-	T  string  `xml:"t,attr,omitempty"` // Type.
-	F  *xlsxF  `xml:"f,omitempty"`      // Formula
-	V  string  `xml:"v,omitempty"`      // Value
+	T  string  `xml:"t,attr,omitempty"`  // Type.
+	T2 string  `xml:"t2,attr,omitempty"` // Type2. (richtext)
```
xuri, line 466: The field `T2` is not defined in the ECMA-376, ISO/IEC 29500 spec, and it doesn't have a good semanticization, I don't think this fixed in the right way.

xuri (changes requested):

xmlWorksheet.go
tonny-zhang: got it, I will try other

PR #793, opened 2021-03-03
Fix UpdateLinkedValue which returns an error when has graph sheet

yuki2006:
# PR Details

If you create a graph sheet by the following procedure,
`UpdateLinkedValue` fails and no other sheets are recalculated.

(sorry for the Japanese page)
https://www.officepro.jp/excelgraph/init/index2.html#section1

image

## Description

Create a custom error `ChartSheetError` and
In `UpdateLinkedValue`, if `ChartSheetError` is returned from `workSheetReader`, it will not return.

## Types of changes

## Checklist

### DIFF
--- a/excelize.go
+++ b/excelize.go
@@ -63,10 +63,6 @@ type Options struct {
 	Password string
 }
 
-type ChartSheetError struct{ err error }
-
-func (e ChartSheetError) Error() string { return e.err.Error() }
-
 // OpenFile take the name of an spreadsheet file and returns a populated spreadsheet file struct
 // for it. For example, open spreadsheet with password protection:
 //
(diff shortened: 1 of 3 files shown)


excelize_test.go
```diff
@@ -1242,6 +1242,15 @@ func TestAttrValToInt(t *testing.T) {
 	assert.EqualError(t, err, `strconv.Atoi: parsing "s": invalid syntax`)
 }
 
+func TestUpdateLinkedValue(t *testing.T) {
+	// Test update the spreadsheet file.
```
xuri, line 1246: I suggest that [prepare chart sheet](https://github.com/360EntSecGroup-Skylar/excelize/blob/7ef1892f320879670c4e9a8536a70281d782625e/chart_test.go#L212-L221) in new spreadsheet file instead of add new workbook.

excelize.go
```diff
@@ -313,7 +317,11 @@ func (f *File) UpdateLinkedValue() error {
 	wb.CalcPr = nil
 	for _, name := range f.GetSheetList() {
 		xlsx, err := f.workSheetReader(name)
-		if err != nil {
+		var chartSheetErr *ChartSheetError
```
xuri, line 320: We can just add a simple checking at here:

```go
if err != nil {
	if err.Error() == fmt.Sprintf("sheet %s is chart sheet", trimSheetName(name)) {
		continue
	}
	return err
}
```

xuri (changes requested):

yuki2006:
@xuri Thank you comments. I fixed it!

xuri:
LGTM, thank you @yuki2006.

Issue #831, opened 2021-04-27
SetCellValue looping yields ugly code


nnordrum:
**Description**

Using `SetCellValue` is "hard" when you're looping.  Because `axis` is a single variable representing two things, you have to combine it each time to loop.

e.g.
```go
const Sheet1 = "Sheet1"
f.SetCellValue(Sheet1, fmt.Sprintf("%s%d", "A", line), valA)
f.SetCellValue(Sheet1, fmt.Sprintf("%s%d", "B", line), valB)
```

If code were changed, you could have this:
```go
consts (
  Sheet1 = "Sheet1"

  ColA = iota + 1
  ColB
)
for l, x := range factory() {
  line := l+2
  f.SetCellValueNumericSplit(Sheet1, ColA, line, valA)
  f.SetCellValueNumericSplit(Sheet1, ColB, line, valB)
}
```

Much easier and cleaner IMO.

nnordrum:
I just discovered `SetSheetRow` in an example... playing with that now

nnordrum:
that's way more convenient! thoughts on adding this to `SetSheetRow`? (also? instead?)

xuri:
Base on the discussion on #832, I'll close this issue,  if you have any idea, please let me know.

PR #832, opened 2021-04-27
Update cell.go

nnordrum:
# PR Details

Added to API to better support looping and more concise code.

## Description

Basically just unpacked axis and made two helper methods with `col` as a `string` or `int`.

## Related Issue

https://github.com/360EntSecGroup-Skylar/excelize/issues/831

## Motivation and Context

Cleaner and more concise code.

## How Has This Been Tested

TODO: as this was just delegating methods, I wasn't sure how you wanted this tested. I will be more than happy to add test methods.

## Types of changes

## Checklist

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -43,6 +43,28 @@ func (f *File) GetCellValue(sheet, axis string) (string, error) {
 	})
 }
 
+// Opinionated version of SetCellValue which is better for loops
+func (f *File) SetCellValueSplit(sheet, col string, row int, value interface{}) error {
+	if name, err := JoinCellName(col, row); err == nil {
+		return f.SetCellValue(sheet, name, value)
+	} else {
+		return err
+	}
+}
+
+// Opinionated version of SetCellValue which is better for loops of loops
+func (f *File) SetCellValueNumericSplit(sheet string, col, row int, value interface{}) error {
+	if name, err := ColumnNumberToName(col); err == nil {
+		if joinedName, err := JoinCellName(name, row); err == nil {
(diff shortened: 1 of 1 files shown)


xuri:
Thanks for your PR, the Excelize already provided coordinates convert function between tuple integer and string, which can be used in the loop, for example, typically to generate 10 rows by 5 columns matrix cell like this, so I don't think this exported wrapper function is necessary.

```go
func main() {
    file := excelize.NewFile()
    for r := 1; r <= 10; r++ {
        for c := 1; c <= 5; c++ {
            cell, err := excelize.CoordinatesToCellName(c, r)
            if err != nil {
                fmt.Println(err)
                continue
            }
            file.SetCellValue("Sheet1", cell, "--")
        }
    }
    if err := file.SaveAs("Book1.xlsx"); err != nil {
        fmt.Println(err)
    }
}
```

nnordrum:
agreed. I would publicize this and `SetSheetRow` a LOT more. The only addition I'd make would be an "overload" of `CoordinatesToCellName` that didn't return a tuple so you could use it inline.

```go
func main() {
    file := excelize.NewFile()
    for r := 1; r <= 10; r++ {
        for c := 1; c <= 5; c++ {
            file.SetCellValue("Sheet1", excelize.CoordinatesToCellNameOrPanic(c, r), "--")
        }
    }
    if err := file.SaveAs("Book1.xlsx"); err != nil {
        fmt.Println(err)
    }
}
```

or whatever your preferred pattern would be.

Issue #844, opened 2021-05-19
The GetCellFormula in same cell  not working properly


raochq:
when I copy a sheet to another file.  The formula in same cell may not correct.

```
package main

import (
	"github.com/360EntSecGroup-Skylar/excelize/v2"
)

func main() {
	sheetName := "Sheet1"
	f1, _ := excelize.OpenFile("input.xlsx")
	f2 := excelize.NewFile()
	rows, _ := f1.GetRows(sheetName)
	for r, row := range rows {
		for c, s := range row {
			axis, _ := excelize.CoordinatesToCellName(c+1, r+1)
			formula, _ := f1.GetCellFormula(sheetName, axis)
			if formula != "" {
				f2.SetCellFormula(sheetName, axis, formula)
			} else {
				f2.SetCellValue(sheetName, axis, s)
			}
		}
	}
	f2.SaveAs("output.xlsx")
}
```
[input.xlsx](https://github.com/360EntSecGroup-Skylar/excelize/files/6506381/input.xlsx)

PR #846, opened 2021-05-19
fix shared formula incorrect (#844)

raochq:
fix shared formula incorrect (#844)


jaby:
@xuri Any chance this pull request will be accepted soon?  I've tried it out and it fixes a problem with a shared formula in the same worksheet.  In my case it concerns a range where the first row's B-column-formula  (B1) is returned for each of the following rows, resulting in the wrong B3, B4 etc cell calculation. 
Thanks in advance!

And thanks @raochq for the solution!

xuri:
I need some time to review this PR.

cwouyang:
> 
> 
> I need some time to review this PR.

Hi @xuri, any progress? I encountered same issue and needed this MR as well 🥺

xuri (commented):
Hi @raochq, thanks for your contribution. I have committed some changes base on your branch, in this commit:

- Separate sub function `parseSharedFormula` reduce cyclomatic complexities
- Make `SplitCellName` support absolute cell coordinates, and remove the `CellNameToCoordinates2`, `CellNameToCoordinates` is 2x faster and less memeory allocation than `CellNameToCoordinates2`
- Add unit test for the function `GetCellFormula`

PR #847, opened 2021-05-24
fix the bug that  there was no count attribute in   sharedStrings file, fix the bug  that setting OutlineSummaryBelow false

qwaszx102938:
# PR Details

1. fix the inserting new string value bug  when there is no  Count attribute in  sharedStrings  file。
2. fix the bug  in  “f.SetSheetPrOptions(sheetname,excelize.OutlineSummaryBelow(false)) method”
  
## Description

1. According the  part 1 of the 5th edition of the ECMA-376 Standard for Office Open XML document , 18.4.9 page 1724,  the count  Attributes in sst(Shared String Table) is optional。when count  attributes is missing  in the sst and you want to insert a new string value  to  the cell  in the sheet , there is a bug that the cell value index  will start by 0 ,missing the former values in 
sst(Shared String Table).So I insert a logic calculate the count  of former values  when the count attribute is missing   to fix the bug.

2. According the  part 1 of the 5th edition of the ECMA-376 Standard for Office Open XML document 18.3.1.61 page 1653. and  the definition of outlinePr,  the  default values of  “showOutlineSymbols ”，“summaryBelow” ，“summaryRight” attributes are true. if you set those to false ,  the attribute will be gone and the behavior will be true because of the  “omitempty” in the tags and the default value. so i try to remove the "omitempty" in the tag.

## Motivation and Context
fixing two bugs.

## How Has This Been Tested
test local

## Types of changes

## Checklist

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -322,6 +322,9 @@ func (f *File) sharedStringsReader() *xlsxSST {
 			Decode(&sharedStrings); err != nil && err != io.EOF {
 			log.Printf("xml decode error: %s", err)
 		}
+		if sharedStrings.Count == 0 {
+			sharedStrings.Count = len(sharedStrings.SI)
+		}
 		if sharedStrings.UniqueCount == 0 {
 			sharedStrings.UniqueCount = sharedStrings.Count
 		}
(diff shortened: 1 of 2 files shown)


xuri:
LGTM, thanks @qwaszx102938.

Issue #409, opened 2019-05-20
Don't require UTC time


carlmjohnson:
Excelize v2, unlike v1, requires time.Time to be UTC. I think this is a mistake because Excel has no notion of timezone, so you can't do the Right Thing, which would be to include UTC time and then format it to be the time zone of your intended audience. With v1, I just set the timezone to be EST (what my audience expects), but with v2, I have to set the timezone to EST, then copy the time over to a fake UTC time.

**Steps to reproduce the issue:**

Create a time.Time with a timezone other than UTC, and use `SetCellValue`.

**Describe the results you received:**

Received `errors.New("only UTC time expected")`.

**Describe the results you expected:**

No error.

**Excelize version or commit ID:**

v2.0.0

dolmen:
👍 
I see no reason why the code limits Location() to be UTC.
https://github.com/360EntSecGroup-Skylar/excelize/blob/master/date.go#L35

And this change has not been mentioned in changelog for v2. So this is a bug, isn't it?

gitdlam:
i did this as a workaround. 
```
func fixTime(t time.Time) time.Time {
	t, _ = time.ParseInLocation("2006-01-02 15:04", t.Format("2006-01-02 15:04"), time.UTC)
	return t
}
```

PR #853, opened 2021-05-31
#409 Remove UTC timezone requirement from date.go

Alluuu:
According to issue #409

There is absolutely no reason for the timezone to be in UTC, and converting the local times to UTC while keeping values is hacky at least.

Excel has no understanding of timezones, hence the user of this library should know what timezone their values are supposed to be, by following the timezone within their timeTime structs.

There is no need for changes in documentation, as this requirement doesn't seem to be documented.

# PR Details

## Types of changes

## Checklist

### DIFF
--- a/date.go
+++ b/date.go
@@ -30,13 +30,6 @@ var (
 func timeToExcelTime(t time.Time) (float64, error) {
 	// TODO in future this should probably also handle date1904 and like TimeFromExcelTime
 
-	// Force user to explicit convet passed value to UTC time.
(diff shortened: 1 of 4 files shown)


date.go
```diff
@@ -30,13 +30,6 @@ var (
 func timeToExcelTime(t time.Time) (float64, error) {
 	// TODO in future this should probably also handle date1904 and like TimeFromExcelTime
 
-	// Force user to explicit convet passed value to UTC time.
```
xuri, line 33: The failure unit test case should be fixed.

xuri (changes requested):

xuri (approved):

PR #863, opened 2021-06-18
feat: stream write to zip directly

zhangzitao:
# PR Details

Reduce the memory cost when saving file.

## Description

If options has set password, the process will write to buffer in memory. Or it will write directly to the file. 

## Types of changes

## Checklist


xuri:
Hi @zhangzitao, thanks for your PR! Could you add a test to cover the change?

xuri:
LGTM, thank you.

Issue #972, opened 2021-07-27
Data validation formula length limit is much stricter than Microsoft Excel


Arnie97:
### Description

Some valid data validation patterns with non-ASCII characters were rejected by Excelize.

### Steps to reproduce the issue:

```go
package main

import (
	"strings"
	"github.com/360EntSecGroup-Skylar/excelize/v2"
)

func main() {
	var cjkStrings = []string{
		strings.Repeat("漢", 255),
	}
	newWorkbookWithDropList(cjkStrings)
}

func newWorkbookWithDropList(choices []string) {
	f := excelize.NewFile()
	dv := excelize.NewDataValidation(false)
	dv.Sqref = "A:A"
	if err := dv.SetDropList(choices); err != nil {
		panic(err)
	}
	f.AddDataValidation(f.GetSheetName(0), dv)
	f.SaveAs("TestDropList.xlsx")
}
```

### Describe the results you received:

```go
panic: data validation must be 0-255 characters
```

### Describe the results you expected:

Produce a file `TestDropList.xlsx`, which could be loaded in Microsoft Excel successfully

### Output of `go version`:

```text
go version go1.17rc1 linux/amd64
```

### Excelize version or commit ID:

* master (e9ae9b4)
* v2.4.0 (d42834f)
* v2.1.0 (e51aff2)

### Environment details (OS, Microsoft Excel™ version, etc.):

The data validation pattern provided above were valid in all the following implementations, but not Excelize:

* Microsoft Office 365 for Mac, 16.49 (21050901)
* Microsoft Office 365 for Windows, 2102 (Build 13801.20808)
* Microsoft Office 2019 Professional Plus for Windows, 2105 (Build 14026.20302)
* Microsoft Office 2013 Professional Plus for Windows, 15.0.4919.1002

xuri:
Thanks for your feedback, I have fixed it, please try to use the master branch code, and this patch will be released in the next version.

PR #975, opened 2021-07-28
Fix data validation issues

### LINKED ISSUE #971: Drop-down list items are not XML escaped
### Description

If any of the XML special characters `<`, `>`, `&`, `"` were used in `SetDropList()` choices, the produced file will be reported as corrupted.

### Steps to reproduce the issue:

```go
package main

import (
	"github.com/360EntSecGroup-Skylar/excelize/v2"
)

func main() {
	var jobs = []string{"R&D", "PM", "UX", "HR"}
	newWorkbookWithDropList(jobs)
}

func newWorkbookWithDropList(choices []string) {
	f := excelize.NewFile()
	dv := excelize.NewDataValidation(false)
	dv.Sqref = "A:A"
	if err := dv.SetDropList(choices); err != nil {
		panic(err)
	}
	f.AddDataValidation(f.GetSheetName(0), dv)
	f.SaveAs("TestDropList.xlsx")
}
```

### Describe the results you received:

image

image

### Describe the results you expected:

image

### Output of `go version`:

```text
go version go1.17rc1 linux/amd64
```

### Excelize version or commit ID:

* master (e9ae9b4)
* v2.4.0 (d42834f)
* v2.1.0 (e51aff2)

### Environment details (OS, Microsoft Excel™ version, etc.):

The following official implementation were tested, and all of them could reproduce this issue:

* Microsoft Office 365 for Mac, 16.49 (21050901)
* Microsoft Office 365 for Windows, 2102 (Build 13801.20808)
* Microsoft Office 2019 Professional Plus for Windows, 2105 (Build 14026.20302)
* Microsoft Office 2013 Professional Plus for Windows, 15.0.4919.1002

The following third-party implementations were also tested, but none of them was affected by the issue:
* Kingsoft WPS
* Apple Numbers
* Google Sheets
* LibreOffice

Arnie97:
# PR Details

## Description

This patch escapes the drop-down list items, and changes the way to calculate the formula length.

## Related Issue

Fixes #971 and #972.

### Changes for #971
This patch included a XML entity mapping table, instead of simply reusing `xml.EscapeText()` in the standard library. This was on purpose. Disassembly of an OOXML file produced by Microsoft Excel indicated that, it implemented a different escaping rule which was more or less incompatible with `xml.EscapeText()`. Consider the following example:

```xml
A<,B>,C",D	,E',F
```

Microsoft Excel escaped it into:

```xml
<formula1>"A&lt;,B&gt;,C"",D	,E',F"</formula1>
```

And my initial attempt to fix #971 in commit b1faede323757cd48f81950a5105143832637859, which employed `xml.EscapeText()`, produced:

```xml
<formula1>"A&lt;,B&gt;,C&#34;,D&#x9;,E&#39;,F"</formula1>
```

The latter was unfortunately still incompatible with Microsoft Excel, not to mention it was longer:

image

### Changes for #972
This patch calculated the string length in UTF-16 code units instead of UTF-8 bytes.
Tests indicated that Microsoft Excel's limitation of data validation formula always worked in this way, even on UTF-8 dominated platforms like macOS. The following snippet was developed to help identification of the upper bound:

```go
package main

import (
	"fmt"
	"strings"
	"unicode/utf16"

	"github.com/360EntSecGroup-Skylar/excelize/v2"
)

func main() {
	var choices = []string{
		strings.Repeat("😅", 100), // U+1F605, "SMILING FACE WITH OPEN MOUTH AND COLD SWEAT"
		strings.Repeat("漢", 55),  // U+06F22, "CJK UNIFIED IDEOGRAPH-6F22"
	}

	joined := strings.Join(choices, ",")
	runes := []rune(joined)
	codeUnits := utf16.Encode(runes)

	fmt.Printf("%4d bytes in utf-8 \n", len(joined))
	fmt.Printf("%4d units in utf-16\n", len(codeUnits))
	fmt.Printf("%4d runes in utf-32\n", len(runes))

	newWorkbookWithDropList(choices)
}

func newWorkbookWithDropList(choices []string) {
	f := excelize.NewFile()
	dv := excelize.NewDataValidation(false)
	dv.Sqref = "A:A"
	if err := dv.SetDropList(choices); err != nil {
		panic(err)
	}
	f.AddDataValidation(f.GetSheetName(0), dv)
	f.SaveAs("TestDropList.xlsx")
}
```

The snippet would print ...

```js
 566 bytes in utf-8
 256 units in utf-16
 156 runes in utf-32
```

... and produce a corrupt file (or panic if the original length validation was not removed from excelize codebase).

image

Change the constant `55` to `54`, and then the snippet would print ...

```js
 563 bytes in utf-8
 255 units in utf-16
 155 runes in utf-32
```

... and produce a valid Excel workbook:

image

## How Has This Been Tested

The patches was tested against:

* Microsoft Office 365 for Mac, 16.49 (21050901)
* Microsoft Office 365 for Windows, 2102 (Build 13801.20808)
* Microsoft Office 2019 Professional Plus for Windows, 2105 (Build 14026.20302)
* Microsoft Office 2013 Professional Plus for Windows, 15.0.4919.1002

## Types of changes

## Checklist

### DIFF
--- a/datavalidation.go
+++ b/datavalidation.go
@@ -12,8 +12,6 @@
 package excelize
 
 import (
-	"bytes"
-	"encoding/xml"
 	"fmt"
 	"strings"
 )
@@ -76,6 +74,15 @@ const (
 	DataValidationOperatorNotEqual
 )
 
+// formulaEscaper mimics the Excel escaping rules for data validation,
+// which converts `"` to `""` instead of `&quot;`.
+var formulaEscaper = strings.NewReplacer(
+	`&`, `&amp;`,
+	`<`, `&lt;`,
+	`>`, `&gt;`,
+	`"`, `""`,
+)
+
 // NewDataValidation return data validation struct.
 func NewDataValidation(allowBlank bool) *DataValidation {
 	return &DataValidation{
@@ -112,13 +119,7 @@ func (dd *DataValidation) SetInput(title, msg string) {
 
 // SetDropList data validation list.
 func (dd *DataValidation) SetDropList(keys []string) error {
-	formula := "\"" + strings.Join(keys, ",") + "\""
-	buf := bytes.NewBuffer(nil)
-	if err := xml.EscapeText(buf, []byte(formula)); err != nil {
-		return err
-	}
-	formula = buf.String()
-
+	formula := `"` + formulaEscaper.Replace(strings.Join(keys, ",")) + `"`
 	if dataValidationFormulaStrLen < len(formula) {
 		return fmt.Errorf(dataValidationFormulaStrLenErr)
 	}
--- a/datavalidation.go
+++ b/datavalidation.go
@@ -12,6 +12,8 @@
 package excelize
 
 import (
+	"bytes"
+	"encoding/xml"
 	"fmt"
 	"strings"
 )
@@ -111,6 +113,12 @@ func (dd *DataValidation) SetInput(title, msg string) {
 // SetDropList data validation list.
 func (dd *DataValidation) SetDropList(keys []string) error {
 	formula := "\"" + strings.Join(keys, ",") + "\""
+	buf := bytes.NewBuffer(nil)
+	if err := xml.EscapeText(buf, []byte(formula)); err != nil {
+		return err
+	}
+	formula = buf.String()
+
 	if dataValidationFormulaStrLen < len(formula) {
 		return fmt.Errorf(dataValidationFormulaStrLenErr)
 	}
--- a/datavalidation.go
+++ b/datavalidation.go
@@ -132,14 +132,8 @@ func (dd *DataValidation) SetDropList(keys []string) error {
 
 // SetRange provides function to set data validation range in drop list.
 func (dd *DataValidation) SetRange(f1, f2 float64, t DataValidationType, o DataValidationOperator) error {
-	formula1 := fmt.Sprintf("%f", f1)
-	formula2 := fmt.Sprintf("%f", f2)
-	if dataValidationFormulaStrLen+21 < len(dd.Formula1) || dataValidationFormulaStrLen+21 < len(dd.Formula2) {
-		return fmt.Errorf(dataValidationFormulaStrLenErr)
-	}
-
-	dd.Formula1 = fmt.Sprintf("<formula1>%s</formula1>", formula1)
-	dd.Formula2 = fmt.Sprintf("<formula2>%s</formula2>", formula2)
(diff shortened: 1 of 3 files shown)


Arnie97:
Other thoughts about these issues:

* I've noticed existing usages of `xml.EscapeText()` in [`stream.go`](https://github.com/qax-os/excelize/blob/v2.4.0/stream.go#L400-L409), but I did not further investigate into the code to check whether their use case were compatible with Microsoft Excel, so they were left untouched.

* I've noticed and removed some hard code integer literals `21` from the length validation code, but haven't tested this change yet.  I guess it stands for:

```go
len("<formula1></formula1>")
```

* AFAIK there's no way to escape the separator commas `,` in the validation list.  However I did not add an assertion for them either, in order to avoid breaking any existing projects, which might invoke the excelize library in the following way:

```go
const choices = "Alpha,Beta,Gamma"
...
dv.SetDropList([]string{choices})
...
```

xuri:
Thanks for your PR. I have fixed it by commit [`7dbf88f`](https://github.com/qax-os/excelize/commit/7dbf88f221f278075d4ff9e153b21236d0826c33)
. Make formula1 and formula as string data type instead of the `formulaEscaper`, we can let the XML encoder escape characters.

xuri:
Thanks for your PR. I have fixed it by commit [`7dbf88f`](https://github.com/xuri/excelize/commit/7dbf88f221f278075d4ff9e153b21236d0826c33)
. Make `formula1` and `formula2` as string data type instead of the `formulaEscaper`, we can let the XML encoder escape characters.

xuri:
Note that the value of `formula` in the drop list should have a quote, overwise the rule will be invalid. I closed the PR.

Arnie97:
@xuri While your patch looks much cleaner, the XML encoder in the standard library is NOT compatible with Microsoft Excel.
May you please read the "Motivation and Context" section above? With the current master branch 7dbf88f, the result generated for my example test case `A<,B>,C",D	,E',F` is still reported as a corrupt file.

xuri:
> @xuri While your patch looks much cleaner, the XML encoder in the standard library is NOT compatible with Microsoft Excel.
> May you please read the "Motivation and Context" section above? With the current master branch [7dbf88f](https://github.com/xuri/excelize/commit/7dbf88f221f278075d4ff9e153b21236d0826c33), the result generated for my example test case `A<,B>,C",D ,E',F` is still reported as a corrupt file.

Yeah, thanks for your PR, I have added error messages and minimum range checking base on this branch.

Issue #269, opened 2018-09-04
Write text


Davidiasdesousa:
How i can write text with break line.

xuri:
Thanks for your issue, reference issue #69 (set multiline texts to cells support), #100 and #172. I have added this feature request to the Roadmap.

xuri:
I close this and merge it into issue #172 is still discussing this problem.

xuri:
Hi @Davidiasdesousa, sorry for my late reply. I have added rich text support for this lib, now you set cell with rich text by the [`SetCellRichText`](https://github.com/360EntSecGroup-Skylar/excelize/blob/66d0272f6af59b5f0c97a304379a795420a43e8b/cell.go#L460) function.

Issue #69, opened 2017-06-29
There is a little bug when writing multiline texts to cells.


zzc-tongji:
**Test Code**

```go
package main

import (
	"fmt"

	"github.com/Luxurioust/excelize"
)

func main() {
	// read
	xlsx, err := excelize.OpenFile("test-before.xlsx")
	if err != nil {
		fmt.Printf("%s\n", err.Error())
		return
	}
	table := xlsx.GetRows("sheet1")
	fmt.Printf("\n")
	fmt.Printf("Before:\n")
	for i := 0; i < len(table); i++ {
		for j := 0; j < len(table[0]); j++ {
			fmt.Printf("%s ", table[i][j])
		}
		fmt.Printf("\n")
	}
	// change
	xlsx.SetCellValue("sheet1", "A2", "A2\nExample 2")
	xlsx.SetCellValue("sheet1", "B2", "A2\n例子2")
	// write
	xlsx.SaveAs("test-after.xlsx")
	// read again
	xlsx, err = excelize.OpenFile("test-after.xlsx")
	if err != nil {
		fmt.Printf("%s\n", err.Error())
		return
	}
	table = xlsx.GetRows("sheet1")
	fmt.Printf("\n")
	fmt.Printf("After:\n")
	for i := 0; i < len(table); i++ {
		for j := 0; j < len(table[0]); j++ {
			fmt.Printf("%s ", table[i][j])
		}
		fmt.Printf("\n")
	}
	fmt.Printf("\n")
}
```

**Test File**

[test-before.xlsx](https://github.com/Luxurioust/excelize/files/1110550/test-before.xlsx)
![test-before](https://user-images.githubusercontent.com/20331014/27666036-8e32ad26-5ca3-11e7-9e0d-89f3a04f83ea.png)

[test-after.xlsx](https://github.com/Luxurioust/excelize/files/1110551/test-after.xlsx)
![test-after](https://user-images.githubusercontent.com/20331014/27666038-9739d0e8-5ca3-11e7-94af-e5651a27baa5.png)

**Test Environment**

- golang 1.8 darwin/amd64
- Microsoft Excel 2016 for Mac

**Problem**

It seems like excelize can parse multiline texts in cells (with '\n' in the middle). However, multiline texts written by excelize could not be displayed correctly by Microsoft Excel, though they can be parsed correctly by excelize again.

zzc-tongji:
I got it.

[...]

PR #976, opened 2021-07-28
Preserve whitespace in cells

Arnie97:
# PR Details

## Description
This patch backports the more reasonable whitespace detection condition from `SetCellRichText()` to `setSharedString()`.

## Related Issue
#69, #269

## Motivation and Context

A relevant formatting fix was first introduced in `setSharedString()` in dc012645. Unfortunately, a strict condition was applied, so the patch only fixed formatting for the cells with prefix and trailing spaces, not the ones with line ending characters:

```go
t := xlsxT{Val: val}
// Leading and ending space(s) character detection.
if len(val) > 0 && (val[0] == 32 || val[len(val)-1] == 32) {
	ns := xml.Attr{
		Name:  xml.Name{Space: NameSpaceXML, Local: "space"},
		Value: "preserve",
	}
	t.Space = ns
}
```

## How Has This Been Tested

* Microsoft Office 365 for Mac, 16.49 (21050901)
* Microsoft Office 2019 Professional Plus for Windows, 2105 (Build 14026.20302)
* Microsoft Office 2013 Professional Plus for Windows, 15.0.4919.1002

## Types of changes

## Checklist

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -308,14 +308,14 @@ func (f *File) setSharedString(val string) int {
 	sst.Count++
 	sst.UniqueCount++
 	val = bstrMarshal(val)
-	t := xlsxT{Val: val}
-	// Leading and ending space(s) character detection.
-	if len(val) > 0 && (val[0] == 32 || val[len(val)-1] == 32) {
-		ns := xml.Attr{
+	t := xlsxT{
+		Val: val,
+	}
+	if strings.ContainsAny(val, "\r\n ") {
+		t.Space = xml.Attr{
 			Name:  xml.Name{Space: NameSpaceXML, Local: "space"},
 			Value: "preserve",
 		}
-		t.Space = ns
 	}
 	sst.SI = append(sst.SI, xlsxSI{T: &t})
 	f.sharedStringsMap[val] = sst.UniqueCount - 1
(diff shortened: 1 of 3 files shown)


xuri:
Thanks for your PR. I have added the maximum character limit and escaped string literal conversion based on your PR, and only check prefix and suffix characters instead of  `strings.ContainsAny` will get better performance. This patch will be released in the next version.

Arnie97:
I noticed that `strings.ContainsAny()` have visible impact on performance, but unfortunately...

1. The new patch did not fix the issue.  As clarified in #69, I would expect `"first line\nsecond line"` to be displayed correctly.  I'm still looking for better implementations that keeps the *`O(1)`* time complexity for a single cell though.
2. A much more critical factor is affecting the performance of `SetCellRichText()`, and that was my original motivation for this pull request. I've opened a separate issue for further discussions.

xuri:
Note that don't forget to set the wrap text style for the cell with multi-line cell, for example

```go
f := excelize.NewFile()
styleID, err := f.NewStyle(&excelize.Style{Alignment: &excelize.Alignment{WrapText: true}})
if err != nil {
    fmt.Println(err)
}
if err := f.SetCellStyle("Sheet1", "A1", "A1", styleID); err != nil {
    fmt.Println(err)
}
if err := f.SetCellValue("Sheet1", "A1", "first line\nsecond line"); err != nil {
    fmt.Println(err)
}
```

Arnie97:
Yes, you're right.  I've found that the `WrapText: true` style was later overwritten with another custom style in my code. Apologize for the misleading reply.

PR #1023, opened 2021-09-22
passing excelize.File by value 

source-transformer:
# PR Details

## Description

passing excelize.File by value will result in the caller getting shared-string-table index returned from the cell being queried (when setting strings in cells). Effectively a new instance of the shared strings table will be created.

## Motivation and Context

You can't pass excelize.File by value and update cells containing strings

## How Has This Been Tested

Added unit test to show how the behavior is now correct

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x ] My code follows the code style of this project.
- [ x] I have read the **CONTRIBUTING** document.
- [ x] I have added tests to cover my changes.
- [ x] All new and existing tests passed.

### DIFF
--- a/file.go
+++ b/file.go
@@ -52,6 +52,10 @@ func NewFile() *File {
 	ws, _ := f.workSheetReader("Sheet1")
 	f.Sheet.Store("xl/worksheets/sheet1.xml", ws)
 	f.Theme = f.themeReader()
+
+	f.SharedStrings = &xlsxSST{}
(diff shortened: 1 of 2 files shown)


xuri:
Thanks for your PR, why you need to passing excelize.File by value and the unit test is failed. I think this change is not required.

xuri:
We still don't have enough info on the code and response to this pull request, so I'll close this, if you have any questions, please let me know to reopen this anytime.

PR #1049, opened 2021-11-03
return immediately when matched for efficiency

seanliang:
# PR Details
Several functions in sheet.go loop all the sheets even the condition is met

## Description
Return immediately when matched

## Motivation and Context
For efficiency

## Types of changes
Bug fix (non-breaking change which fixes an issue)


xuri:
LGTM, thanks for your PR.

PR #1115, opened 2022-01-05
feat: implement SHA-512 algorithm to ProtectSheet

Jonham:
# implement SHA-512 algorithm to protectSheet and add ValidateSheetProtect

## Description

add genISOPasswordHash, which implements SHA-512 algorithm salt hash generation that supports workbookProtection and worksheetProtection.

## Motivation and Context

Excel 2007 supports hash password. Since Excel 2013, SHA-512 algorithm are the default method in creating workbook protection and worksheet protection. Although worksheet protection.password works well with `genSheetPasswd` (Excel 2013, 2019 both support), workbook protection only support SHA-512. And next I will implement protectWorkbook.

## How Has This Been Tested

I had set password to workbook protection and worksheet protection on file `test/Worksheet-Workbook-Protection.xlsx` with Mac Excel 2019.
unzip it and read both protection xml lines to create `func TestGenISOPasswordHash` test1 and test2.
test3 just run the algorithm twice to test whether hash value match.

## Types of changes

## Checklist

  > no document about ProtectSheet


crypt.go
```diff
@@ -127,6 +129,105 @@ type StandardEncryptionVerifier struct {
...
+	// pseudorandom function. When verifying a password, the salt value retrieved from the
+	// document MUST be used. The salt MUST NOT be larger than 65,536 bytes.
+	SaltValue string
+	// default: 100000 1e5. A 32-bit unsigned integer that specifies the number of
+	// times to iterate on a hash of a password. It MUST NOT be greater than 10,000,000.
+	SpinCount int
+}
+
+/*
+genISOPasswordHash implements "ISO password hashing algorithm"
```
xuri, line 156: Just code style suggestions: using the `//` comment delimiter comment out document block.

crypt_test.go
```diff
@@ -36,3 +36,61 @@ func TestEncryptionMechanism(t *testing.T) {
 func TestHashing(t *testing.T) {
 	assert.Equal(t, hashing("unsupportHashAlgorithm", []byte{}), []uint8([]byte(nil)))
 }
+
+func TestGenISOPasswordHash(t *testing.T) {
+	/*
```
xuri, line 41: I think this example can be removed, developer's can be determined fields would be like by data structure definition.

sheet.go
```diff
@@ -1167,8 +1176,31 @@ func (f *File) ProtectSheet(sheet string, settings *FormatSheetProtection) error
 		Sheet:               true,
 		Sort:                settings.Sort,
 	}
+
 	if settings.Password != "" {
-		ws.SheetProtection.Password = genSheetPasswd(settings.Password)
+		// settings.AlgorithmName == "SHA-512"
+		if settings.AlgorithmName == hashAlgorithmSHA512 {
+			err, result := genISOPasswordHash(saltHashData{
```
xuri, line 1183: I suggest that make `genSheetPasswd` support password generation with the different crypto algorithm, and it can be called in `ProtectSheet` and `UnprotectSheet`, since that password generation should be implemented in the same scope, the caller just as a wrapper.

sheet.go
```diff
@@ -1183,6 +1215,35 @@ func (f *File) UnprotectSheet(sheet string) error {
 	return err
 }
 
+// ValidateSheetProtect provides a function to validate worksheet protect password.
+func (f *File) ValidateSheetProtect(sheet string, password string) (error, bool) {
```
xuri, line 1219: I propose to add the second optional parameter for the existing function `UnprotectSheet` (semantics of unlocking with password) instead of adding a new exported function `ValidateSheetProtect`. If the user-specified the second parameter is the wrong password, the error will be returned, if no second parameter has been specified, it will be forced to remove worksheet protection.

xuri (changes requested):

sheet_test.go
```diff
@@ -477,3 +477,99 @@ func newSheetWithSave() {
...
+	sheetName := f.GetSheetName(0)
+	err := f.ProtectSheet(sheetName, &FormatSheetProtection{
+		AlgorithmName: "SHA-128",
+		Password:      password,
+	})
+
+	assert.EqualError(t, err, "Algorithm NOT implement")
+}
+
+func TestValidateSheetProtect(t *testing.T) {
```
xuri, line 542: Some unit test has been skipped, please make sure the test has been all finished and the code coverage has covered changes.

xuri (changes requested):
Remove the test spreadsheet attachment `test/Worksheet-Workbook-Protection.xlsx`, just test and assert with XML snippets is enough.

sheet.go
```diff
@@ -1167,8 +1176,31 @@ func (f *File) ProtectSheet(sheet string, settings *FormatSheetProtection) error
...
+				return err
+			}
+			ws.SheetProtection.Password = ""
+
+			ws.SheetProtection.AlgorithmName = hashAlgorithmSHA512
+			ws.SheetProtection.SaltValue = result.SaltValue
+			ws.SheetProtection.HashValue = result.HashValue
+			ws.SheetProtection.SpinCount = result.SpinCount
+		} else if settings.AlgorithmName == "" {
+			// SHA-1 password
```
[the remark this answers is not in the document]
Jonham: OK，I mistake

sheet.go
```diff
@@ -1167,8 +1176,31 @@ func (f *File) ProtectSheet(sheet string, settings *FormatSheetProtection) error
...
+			ws.SheetProtection.Password = genSheetPasswd(settings.Password)
+		} else {
+			return errors.New("Algorithm NOT implement")
```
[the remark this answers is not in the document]
Jonham: Currently I have found no more protection algorithm used on sheetProtection or bookProtection. Anyone need more algorithm, please issue and I would like to add.

sheet.go
Jonham: `genSheetPasswd` only requires one password param generate one XOR result password, but `genISOPasswordHash` requires 3(AlgorithmName, password, SpinCount) and return 4 (algorithmName, saltValue, hashValue and spinCount). I think it will be more complicated when these two method become one. Let's keep it separated entil more new algorithm need to be implemented.

sheet.go
Jonham: I wil rewrite this.

sheet_test.go
Jonham: I've rewrite unit test case.

Jonham:
I've add some test case and rewrite follow review comments.

Jonham:
sorry for my carelessness. I will run `go test` and `go test -coverprofile=cover.out` before commit next time.

xuri:
I made some changes based on your branch.

PR #1138, opened 2022-02-04
Add Showruler to Sheetview

JDavidVR:
This adds showRuler="false" to hide ruler from sheetView

```
<sheetViews>
<sheetView workbookViewId="0" zoomScaleNormal="100" view="pageLayout" showRuler="false" tabSelected="1" showRowColHeaders="0" showGridLines="0"/>
</sheetViews>
```

image

### DIFF
--- a/sheetview.go
+++ b/sheetview.go
@@ -62,7 +62,10 @@ type (
 	// how sheet is displayed, by default it uses empty string
 	// available options: pageLayout, pageBreakPreview
 	View string
-	
+	// ShowRuler is a SheetViewOption. It specifies a flag indicating
+	// this sheet should display ruler.
+	ShowRuler bool
+
 	/* TODO
 	// ShowWhiteSpace is a SheetViewOption. It specifies a flag indicating
 	// whether page layout view shall display margins. False means do not display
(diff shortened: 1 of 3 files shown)


xmlWorksheet.go
```diff
@@ -204,6 +204,7 @@ type xlsxSheetView struct {
 	WorkbookViewID           int              `xml:"workbookViewId,attr"`
 	Pane                     *xlsxPane        `xml:"pane,omitempty"`
 	Selection                []*xlsxSelection `xml:"selection"`
+	ShowRuler                *bool            `xml:"showRuler,attr,omitempty"`
```
xuri, line 207: Newly added field `showRuler` should be following order in [OOXML Spec](https://github.com/sc34wg4/OOXMLSchemas/blob/5d7db375ff5f5ecfd38d39d088a6b81a6c41e39e/Part1/OfficeOpenXML-XMLSchema-Strict/sml.xsd#L2315-L2341) to avoid generating file corrupted with compatibility issue.

xuri (changes requested):
Thanks for your PR. I've left a comment.

sheetview.go
```diff
@@ -62,7 +62,10 @@ type (
 	// how sheet is displayed, by default it uses empty string
 	// available options: pageLayout, pageBreakPreview
```
xuri, line 63: available options: pageLayout, pageBreakPreview -> available options:  normal, pageLayout, pageBreakPreview

JDavidVR: Done!

xmlWorksheet.go
JDavidVR: Done!

xuri:
LGTM, thank you very much.

PR #1149, opened 2022-02-16
improve-vba-thisworkbook-compatibility

JDavidVR:
please refer to this issue: https://github.com/qax-os/excelize/issues/1148 it has everything documented


xuri:
Thanks for your pull request. Because the default code name `ThisWorkbook` can be changed in the Visual Basic Editor, we need to provide a function to allow users to change it. I have added two exported functions `SetWorkbookPrOptions` and `GetWorkbookPrOptions` to support setting and getting the code name property of the workbook. I'll close this PR. Now you can set the codename of the workbook like this: 

```go
err := f.SetWorkbookPrOptions(excelize.CodeName("ThisWorkbook"))
```

PR #1154, opened 2022-02-17
add-FilterPrivacy-to-workbookPr

JDavidVR:
more information on this ticket: https://github.com/qax-os/excelize/issues/1152

### DIFF
--- a/workbook.go
+++ b/workbook.go
@@ -123,6 +123,11 @@ func (o CodeName) setWorkbookPrOption(pr *xlsxWorkbookPr) {
 	pr.CodeName = string(o)
 }
 
+// setWorkbookPrOption implements the WorkbookPrOption interface.
+func (o FilterPrivacy) setWorkbookPrOption(pr *xlsxWorkbookPr) {
+	pr.FilterPrivacy = bool(o)
+}
+
 // GetWorkbookPrOptions provides a function to gets workbook properties.
 //
 // Available options:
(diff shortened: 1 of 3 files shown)


workbook.go
```diff
@@ -123,6 +123,11 @@ func (o CodeName) setWorkbookPrOption(pr *xlsxWorkbookPr) {
 	pr.CodeName = string(o)
 }
 
+// setWorkbookPrOption implements the WorkbookPrOption interface.
```
xuri, line 126: I suggest the get and set implement functions following the order in [OOXML Spec](https://github.com/sc34wg4/OOXMLSchemas/blob/5d7db375ff5f5ecfd38d39d088a6b81a6c41e39e/Part1/OfficeOpenXML-XMLSchema-Strict/sml.xsd#L4200-L4220), move this function before `CodeName`, and update the available options document of the functions `SetWorkbookPrOptions` and `GetWorkbookPrOptions`

workbook.go
```diff
@@ -145,3 +150,13 @@ func (o *CodeName) getWorkbookPrOption(pr *xlsxWorkbookPr) {
 	}
 	*o = CodeName(pr.CodeName)
 }
+
+// getWorkbookPrOption implements the WorkbookPrOption interface and get the
+// filter privacy of thw workbook.
+func (o *FilterPrivacy) getWorkbookPrOption(pr *xlsxWorkbookPr) {
+	if pr == nil {
+		*o = true
```
xuri, line 158: This case was not been tested, please add a test case for this.

xuri (changes requested):
Thanks for your contribution. I've left some comments.

PR #1248, opened 2022-06-10
adds the ability to open online files to the function OpenFile

ahah43:
# PR Details

adds the ability to open online files to the function OpenFile
## Description

The given "file name" passed to the OpenFile function is to be checked first if it is an url. If so, the file at that address is loaded, else, the old behaviour of treating "file name" as a path continuous.
## Types of changes

## Checklist

### DIFF
--- a/excelize.go
+++ b/excelize.go
@@ -25,7 +25,8 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-
+	"net/http" 			// https://github.com/ahah43
+"net/url"				// https://github.com/ahah43
 	"golang.org/x/net/html/charset"
 )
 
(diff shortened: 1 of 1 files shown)


excelize.go
```diff
@@ -25,7 +25,8 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-
+	"net/http" 			// https://github.com/ahah43
+"net/url"				// https://github.com/ahah43
```
xuri, line 29: The code style should be following the [Contributing Guidelines](https://github.com/xuri/excelize/blob/master/CONTRIBUTING.md), it should be formatted, and remove the comments.

xuri (commented):
Thanks for your pull request, the [`OpenReader`](https://xuri.me/excelize/en/workbook.html#OpenReader) function can be read data stream from `io.Reader`. Fetch, or receive files from the network by HTTP or other protocols was not in the scope of this library, so I don't suggest to add this feature.

Issue #1267, opened 2022-07-02
Several unhandled errors


MJacred:
**Description**

Code has multiple unhandled errors and just continues running, risking potential crashes, ignoring bad parameters, etc. instead of checking and handling or returning the error.

PR coming up right away - open to feedback.

PR #1268, opened 2022-07-02
WIP: Handle and return more errors

MJacred:
# PR Details

## Description

> Several errors are never checked against

Changes regarding error handling
* some getter funcs now just return a pointer, there are new `New*()` funcs to handle initialization; full list:
  * `NewCalcChainReader()`
  * `NewContentTypesReader()`
  * `NewStylesReader()`
  * `NewThemeReader()`
  * `NewWorkbookReader()`
  * NOTE: they are exported for now, but can un-export them if requested
  * The reason they were separated in getters and `New*()` is to prevent more funcs that return errors where it's not necessary. Especially as those funcs are used pretty heavily
* in some cases, no error was returned, because existing code already handled error -> new code in same area of said funcs respect the existing handling
* in some cases, it seems errors should be fine, in those cases I want to at least return the first error (see `firstError`). And if an error occurred, handle it using `continue` or whatever approach fits best the respective context

Minor changes picked up on:
* in NewSheet(), don't call DeleteSheet() -> not needed, because if sheet exists, it is returned
* in OpenReader() also set: ContentTypes, WorkBook

## Related Issue

#1267

## Motivation and Context

* fixing potential crashes
* enable library user to react and potentially handle errors
  * currently, some funcs fail silently or could crash

## How Has This Been Tested

* running package tests locally (using golang version 1.16)

## Types of changes

  - because of new error returns

## Checklist

  - due to stricter error handling, some tests are currently failing
- [ ] cleanup commits once everything else is fine: Sign the work, commit message lingua
- [ ] in code comments, add info regarding `firstError` returned

## Misc

Open to feedback

### DIFF
--- a/adjust_test.go
+++ b/adjust_test.go
@@ -7,7 +7,8 @@ import (
 )
 
 func TestAdjustMergeCells(t *testing.T) {
-	f := NewFile()
+	f, err := NewFile()
+	assert.NoError(t, err)
 	// testing adjustAutoFilter with illegal cell coordinates.
 	assert.EqualError(t, f.adjustMergeCells(&xlsxWorksheet{
 		MergeCells: &xlsxMergeCells{
@@ -264,7 +265,8 @@ func TestAdjustMergeCells(t *testing.T) {
 		assert.Equal(t, 0, len(c.ws.MergeCells.Cells), c.label)
 	}
 
-	f = NewFile()
+	f, err = NewFile()
+	assert.NoError(t, err)
 	p1, p2 := f.adjustMergeCellsHelper(2, 1, 0, 0)
 	assert.Equal(t, 1, p1)
 	assert.Equal(t, 2, p2)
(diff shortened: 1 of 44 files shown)


xuri (changes requested):
Thanks for your PR. This PR contains a lot of code and I need some time to review. The failure unit test case should be fixed.

MJacred:
Hi,
Yes, I'm working on the failing unit tests.  

Would appreciate help with errors
* `error in sheet "Sheet2": overlapRange: invalid range length: range of "A6" is smaller than minimum range length of 2`
* `xml decode error: XML syntax error on line 1: invalid UTF-8`

MJacred:
@xuri:

ok, it seems Book1.xlsx had cell A6 flagged as "merged cell". A cell merge of one cell makes no sense  
Therefore, I deleted cell A6 in the excel

or was that intentional? Guessing because of this line

```
value, err = f.GetCellValue("Sheet2", "A6") // Merged cell ref is single coordinate.
```

If it was intentional, it would make sense to delete that in the code before saving.

PR #1272, opened 2022-07-06
Improve column number error handling

MJacred:
# PR Details

## Description

changes
* rename `TotalColumns` to `MaxColumns`
* introduce its minimum value counterpart `MinColumns`
* make some error texts more conform ("greater/less than or equal to")

## Types of changes

## Checklist

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -61,7 +61,7 @@ func newInvalidStyleID(styleID int) error {
 // newFieldLengthError defined the error message on receiving the field length
 // overflow.
 func newFieldLengthError(name string) error {
-	return fmt.Errorf("field %s must be less or equal than 255 characters", name)
+	return fmt.Errorf("field %s must be less than or equal to 255 characters", name)
 }
 
 // newCellNameToCoordinatesError defined the error message on converts
@@ -76,10 +76,10 @@ var (
 	ErrStreamSetColWidth = errors.New("must call the SetColWidth function before the SetRow function")
 	// ErrColumnNumber defined the error message on receive an invalid column
 	// number.
-	ErrColumnNumber = errors.New("column number exceeds maximum limit")
+	ErrColumnNumber = fmt.Errorf(`the column number must be greater than or equal to %d and less than or equal to %d`, MinColumns, MaxColumns)
 	// ErrColumnWidth defined the error message on receive an invalid column
 	// width.
-	ErrColumnWidth = fmt.Errorf("the width of the column must be smaller than or equal to %d characters", MaxColumnWidth)
+	ErrColumnWidth = fmt.Errorf("the width of the column must be less than or equal to %d characters", MaxColumnWidth)
 	// ErrOutlineLevel defined the error message on receive an invalid outline
 	// level number.
 	ErrOutlineLevel = errors.New("invalid outline level")
@@ -102,7 +102,7 @@ var (
 	ErrMaxRows = errors.New("row number exceeds maximum limit")
 	// ErrMaxRowHeight defined the error message on receive an invalid row
 	// height.
-	ErrMaxRowHeight = fmt.Errorf("the height of the row must be smaller than or equal to %d points", MaxRowHeight)
+	ErrMaxRowHeight = fmt.Errorf("the height of the row must be less than or equal to %d points", MaxRowHeight)
 	// ErrImgExt defined the error message on receive an unsupported image
 	// extension.
 	ErrImgExt = errors.New("unsupported image extension")
@@ -143,7 +143,7 @@ var (
 	ErrCustomNumFmt = errors.New("custom number format can not be empty")
 	// ErrFontLength defined the error message on the length of the font
 	// family name overflow.
-	ErrFontLength = fmt.Errorf("the length of the font family name must be smaller than or equal to %d", MaxFontFamilyLength)
+	ErrFontLength = fmt.Errorf("the length of the font family name must be less than or equal to %d", MaxFontFamilyLength)
 	// ErrFontSize defined the error message on the size of the font is invalid.
 	ErrFontSize = fmt.Errorf("font size must be between %d and %d points", MinFontSize, MaxFontSize)
 	// ErrSheetIdx defined the error message on receive the invalid worksheet
--- a/lib.go
+++ b/lib.go
@@ -211,7 +211,7 @@ func ColumnNameToNumber(name string) (int, error) {
 		}
 		multi *= 26
 	}
-	if col > TotalColumns {
+	if col > MaxColumns {
 		return -1, ErrColumnNumber
 	}
 	return col, nil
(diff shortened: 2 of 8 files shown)


MJacred:
how should I add unit tests?

change existing unit tests to switch between both SetColWidth*() funcs?

e.g.

```golang
func TestColWidth(t *testing.T) {
	f := NewFile()
	assert.NoError(t, f.SetColWidth("Sheet1", "B", "A", 12))
	assert.NoError(t, f.SetColWidth("Sheet1", "A", "B", 12))
	…
}
```

becomes

```golang
func TestColWidth(t *testing.T) {
	f := NewFile()
	assert.NoError(t, f.SetColWidth("Sheet1", "B", "A", 12))
	assert.NoError(t, f.SetColWidth("Sheet1", 1, 2, 12))
	…
}
```

?

xuri (commented):
Thanks for your PR. I don't suggest adding this new exported function, the library has provided the [`ColumnNumberToName`](https://pkg.go.dev/github.com/xuri/excelize/v2#ColumnNumberToName) function to convert the integer to an Excel sheet column name, the developer can use this to get the column name and then set the width of the column, and there are two kinds of cell reference styles A1 and R1C1 in the Excel, we needn't provides two exported functions to these two kinds styles in one feature, to keep the library simplify and tiny.

errors.go
```diff
@@ -70,13 +70,27 @@ func newCellNameToCoordinatesError(cell string, err error) error {
 	return fmt.Errorf("cannot convert cell %q to coordinates: %v", cell, err)
 }
 
+// newColumnNumberError defined the error message on receiving invalid column
+// number.
+func newColumnNumberError(count int) error {
```
xuri, line 75: Use the `col` instead of `count` for better semantic, because this function is only used internally, it can be simplified like this:

```go
func newColumnNumberError(col int) error {
    if col < MinColumns {
        return fmt.Errorf("column number less than the minimum limit %d", MinColumns)
    }
    return fmt.Errorf("column number exceeds the maximum limit %d", MaxColumns)
}
```

errors.go
```diff
@@ -70,13 +70,27 @@ func newCellNameToCoordinatesError(cell string, err error) error {
...
+}
+
 var (
 	// ErrStreamSetColWidth defined the error message on set column width in
 	// stream writing mode.
 	ErrStreamSetColWidth = errors.New("must call the SetColWidth function before the SetRow function")
 	// ErrColumnNumber defined the error message on receive an invalid column
 	// number.
-	ErrColumnNumber = errors.New("column number exceeds maximum limit")
+	ErrColumnNumber = errors.New("invalid column number")
```
xuri, line 93: I suggest removing this exported error variable and using the `newColumnNumberError` instead of it.

lib.go
```diff
@@ -211,8 +211,8 @@ func ColumnNameToNumber(name string) (int, error) {
 		}
 		multi *= 26
 	}
-	if col > TotalColumns {
-		return -1, ErrColumnNumber
+	if col < MinColumns || col > TotalColumns {
```
xuri, line 214: The value of col won't be 0, so needn't the `col < MinColumns` statement.

xuri (changes requested):
Alright, I've left some comments, and I'd like to accept this patch.

MJacred:
I applied your suggestions (though the constant is currently called `TotalColumns`, but I can rename it).

Though now you cannot make an `assert.ErrorIs()` check anymore in the tests or an `errors.Is()` in actual app code.

errors.go
```diff
@@ -70,13 +70,19 @@ func newCellNameToCoordinatesError(cell string, err error) error {
 	return fmt.Errorf("cannot convert cell %q to coordinates: %v", cell, err)
 }
 
+// newColumnNumberError defined the error message on receiving invalid column
```
xuri, line 73: I agree with you. We need to declare two exported error variable `ErrMinColumnNumber` and `ErrMaxColumnNumber`, remove the `newColumnNumberError` function, and rename the `TotalColumns` to `MaxColumns`.

```go
    // ErrMaxColumnNumber defined the error message on receive a column
    // number exceeds maximum limit.
    ErrMaxColumnNumber = fmt.Errorf("column number exceeds the maximum limit %d", MaxColumns)
    // ErrMinColumnNumber defined the error message on receive a column
    // number less than the minimum limit.
    ErrMinColumnNumber = fmt.Errorf("column number less than the minimum limit %d", MinColumns)
```

lib.go
```diff
@@ -225,11 +225,8 @@ func ColumnNameToNumber(name string) (int, error) {
 //     excelize.ColumnNumberToName(37) // returns "AK", nil
 //
 func ColumnNumberToName(num int) (string, error) {
-	if num < 1 {
-		return "", fmt.Errorf("incorrect column number %d", num)
-	}
-	if num > TotalColumns {
-		return "", ErrColumnNumber
+	if num < MinColumns || num > TotalColumns {
```
xuri, line 228: Separate these two condition statement.

xuri (approved):

errors.go
MJacred: Hm… while having 2 error types is the most accurate, in normal usage just knowing that the column number is wrong (as a type) is sufficient - imho. And therefore makes checking for library user easier - if they want to handle the column number error.  
We can add extra detail using a func like in the original PR version to tell if greater than MAX or less than MIN:

```golang
ErrColumnNumber = errors.New("invalid column number")

func newColumnNumberError(count int) error {
    var detail string
	if count < MinColumns {
		detail = fmt.Sprintf("%d is less than minimum limit of 1 columns", count)
	} else if count > TotalColumns {
		detail = fmt.Sprintf("%d exceeds maximum limit of %d columns", count, TotalColumns)
	} else {
		detail = "internal error: incorrect error usage"
	}
	return fmt.Errorf("%w: %s", ErrColumnNumber, detail)
}
```

And if the library user wants further error inspection, they can check against exported `MaxColumns`/`MinColumns` constants to do special handling on their side, if any at all.  
2 errors make checking more elaborate…

Or do you want to stick with the 2 error types?

xuri: I suggest using two error types and letting the user know which kind of error directly in details. Do you thinks so?

MJacred: Do you mean sth. like this?

```golang
ErrColumnNumber = errors.New("column number out of bounds")
ErrMaxColumnNumber = fmt.Errorf("column number exceeds the maximum limit %d", MaxColumns)
ErrMinColumnNumber = fmt.Errorf("column number less than the minimum limit %d", MinColumns)

func newColumnNumberError(count int) error {
    var detail error
	if count < MinColumns {
		detail = ErrMinColumnNumber
	} else {
		detail = ErrMaxColumnNumber
	}
	return fmt.Errorf("%w: %w", ErrColumnNumber, detail)
}
```

If you did think of this, then yes, that would be nice. But sadly impossible, as golang currently does not allow multi-wrapping errors like this…

If you meant sth. else, then I'm lost

xuri (changes requested):
I agree with you, and I suggest making following changes

- Rename exported constant `TotalColumns` to `MaxColumn`
- Still use the original exported error variable `ErrColumnNumber`, but change the error message to:
```go
fmt.Errorf(`the column number must be greater than %d and smaller than %d`, MinColumns, MaxColumns)
```
- No need the `newColumnNumberError` function

lib.go
MJacred: is still still valid?

xuri (approved):

PR #1282, opened 2022-07-20
1881 import followup

rodoard:
f.GetCellFormula returns SI
Optimize f.GetTotalRows


rodoard:
whoops wrong repo sorry

Issue #1283, opened 2022-07-21
tabColor: support options other than RGB


thomascharbonnel:
**Description**

Related to #785. I've noticed the tab color feature only supports RGB (read and write) currently. Do you think it would be okay to add support for the other types of tab colors (eg. "theme", "indexed")? I should be able to pick up this task (and the other one we discussed earlier) in the coming few weeks.

xuri:
Thanks for your issue. Contributions are welcome. I think we need to make following works for this feature:

- Add new options `TabColorIndexed`, `TabColorTheme`, `TabColorTint` for the `SheetPrOption`
- Implement setter and getter for these options
- Add unit test for this feature

thomascharbonnel:
Thanks for the help!
I've written some code to support TabColorTheme first, does this change make sense? https://github.com/thomascharbonnel/excelize/commit/2e9ea923188bb71e226776a11e96d491fe0c7c81
I'm not sure what to do when `theme` is unset, as Golang's default int value of 0 is a valid theme value. I've chosen -1 to indicate that theme is unset. What do you think?

xuri:
Okay, I agree with you.

thomascharbonnel:
Thank you for the feedback, i'll keep working on this next week.
I think i'll make the theme an enum (list of possible values: https://docs.microsoft.com/en-us/dotnet/api/documentformat.openxml.drawing.colorscheme?view=openxml-2.8.1), this way `-1` will be a little bit cleaner rather than a magic number.
One strange thing is that Excel does not seem to follow the spec: https://stackoverflow.com/a/2761369

xuri:
Okay, make the `TabColorTheme` accept theme enumeration, and returns the theme index as is. In Excel the first four theme color index was different with spec. We need to make a patch for it when getting the theme-based color. For example: [get the cell background color](https://xuri.me/excelize/en/utils.html#ThemeColor).

PR #1285, opened 2022-07-25
add support for more tabColor options and split RGB #1283

thomascharbonnel:
This commit renames `TabColor` into `TabColorRGB` (but keeps an alias
for backwards compatibility), as well as adds support for more tab color
  options (Theme, Indexed and Tint).

Signed-off-by: Thomas Charbonnel <github@charbonnel.email>

# Add support for more TabColor options

## Description

The following changes were made:
- Rename `TabColor` into `TabColorRGB`
- Add `TabColorTheme` as well as an enum for all its values
- Add `TabColorIndexed` as well as a default const for when it is unset
- Add `TabColorTint` as well as a default const for when it is unset

## Related Issue

#1283 

## Motivation and Context

The current version of Excelize only supports TabColor of type "RGB". Excel can set tab colors based on other dimensions (see [here](https://docs.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.tabcolor?view=openxml-2.8.1)).

## How Has This Been Tested

By adding tests to the existing tests for worksheet options in `sheetpr_test.go`
All the tests passed on Linux (Fedora 34, go1.16.13)

## Types of changes

## Checklist

### DIFF
--- a/sheetpr.go
+++ b/sheetpr.go
@@ -33,12 +33,35 @@ type (
 	Published bool
 	// FitToPage is a SheetPrOption
 	FitToPage bool
-	// TabColor is a SheetPrOption
-	TabColor string
+	// TabColorIndexed is a TabColor option, within SheetPrOption
+	TabColorIndexed int
+	// TabColorRGB is a TabColor option, within SheetPrOption
+	TabColorRGB string
+	// TabColorTheme is a TabColor option, within SheetPrOption
+	TabColorTheme int
+	// TabColorTint is a TabColor option, within SheetPrOption
+	TabColorTint float64
 	// AutoPageBreaks is a SheetPrOption
 	AutoPageBreaks bool
 	// OutlineSummaryBelow is an outlinePr, within SheetPr option
 	OutlineSummaryBelow bool
+)
+
+const (
+	TabColorThemeLight1 int = iota // Inverted compared to the spec because that's how Excel maps them
+	TabColorThemeDark1
+	TabColorThemeLight2
+	TabColorThemeDark2
+	TabColorThemeAccent1
+	TabColorThemeAccent2
+	TabColorThemeAccent3
+	TabColorThemeAccent4
+	TabColorThemeAccent5
+	TabColorThemeAccent6
+	TabColorThemeHyperlink
+	TabColorThemeFollowedHyperlink
+
+	TabColorUnsetValue int = -1
 )
 
 // setSheetPrOption implements the SheetPrOption interface.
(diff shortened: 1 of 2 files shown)


sheetpr.go
```diff
@@ -33,14 +33,42 @@ type (
...
+	// Equivalent to TabColorRGB, alias added for backwards compatibility
+	TabColor = TabColorRGB
 	// AutoPageBreaks is a SheetPrOption
 	AutoPageBreaks bool
 	// OutlineSummaryBelow is an outlinePr, within SheetPr option
 	OutlineSummaryBelow bool
 )
 
+const (
+	TabColorThemeUnset  int = iota - 1
```
xuri, line 53: Thanks for your PR. Sorry, after reconsidering and referencing [the specification of tab color](https://github.com/sc34wg4/OOXMLSchemas/blob/759b6d4cbd0a353eb414a69a314c60ff8d389733/Part1/OfficeOpenXML-XMLSchema-Strict/sml.xsd#L3473-L3479). I think we need to return 0 if the theme index or tint is unset, so return the enumeration `TabColorThemeLight1` when theme index is unset, and return `0` when tint is unset, the `TabColorThemeUnset`, `TabColorIndexedUnset` and `TabColorTintUnset` is no longer required.

sheetpr.go
```diff
@@ -33,14 +33,42 @@ type (
...
+	// TabColorIndexed is a TabColor option, within SheetPrOption
+	TabColorIndexed int
+	// TabColorRGB is a TabColor option, within SheetPrOption
+	TabColorRGB string
+	// TabColorTheme is a TabColor option, within SheetPrOption
+	TabColorTheme int
+	// TabColorTint is a TabColor option, within SheetPrOption
+	TabColorTint float64
+	// Equivalent to TabColorRGB, alias added for backwards compatibility
+	TabColor = TabColorRGB
```
xuri, line 45: Remove the `TabColor`, using the `TabColorRGB` instead of it, I will add a notice of this breaking change in the release note of [the next released version](https://xuri.me/excelize/en/releases/v2.6.1.html).

xuri (changes requested):

sheetpr.go
thomascharbonnel: You mean, because they're unsigned ints?
This means that if someone gets the theme for a cell that has no style, they'll get "light 1" (which isn't true)?

sheetpr.go
thomascharbonnel: updated

thomascharbonnel:
Regarding the documentation, i'm not sure if it's updated automatically when i changed the example tests, is that enough?

sheetpr.go
thomascharbonnel: I think setting tint to 0.0 by default makes sense as that means there's no tint, but it's not true for theme or indexed

thomascharbonnel: For ref.:
- Indexed = 0 is black color (https://docs.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.indexedcolors?view=openxml-2.8.1)
- Theme = 0 is light1, whose value depends on the theme

Would it make more sense to return a nil pointer for situations where the value is optional?

xuri: You're right, just return `0` in float64 data type when the tint is unset, and return `-1` for theme or index unset. So, rename the enumeration `TabColorThemeUnset` to `TabColorUnsetValue` as return value when theme and index unset, the `TabColorIndexedUnset` and `TabColorTintUnset` is no longer required.

xuri (changes requested):

xuri:
> Regarding the documentation, i'm not sure if it's updated automatically when i changed the example tests, is that enough?

Yes, the example code will be verified by running them as tests.

sheetpr.go
thomascharbonnel: Updated PR

xuri (approved):

xuri:

xuri:
I have added a new exported `ColorMappingType` used for color transformation, using it instead of integer data type theme color enumeration, and using the `ColorMappingType` prefix instead of the enumeration with `TabColorTheme`, because these enumerations not only can be used for the tab theme color, which also can be used for the drawing markup language and themes, so I think we should rename them.

thomascharbonnel:
Okay looks good 👌 

Thomas Charbonnel

PR #1297, opened 2022-08-01
#1296 add GetRowOpts to Rows

### LINKED ISSUE #1296: StreamReader (Rows) should provide access to Row style
**Description**

I'm thinking of changing `Rows{}` to provide the StyleID for the current row being read. Is this idea okay?

thomascharbonnel:
Signed-off-by: Thomas Charbonnel <github@charbonnel.email>

# PR Details

This PR fixes #1296 and adds a method called `GetStyleID` on struct `Rows{}` in order to get access to a row's style ID while stream-reading rows inside an XLSX file.

## Description

- Add `styleID` property to `Rows{}`
- Set row's style in `rows.Next()`
- Provide getter `rows.GetStyleID`
- Create relevant test

## Related Issue

#1296 

## Motivation and Context

This PR adds the ability to read a row's style, which wasn't present before.

## How Has This Been Tested

Test fixture file `Book1.xlsx` was updated to add a row style on the first row of Sheet2. The test makes sure `GetStyleID` returns the right ID.

## Types of changes

## Checklist


thomascharbonnel:
On second thought, it might be better to return a `RowOpt` struct instead

rows.go
```diff
@@ -101,6 +102,9 @@ func (rows *Rows) Next() bool {
 					rows.curRow = rowNum
 				}
 				rows.token = token
+				if styleID, _ := attrValToInt("s", xmlElement.Attr); styleID != 0 {
```
xuri, line 105: I agree with the return style ID of the row as is, without checking if which is less than `style.CellXfs.Count`, but I suggest that make `styleID` in uint data type and add the simple guard like this:


define a new exported constant in the `xmlDrawing.go`:

```go
const MaxCellStyles = 64000
```

```go
if styleID, _ := attrValToInt("s", xmlElement.Attr); 0 <= styleID && styleID < MaxCellStyles {
	rows.styleID = uint(styleID)
}
```

xuri (changes requested):

rows.go
```diff
@@ -111,6 +115,14 @@ func (rows *Rows) Next() bool {
 	}
 }
 
+// GetStyleID provise a function to get the current Rows' style ID. It will
```
xuri, line 118: I suggest changing the description of this exported function like this, and no example is needed.

```text
GetStyleID will return the style ID of the current row. The default style ID is 0.
```

xuri (changes requested):

rows.go
thomascharbonnel: Actually, i was thinking, does it make more sense to return a `RowOpts` instead of just the StyleID?

thomascharbonnel:
I've updated the PR to return RowOpts instead.
I haven't changed styleid to `uint` because that would mean changing `excelize.NewStyle` too

xuri (changes requested):
Thanks for your update. I've left some comments.

thomascharbonnel:
THanks for the comments! I'm still working on this, i've found a bug in `Next()` or `Columns()` (basically if you add `rows.Columns()` after the `rows.Next()` in my new test, the test fails, i think there is bug in with the way `curRow` is incremented.

rows_test.go
```diff
@@ -96,6 +96,31 @@ func TestRowsIterator(t *testing.T) {
...
+	expectedRowStyleID1 := RowOpts{Height: 17.0, Hidden: false, StyleID: 1}
+	expectedRowStyleID2 := RowOpts{Height: 17.0, Hidden: false, StyleID: 0}
+	expectedRowStyleID3 := RowOpts{Height: 17.0, Hidden: false, StyleID: 2}
+	f, err := OpenFile(filepath.Join("test", "Book1.xlsx"))
+	require.NoError(t, err)
+
+	rows, err := f.Rows(sheetName)
+	require.NoError(t, err)
+
+	rows.Next()
```
xuri, line 110: I think we should read the XML stream only once for better performance, and we can stash the row's attribute in `seekRowOpts` so we can reuse it. I have written some demo codes for reference, the variable `ro` should be renamed as I have suggested in the previous code review, and in the next step, the duplicate code of extract r element attributes can be simplified (the cyclomatic complexity of code should be < 15).

```go
// Rows defines an iterator to a sheet.
type Rows struct {
	err                     error
	curRow, seekRow         int
	needClose, rawCellValue bool
	sheet                   string
	f                       *File
	tempFile                *os.File
	sst                     *xlsxSST
	decoder                 *xml.Decoder
	token                   xml.Token
	curRowOpts, seekRowOpts RowOpts
}

// Next will return true if find the next row element.
func (rows *Rows) Next() bool {
	rows.seekRow++
	if rows.curRow >= rows.seekRow {
		rows.curRowOpts = rows.seekRowOpts
		return true
	}
	for {
		token, _ := rows.decoder.Token()
		if token == nil {
			return false
		}
		switch xmlElement := token.(type) {
		case xml.StartElement:
			if xmlElement.Name.Local == "row" {
				rows.curRow++
				if rowNum, _ := attrValToInt("r", xmlElement.Attr); rowNum != 0 {
					rows.curRow = rowNum
				}
				rows.token = token
				ro := RowOpts{}
				if styleID, err := attrValToInt("s", xmlElement.Attr); err == nil && styleID > 0 && styleID < MaxCellStyles {
					ro.StyleID = styleID
				}
				if hidden, err := attrValToBool("hidden", xmlElement.Attr); err == nil {
					ro.Hidden = hidden
				}
				if height, err := attrValToFloat("ht", xmlElement.Attr); err == nil {
					ro.Height = height
				}
				rows.curRowOpts = ro
				return true
			}
		case xml.EndElement:
			if xmlElement.Name.Local == "sheetData" {
				return false
			}
		}
	}
}
```

```go
// Columns return the current row's column values. This fetches the worksheet
// data as a stream, returns each cell in a row as is, and will not skip empty
// rows in the tail of the worksheet.
func (rows *Rows) Columns(opts ...Options) ([]string, error) {
	if rows.curRow > rows.seekRow {
		return nil, nil
	}
	var rowIterator rowXMLIterator
	var token xml.Token
	rows.rawCellValue, rows.sst = parseOptions(opts...).RawCellValue, rows.f.sharedStringsReader()
	for {
		if rows.token != nil {
			token = rows.token
		} else if token, _ = rows.decoder.Token(); token == nil {
			break
		}
		switch xmlElement := token.(type) {
		case xml.StartElement:
			rowIterator.inElement = xmlElement.Name.Local
			if rowIterator.inElement == "row" {
				rowNum := 0
				if rowNum, rowIterator.err = attrValToInt("r", xmlElement.Attr); rowNum != 0 {
					rows.curRow = rowNum
				} else if rows.token == nil {
					rows.curRow++
				}
				ro := RowOpts{}
				if styleID, err := attrValToInt("s", xmlElement.Attr); err == nil && styleID > 0 && styleID < MaxCellStyles {
					ro.StyleID = styleID
				}
				if hidden, err := attrValToBool("hidden", xmlElement.Attr); err == nil {
					ro.Hidden = hidden
				}
				if height, err := attrValToFloat("ht", xmlElement.Attr); err == nil {
					ro.Height = height
				}
				rows.seekRowOpts = ro
				if rows.curRow > rows.seekRow {
					rows.token = nil
					return rowIterator.columns, rowIterator.err
				}
			}
			if rows.rowXMLHandler(&rowIterator, &xmlElement, rows.rawCellValue); rowIterator.err != nil {
				rows.token = nil
				return rowIterator.columns, rowIterator.err
			}
			rows.token = nil
		case xml.EndElement:
			if xmlElement.Name.Local == "sheetData" {
				return rowIterator.columns, rowIterator.err
			}
		}
	}
	return rowIterator.columns, rowIterator.err
}
```

xuri (changes requested):

rows_test.go
thomascharbonnel: Thanks for your help, the tests are passing now!

sheet.go
```diff
@@ -928,6 +928,34 @@ func attrValToInt(name string, attrs []xml.Attr) (val int, err error) {
 	return
 }
 
+// attrValToFloat provides a function to convert the local names to a float64
```
xuri, line 931: I suggest adding the unit test in the `excelize_test.go` like this to improve code coverage in the report.

```go
func TestAttrValToBool(t *testing.T) {
	_, err := attrValToBool("hidden", []xml.Attr{
		{Name: xml.Name{Local: "hidden"}},
	})
	assert.EqualError(t, err, `strconv.ParseBool: parsing "": invalid syntax`)
}

func TestAttrValToFloat(t *testing.T) {
	_, err := attrValToFloat("ht", []xml.Attr{
		{Name: xml.Name{Local: "ht"}},
	})
	assert.EqualError(t, err, `strconv.ParseFloat: parsing "": invalid syntax`)
}
```

xuri (changes requested):
Thanks for your update. In addition, I suggest adding some unit tests.

sheet.go
thomascharbonnel: Hi, done thank you! I've added a test for the positive situation (when the XML attribute is read successfully).

xuri (approved):

xuri:
LGTM, thanks for your great work.

Issue #1299, opened 2022-08-03
StreamWriter.SetRow should not insert a cell when value is nil


thomascharbonnel:
**Description**

When using StreamWriter's `SetRow`, one has to pass a `[]interface{}` of values. However, when a value given is nil (eg. `values := []any{nil, nil, excelize.Cell{}}`, Excelize still creates an XML `c` cell. This creates problems when a row-level style has been set.

**Steps to reproduce the issue:**
1. Create a new document using StreamWriter
2. Set a row-level style using `RowOpts` (this style should apply to all the cells in that row unless overwritten by a specific cell
3. Insert a `Cell{}` preceded by `nil` cells using `SetRow`. (eg. A1 and B1 are nil but C1 contains "My string")

**Describe the results you received:**

If N `nil` cells are inserted before a `Cell{}`, Excel will show those cells using the default style (white background) rather than the row-level style.
Ie. the XML worksheet will contain `<c t="str" r="A1"></c><c t="str" r="B1"></c><c t="str" r="C1">My string</c>`.

**Describe the results you expected:**

Empty cells should not be written to the XML worksheet, so A1 and B1 should be empty cells using the row-level style.
XML should be `<c t="str" r="C1">My string</c>`.
This is Excel's behavior

**Output of `go version`:**

```text
go version 1.18.3 linux/amd64
```

**Excelize version or commit ID:**

```text
master
```

**Environment details (OS, Microsoft Excel™ version, physical, etc.):**

Running on Red Hat Linux.

**Library code to change**

https://github.com/qax-os/excelize/blob/master/stream.go#L329 This is the issue, the loop should test if `val == nil` and if yes it should `continue`.

thomascharbonnel:
It's worth discussing whether an empty `Cell{}` should also be written or not, on top of `nil`

PR #1301, opened 2022-08-03
#1299 don't write nil values in SetRow

thomascharbonnel:
# PR Details

In `SetRow`, ignore `nil` values when writing to XML file.

## Description

- In `SetRow`, skip `nil` values when writing
- Add a test to enforce this behavior

## Related Issue

#1299

## Motivation and Context

Before this change, when setting `nil` values using `SetRow`, Excelize would create empty `<c>` elements. These elements are seen by Excel as separate cells with a style separate from the row-level style. Which breaks row-level styling.

## How Has This Been Tested

A test was added to check the parsed XML output.

## Types of changes

## Checklist


xuri (approved):

PR #1343, opened 2022-09-10
Fix DeleteComment slice bounds out of range

dafengge0913:
## Description
Modify slice length while range will cause "slice bounds out of range" error.

## How Has This Been Tested
Test by TestDeleteComment function in comment_test.go


xuri (approved):
LGTM. Thanks for your contribution. Since that Excel doesn't allow inserting multiple comments in one cell, I think we also need to add checking for this when adding comments.

PR #1393, opened 2022-11-15
repair:The xml alias does not take effect

renxiaotu:
> The following code paging element is aliased as`<rowBreaks>`, but it is coded as`<xlsxBreaks>`

# Recurrence method demo

```go
package main

import (
	"fmt"
	"github.com/xuri/excelize/v2"
	"math/rand"
)

func main() {
	file := excelize.NewFile()
	defer func() {
		if err := file.Close(); err != nil {
			fmt.Println(err)
		}
	}()

	streamWriter, err := file.NewStreamWriter("Sheet1")
	if err != nil {
		fmt.Println(err)
	}

	cell, err := excelize.CoordinatesToCellName(1, 20)
	if err != nil {
		fmt.Println(err)
	}

	err = streamWriter.File.InsertPageBreak("Sheet1", cell)
	if err != nil {
		fmt.Println(err)
	}

	for rowID := 1; rowID <= 50; rowID++ {
		row := make([]interface{}, 5)
		for colID := 0; colID < 5; colID++ {
			row[colID] = rand.Intn(640000)
		}
		cell, _ := excelize.CoordinatesToCellName(1, rowID)
		if err := streamWriter.SetRow(cell, row); err != nil {
			fmt.Println(err)
		}
		if rowID%20 == 0 {
			err = streamWriter.File.InsertPageBreak("Sheet1", cell)
			if err != nil {
				fmt.Println(err)
			}
		}
	}

	if err := streamWriter.Flush(); err != nil {
		fmt.Println(err)
	}

	if err := file.SaveAs("Book1.xlsx"); err != nil {
		fmt.Println(err)
	}
}
```

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -655,10 +655,20 @@ func (sw *StreamWriter) Flush() error {
 // names order range.
 func bulkAppendFields(w io.Writer, ws *xlsxWorksheet, from, to int) {
 	s := reflect.ValueOf(ws).Elem()
+	t := reflect.TypeOf(ws).Elem()
 	enc := xml.NewEncoder(w)
 	for i := 0; i < s.NumField(); i++ {
 		if from <= i && i <= to {
-			_ = enc.Encode(s.Field(i).Interface())
+			local := t.Field(i).Tag.Get("xml")
+			if local == "" {
+				_ = enc.Encode(s.Field(i).Interface())
+			} else {
+				_ = enc.EncodeElement(s.Field(i).Interface(), xml.StartElement{
+					Name: xml.Name{
+						Local: local,
+					},
+				})
+			}
 		}
 	}
 }


xuri:
Thanks for your PR. This library doesn't support insert page break in stream mode currently, as the documentation says: the normal mode
 functions and stream mode functions can't be work mixed, so don't use the `InsertPageBreak` function in the stream mode. We need to implement this feature in another way.

renxiaotu:
1. I just take page breaks as an example. In fact, the problem is that the `bulkAppendFields` method is not compatible with xml aliases
2. Except for the alias error, the paging data was written normally in the `StreamWriter` mode in my test. Is it because the paging data itself does not work in the `StreamWriter` mode that the paging function of 'StreamWriter' is not open?

xuri:
Please use the `InsertPageBreak` function in normal mode, the normal mode functions and stream mode functions can't be work mixed. The internal `bulkAppendFields` function is just used to serialize partial elements of the sheet data. We need an implemet insert page break function in stream mode as other stream functions such as `MergeCells` and `SetPanes`.

renxiaotu:
If you have a better plan, this pull does not need to be merged, but I still need to implement paging in a mixed way. Although paging data will occupy memory, this is the only way to make my requirements work well at present

xuri:
I'll consider add support for this feature recently.

renxiaotu:
That was great！

xuri:
I have implemented this feature based on your branch, and I have made the pointer of the file un-exported to avoid using normal mode functions in stream mode by mistakes. Please upgrade to the master branch code, and this feature will be released in the next version.

xuri (commented):
I have implemented this feature based on your branch, and I have made the pointer of the file un-exported to avoid using normal mode functions in stream mode by mistakes. Please upgrade to the master branch code, and this feature will be released in the next version. Now you can insert the page break with stream writer like this:

```diff
- err = streamWriter.File.InsertPageBreak("Sheet1", cell)
+ err = streamWriter.InsertPageBreak(cell)
```

renxiaotu:
Great. Your work efficiency is really enjoyable

PR #1406, opened 2022-11-28
This closes #1405, add new function SetSheetBackgroundFromBytes used …

### LINKED ISSUE #1405: Add new function SetSheetBackgroundFromBytes used for setting background image by given image data
Because of business need, I want to add a new method(`SetSheetBackgroundFromBytes`) for struct File. 

the function of this method is similar with existing method(`SetSheetBackground`), they both can add background image for excel files, but the difference between `SetSheetBackground` and `SetSheetBackgroundFromBytes` is, the new method use binary image data as image source. And SetSheetBackgroundFromBytes do not need to specify the extension of image data, It can determine the image format by itself.

由于业务需要，我想为File结构体增加一个新的方法：SetSheetBackgroundFromBytes。

该方法的功能与现存SetSheetBackground方法相同，即添加背景图像，不同的是新增方法支持使用二进制数据设置背景图像，同时，SetSheetBackgroundFromBytes无需指定文件后缀，该方法可自行判断二进制数据中所包含图像的格式。

jianxinhou:
# PR Details

## Description

Because of business need, We added a new method for struct File：`SetSheetBackgroundFromBytes`. 

the function of this method is similar with existing method(SetSheetBackground), they both can add background image for excel files, but the difference between SetSheetBackground and SetSheetBackgroundFromBytes is, the new method use binary image data as image source. And SetSheetBackgroundFromBytes do not need to specify the extension of image data, It can determine the image format by itself.

由于业务需要, 我们为File结构体增加了一个新的方法：`SetSheetBackgroundFromBytes`。

该方法的功能与现存SetSheetBackground方法相同，即添加背景图像，不同的是新增方法支持使用二进制数据设置背景图像，同时，SetSheetBackgroundFromBytes无需指定文件后缀，该方法可自行判断二进制数据中所包含图像的格式。

## Related Issue

None

无

## Motivation and Context

In our business, we want to set the binary image data requested from the upstream service as the background of the excel file directly, so we implement `SetSheetBackgroundFromBytes`.

在业务场景中，我们希望能够将从上游服务中请求到的二进制图片数据直接设置为excel文件的背景，因此实现了`SetSheetBackgroundFromBytes`。

## How Has This Been Tested

- We loaded all the images under `test/images/` and `excelize.svg`, and called `SetSheetBackgroundFromBytes` to set the binary data loaded into memory as the background image, the result shows that this method can distinguish the binary data format and set for background, see `TestSetBackgroundFromBytes` in `sheet_test.go`;
- In addition to the above images, we also tested with other positive and negative example files that are not in this project, and the results show that `SetSheetBackgroundFromBytes` can be executed correctly;
- We found that when we use the image `test/images/excel.wmz` as the background, no matter calling the existing `SetSheetBackground` or the added `SetSheetBackgroundFromBytes`, the set background cannot be displayed correctly, maybe there is a problem with the image itself; 
- We found that when we use the `.svg` file as the background image, it cannot be displayed normally in excel, and you cannot manually set the `.svg` file as the background image in excel. It seems that excel does not support setting the `.svg` file as the background image, but it can be displayed normally in WPS.

- 我们读取了`test/images/`下的所有图像以及`excelize.svg`，并调用`SetSheetBackgroundFromBytes`将加载到内存的二进制数据设置为背景图像，结果显示该方法能正常区分二进制数据格式并设置背景，详见`sheet_test.go`中的`TestSetBackgroundFromBytes`；
- 除上述图像外，我们还使用其他不在本项目中的正例和反例文件做了测试，结果显示`SetSheetBackgroundFromBytes`均能够正确执行；
- 我们发现使用图像`test/images/excel.wmz`作为背景时，无论调用现存的`SetSheetBackground`还是新增的`SetSheetBackgroundFromBytes`都无法正确显示设置的背景，可能是图像本身有问题；
- 我们发现使用`.svg`文件作为背景图像时，在excel中无法正常显示，同时在excel中也无法手动设置`.svg`文件作为背景图像，似乎excel并不支持设置.svg文件作为背景，但在WPS中可以正常显示。

## Types of changes

## Checklist


picture.go
```diff
@@ -13,10 +13,12 @@ package excelize
 
 import (
 	"bytes"
+	"compress/gzip"
 	"encoding/json"
 	"encoding/xml"
 	"image"
 	"io"
+	"io/ioutil"
```
xuri, line 21: I suggest letting the user do the image format detection, and setting sheet background by specifying the format. In addition, the package `ioutil` has been deprecated in Go 1.16. We can use `io` or `os` instead of that.

sheet_test.go
```diff
@@ -463,3 +465,21 @@ func TestAttrValToFloat(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, 42.1, got)
 }
+
+func TestSetSheetBackgroundFromBytes(t *testing.T) {
+	files := []string{"../../excelize.svg", "excel.emf", "excel.emz", "excel.gif", "excel.jpg", "excel.png", "excel.tif", "excel.wmf", "excel.wmz"}
+	f, err := OpenFile("./test/EmptyWorkbook.xlsx")
```
xuri, line 471: The workbook `test/EmptyWorkbook.xlsx` seem to be modified after the test, we need to keep it as the origin.

sheet.go
```diff
@@ -497,11 +497,27 @@ func (f *File) SetSheetBackground(sheet, picture string) error {
 		return ErrImgExt
 	}
 	file, _ := os.ReadFile(filepath.Clean(picture))
-	name := f.addMedia(file, ext)
+	return f.setSheetBackground(sheet, ext, file)
+}
+
+// SetSheetBackgroundFromBytes provides a function to set background picture
+// by given worksheet name and image data
+func (f *File) SetSheetBackgroundFromBytes(sheet string, picture []byte) error {
```
xuri, line 505: Users can be specifying the image format by file `extension` name parameter:

```go
func (f *File) SetSheetBackgroundFromBytes(sheet, extension string, picture []byte) error
```

xuri (changes requested):

sheet_test.go
jianxinhou: The test will set the background image for `f`, but instead of calling `Save` it calls `SaveAs` to save another file, 
rather than itself. I checked before calling test and after calling test, the size of `./test/EmptyWorkbook.xlsx` is 8476 bytes, and the content is also consistent.

jianxinhou: The test will set the background image for f, but instead of calling Save it calls SaveAs to save another file,
rather than itself. I checked before calling test and after calling test, the size of ./test/EmptyWorkbook.xlsx is 8454 bytes, and the content is also consistent.

picture.go
jianxinhou: solved

sheet.go
jianxinhou: Emmm, do you mean that I should use parameters to determine the data format instead of using the binary data?

xuri: Yep, detection of image format by header bytes maybe not be accurate, the image payload binary data can be tampered with, this library cannot guarantee the image is valid even if it has correct header bytes, this is out of the scope of the library, we should leave this job to the user.

sheet_test.go
xuri: All right. Could we use the `NewFile` function to prepare a blank workbook instead of adding a new file in the test folder?

jianxinhou: Resolved

sheet.go
jianxinhou: Resolved

sheet_test.go
```diff
@@ -463,3 +465,30 @@ func TestAttrValToFloat(t *testing.T) {
...
+		"test/images/excel.emz": ".emz",
+		"test/images/excel.gif": ".gif",
+		"test/images/excel.jpg": ".jpg",
+		"test/images/excel.png": ".png",
+		"test/images/excel.tif": ".tif",
+		"test/images/excel.wmf": ".wmf",
+		"test/images/excel.wmz": ".wmz",
+	}
+	f := NewFile()
+	index := f.NewSheet("sheet1")
```
xuri, line 482: The workbook has a default worksheet named "Sheet1" by default when it has been created, so we needn't create that.

In other ways, could we test it by set worksheet background images for multi worksheets instead of creating many workbooks?

sheet_test.go
```diff
@@ -463,3 +465,30 @@ func TestAttrValToFloat(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, 42.1, got)
 }
+
+func TestSetSheetBackgroundFromBytes(t *testing.T) {
+	files := map[string]string{
+		"excelize.svg":          ".svg",
+		"test/images/excel.emf": ".emf",
```
xuri, line 472: We need to consider the Windows directory separator in the path, I suggest using the `filepath.Join` to connect the directory name and file name.

sheet.go
```diff
@@ -497,11 +497,30 @@ func (f *File) SetSheetBackground(sheet, picture string) error {
...
 	}
 	file, _ := os.ReadFile(filepath.Clean(picture))
-	name := f.addMedia(file, ext)
+	return f.setSheetBackground(sheet, ext, file)
+}
+
+// SetSheetBackgroundFromBytes provides a function to set background picture
+// by given worksheet name, extension name and image data
+func (f *File) SetSheetBackgroundFromBytes(sheet, extension string, picture []byte) error {
+	if len(picture) == 0 {
```
xuri, line 506: The error message was not accurate, using the `ErrParameterInvalid` instead of it will be better, and we need add a test for this case.

sheet.go
```diff
@@ -497,11 +497,30 @@ func (f *File) SetSheetBackground(sheet, picture string) error {
...
+	}
+	ext, ok := supportedImageTypes[extension]
+	if !ok {
+		return ErrImgExt
+	}
+	return f.setSheetBackground(sheet, ext, picture)
+}
+
+// setSheetBackground provides a function to set background picture by given
+// worksheet name,  file name extension and image data.
```
xuri, line 517: Trim repeat space characters in the documentation.

sheet.go
```diff
@@ -497,11 +497,30 @@ func (f *File) SetSheetBackground(sheet, picture string) error {
...
+	return f.setSheetBackground(sheet, ext, file)
+}
+
+// SetSheetBackgroundFromBytes provides a function to set background picture
+// by given worksheet name, extension name and image data
+func (f *File) SetSheetBackgroundFromBytes(sheet, extension string, picture []byte) error {
+	if len(picture) == 0 {
+		return ErrImgExt
+	}
+	ext, ok := supportedImageTypes[extension]
```
xuri, line 509: I think the image type mapping statements can be simplified by the move to the `setSheetBackground` function. For example:

```go
// SetSheetBackground provides a function to set background picture by given
// worksheet name and file path.
func (f *File) SetSheetBackground(sheet, picture string) error {
	var err error
	// Check picture exists first.
	if _, err = os.Stat(picture); os.IsNotExist(err) {
		return err
	}
	file, _ := os.ReadFile(filepath.Clean(picture))
	return f.setSheetBackground(sheet, path.Ext(picture), file)
}

// SetSheetBackgroundFromBytes provides a function to set background picture
// by given worksheet name, extension name and image data.
func (f *File) SetSheetBackgroundFromBytes(sheet, extension string, picture []byte) error {
	if len(picture) == 0 {
		return ErrParameterInvalid
	}
	return f.setSheetBackground(sheet, extension, picture)
}

// setSheetBackground provides a function to set background picture by given
// worksheet name, file name extension and image data.
func (f *File) setSheetBackground(sheet, extension string, file []byte) error {
	imageType, ok := supportedImageTypes[extension]
	if !ok {
		return ErrImgExt
	}
	name := f.addMedia(file, imageType)
	sheetXMLPath, _ := f.getSheetXMLPath(sheet)
	sheetRels := "xl/worksheets/_rels/" + strings.TrimPrefix(sheetXMLPath, "xl/worksheets/") + ".rels"
	rID := f.addRels(sheetRels, SourceRelationshipImage, strings.Replace(name, "xl", "..", 1), "")
	if err := f.addSheetPicture(sheet, rID); err != nil {
		return err
	}
	f.addSheetNameSpace(sheet, SourceRelationship)
	return f.setContentTypePartImageExtensions()
}
```

xuri (changes requested):

xuri (approved):
I have fixed some code review issues and updates the unit test case and documentation for this based on your branch. Please upgrade to the master branch code, and this feature will be released in the next version.

PR #1415, opened 2022-12-06
Fix binary string regex

nesstord:
# PR Details

Change regex value for "bstrUnmarshal" function in "lib.go" file.

## Description

The value of the regex "_x[a-zA-Z\d]{4}_" has been changed to "_x[a-fA-F\d]{4}_" because hexadecimal values only contain the characters A-F.

## Related Issue
https://github.com/qax-os/excelize/issues/1414

## Types of changes


xuri (approved):
Thanks for your contribution. I have update test case and made some changes for this based on your branch, and this feature will be released in the next version.

PR #1440, opened 2023-01-07
supports directly read file byte data 

Kingson4Wu:
supports directly read file byte data and returns a populated spreadsheet file struct for it

## Motivation and Context

supports directly read file byte data,  make it convenient to use file by `go:embed`. for example:
<pre>
├── module
│   └── excel
│        ├── Book1.xlsx
│        └── excel.go
</pre>

excel.go read Book1.xlsx  can write like 

```go
//go:embed Book1.xlsx
var bookExcel []byte
f, err := excelize.OpenFileByte(bookExcel)
```

instead of

```go
f, err := excelize.OpenFile("module/excel/Book1.xlsx")
```

### DIFF
--- a/excelize.go
+++ b/excelize.go
@@ -113,6 +113,17 @@ func OpenFile(filename string, opts ...Options) (*File, error) {
 	return f, file.Close()
 }
 
+// OpenFileByte take the file byte data of an spreadsheet file and returns a populated
+// spreadsheet file struct for it.
+//
+func OpenFileByte(b []byte) (*File, error) {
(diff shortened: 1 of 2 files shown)


xuri (commented):
Thanks for your pull request, since the already existing function `OpenReader` support reading data stream from `io.Reader`, for example:

```go
file, err := os.Open("Book1.xlsx")
if err != nil {
    fmt.Println(err)
    return
}
f, err := excelize.OpenReader(file)
```

So I think this change was not necessary, and I will close this, if you have any questions, please let me know.

PR #1456, opened 2023-01-24
Issue 1455 pre generate strings for months

nathj07:
On reviewing the code I was focussed on I figured I could add a couple of slices and then simplify some the code around getting month name abbreviations. At this point I thought I'd get some input, in case you want to go a different way? Or, if you like this approach if you want to implement for all languages I could go ahead and do that.

I have also removed some of the unnecessary `string([]rune(...` casts that were in place. Doing this prevents these tings from escaping to the heap

Either way, it seemed worth getting some feedback before going any further.

## Description

As an example of the proposed change I've introduced a couple of month abbreviation slices and shown how these could be used to simplify their corresponding functions. The existing tests already cover this code and continue to pass.

## Related Issue

[issue-1455](https://github.com/qax-os/excelize/issues/1455)

## Motivation and Context

This is an attempt to reduce the memory needed by these language functions that have string concatenation involved.

## How Has This Been Tested

The existing tests cover this code and they continue to pass. I do not think extra tests are needed here. Since adding these changes I can see, via `go build -gcflags="-m -l"`, that there is less that escapes to the heap.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x ] My code follows the code style of this project.
- [ x] I have read the **CONTRIBUTING** document.
- [ x] All new and existing tests passed.

### DIFF
--- a/numfmt.go
+++ b/numfmt.go
@@ -244,8 +244,12 @@ var (
 	monthNamesWelsh = []string{"Ionawr", "Chwefror", "Mawrth", "Ebrill", "Mai", "Mehefin", "Gorffennaf", "Awst", "Medi", "Hydref", "Tachwedd", "Rhagfyr"}
 	// monthNamesWolof list the month names in the Wolof.
 	monthNamesWolof = []string{"Samwiye", "Fewriye", "Maars", "Awril", "Me", "Suwe", "Sullet", "Ut", "Septàmbar", "Oktoobar", "Noowàmbar", "Desàmbar"}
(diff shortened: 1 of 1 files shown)


nathj07:
@xuri I spotted some commented out code I'd left behind. I tidied this up in the most recent push. Hopefully this is looking good now, let me know what you think

xuri (approved):
Thanks for your contribution. This is a good way to improve performance, I think we can implement it for all languages.

PR #1481, opened 2023-02-23
Fixes #756 remove empty string when set cell value to nil in Stream W…

### LINKED ISSUE #756: Purpose of setting empty string to cell in "nil" case
Hi!

Explain please, what the purpose of setting the value "" to cell in SetCellValue method, when value argument is nil? I don't report is as a bug, because as I can see it's done deliberately, but I don't get a point of that. 
Here is the line of code I'm talking about https://github.com/360EntSecGroup-Skylar/excelize/blob/master/cell.go#L96

The problem is that It fills the cell with some value, and excel doesn't take that kind cells as empty (you can check it by calling ISBLANK() excel function on them), so when you for example set row data by using SetSheetRow and want skip some cells in the row by passing a nil, it fills them with "", and so you not skipping them in fact, only setting to empty string.

So what the point of that line of code? Can you just skip the case of value == nil in SetCellValue?

barismar:
# PR Details

Not setting empty string when set cell value to nil in Stream Writer
## Description

Not setting cell value to `c.setCellValue("")` (empty string) when  `setCellValFunc(c *xlsxC, val interface{})` val param accept `nil` value

## Related Issue

#756

## Motivation and Context

Excel that generated with Stream Writer can use excel function like `ISBLANK()`, `COUNTA()` etc.

## Types of changes

## Checklist

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -536,7 +536,7 @@ func (sw *StreamWriter) setCellValFunc(c *xlsxC, val interface{}) error {
 	case bool:
 		c.T, c.V = setCellBool(val)
 	case nil:
-		c.setCellValue("")
+		return err
 	case []RichTextRun:
 		c.T, c.IS = "inlineStr", &xlsxSI{}
 		c.IS.R, err = setRichText(val)


xuri (commented):
LGTM, thanks for your contribution. This patch will be released in the next version.

PR #1490, opened 2023-03-13
Fix NA error

rpoetrap:
# PR Details

Fix NA error

## Description

Fix NA error

## Types of changes

## Checklist


xuri (changes requested):
Thanks for your pull request. The second parameter of the `newErrorFormulaArg` function represents the error message, which is used to explain the error string. I purpose making the result of the `CalcCellValue` function an error string, and the error message of that as the error value. What do you think about that?

rpoetrap:
I had this problem when trying to use this formula `=IFNA(VLOOKUP(...), ...)`, i thought this problem caused by the Error value in formulaArg is not `#N/A`?
or should I change the condition in `IFNA` function from `arg.Value() == formulaErrorNA` to `arg.String == formulaErrorNA`?

xuri (commented):
I means that using formula error string in the result of the `CalcCellValue` function when error occurs, and using the error message in the `CalcCellValue` function returns error. I have a changes for all formula functions and updates the unit tests based on your branch. For example:

```go
result, err := f.CalcCellValue("Sheet1", "A1")
// Now, the variable 'result' was not empty when the error occurs
// result: "#N/A"
// err: "HLOOKUP no result found"
```

rpoetrap:
Oh that's perfect, btw I just added a fix to the IFNA function.

xuri (approved):

xuri (approved):
LGTM. Thanks for your contribution.

PR #1517, opened 2023-04-10
Fix corrupted utf-8 strings after trimming cell values

IAkumaI:
Usage value[:TotalCellChars] is incorrect. If value contains utf-8 string it may corrupt the last symbol in value.

# PR Details

Bug fix of slicing strings in Go

## Description

Excel has upper limit of character count which is 32767, but excelize checks value length by bytes instead by characters.
If I put utf-8 value (repeated 17000 times two-bytes symbol "Ы", for example) to a cell, excelize won't show any error, but file will be corrupted. Excel will show error on this file. Mac Numbers doesn't show any error, but empty cell and half-empty file.

## Related Issue

[

](https://github.com/qax-os/excelize/issues/1518)

## Types of changes

## Checklist

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -397,8 +397,9 @@ func (f *File) SetCellStr(sheet, cell, value string) error {
 // setCellString provides a function to set string type to shared string
 // table.
 func (f *File) setCellString(value string) (t, v string, err error) {
-	if len(value) > TotalCellChars {
-		value = string([]rune(value)[:TotalCellChars])
+	value_runes := []rune(value)
+	if len(value_runes) > TotalCellChars {
+		value = string(value_runes[:TotalCellChars])
 	}
 	t = "s"
 	var si int
@@ -458,8 +459,9 @@ func (f *File) setSharedString(val string) (int, error) {
 
 // trimCellValue provides a function to set string type to cell.
 func trimCellValue(value string) (v string, ns xml.Attr) {
-	if len(value) > TotalCellChars {
-		value = string([]rune(value)[:TotalCellChars])
+	value_runes := []rune(value)
+	if len(value_runes) > TotalCellChars {
+		value = string(value_runes[:TotalCellChars])
 	}
 	if len(value) > 0 {
 		prefix, suffix := value[0], value[len(value)-1]
--- a/cell.go
+++ b/cell.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 )
 
 // CellType is the type of cell value type.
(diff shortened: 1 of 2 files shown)


IAkumaI:
Maybe it is better to use `https://pkg.go.dev/unicode/utf8#RuneCountInString` instead of converting whole string evert method call.

cell.go
```diff
@@ -397,8 +397,9 @@ func (f *File) SetCellStr(sheet, cell, value string) error {
 // setCellString provides a function to set string type to shared string
 // table.
 func (f *File) setCellString(value string) (t, v string, err error) {
-	if len(value) > TotalCellChars {
-		value = value[:TotalCellChars]
+	value_runes := []rune(value)
```
xuri, line 400: Use camel case style name in Go variables. I agree with you, we can using the utf8 std library at here:

```go
if utf8.RuneCountInString(value) > TotalCellChars {
    value = string([]rune(value)[:TotalCellChars])
}
```

xuri (changes requested):

cell_test.go
```diff
@@ -176,6 +176,18 @@ func TestSetCellFloat(t *testing.T) {
 	assert.EqualError(t, f.SetCellFloat("Sheet:1", "A1", 123.42, -1, 64), ErrSheetNameInvalid.Error())
 }
 
+func TestSetLongUtf8CellValues(t *testing.T) {
```
xuri, line 179: - I suggest using the `TestSetCellValuesMultiByte` function instead of `TestSetLongUtf8CellValues`, so we can run test with all set cell value related cases which prefix by `TestSetCellValues*`
- Using the unicode represent multi byte character
- Using the `value` instead of `long_value`

For example:

```go
value := strings.Repeat("\u4E00", TotalCellChars+1)
```

cell.go
```diff
@@ -458,8 +459,9 @@ func (f *File) setSharedString(val string) (int, error) {
 
 // trimCellValue provides a function to set string type to cell.
 func trimCellValue(value string) (v string, ns xml.Attr) {
-	if len(value) > TotalCellChars {
-		value = value[:TotalCellChars]
+	value_runes := []rune(value)
```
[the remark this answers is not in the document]
IAkumaI: Done

cell.go
IAkumaI: Done

cell_test.go
IAkumaI: Done

cell.go
IAkumaI: Done

IAkumaI:
It's done, but "resolve conversation" button doesn't work. Don't know why :(

xuri (approved):

PR #1692, opened 2023-10-17
Update checkSheet GetTables error: index out of range

lujin1:
# PR Details

## Types of changes

## Checklist


lujin1:
When the sheet does not start with table, an error will be reported when getting tables.

xuri (changes requested):
Thanks for your PR. Could you add unit tests for this changes?

tumbleweedd:
> When the sheet does not start with table, an error will be reported when getting tables.

Could you please provide the test file that caused the panic? I will be very grateful

xuri:
Hi @tumbleweedd, I have already analytics and created a the special case workbook file can be reproduce this problem, but the fix in this changes will be cause the some cell value was missing, I think we need to made some change based on this.

xuri (approved):
LGTM. Thanks for your contribution. I have made some changes based on your branch. This feature will be released in the next version.

xuri:
According to [the benchmark report](https://github.com/xuri/excelize/wiki#performance-figures) of v2.8.1, the fix slows down the get rows about 3 times, and memory using about 17 times. We need to fixed the performance impact later.

Issue #1383, opened 2022-11-02
The file size increases abnormally when saving


seanliang:
The original file size is 8KB, and it contains nothing but only one empty row (row# 1048576), the file size will become 2.8MB when saving.
[origin.xlsx](https://github.com/qax-os/excelize/files/9915838/origin.xlsx)

Unzipped both files and you can find that the file size of sheet1.xml has increased from 1KB to 23MB.

```
package main

import (
	"log"
	"os"

	"github.com/xuri/excelize/v2"
)

func main() {
	fi, err := os.Stat("origin.xlsx")
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Origin file size: %v", fi.Size())
	file, _ := excelize.OpenFile("origin.xlsx")
	file.SetCellValue("Test", "A1", "test")
	file.SaveAs("copy.xlsx")
	fi, _ = os.Stat("copy.xlsx")
	log.Printf("Copy file size: %v", fi.Size())
}
```

The output is
```
Origin file size: 8382
Copy file size: 2759163
```

xuri:
Thanks for your issue, I have fixed it. Please upgrade to the master branch code, and this patch will be released in the next version.

seanliang:
Awesome! Thanks for your efforts!

PR #1713, opened 2023-11-06
Reduce memory consumption

### LINKED ISSUE #1712: Memory consumption by trimming the rows and cells
**Description**

By having a lot of rows and columns, memory consumption increases a lot. The reason is that we always allocate a new underlying array instead of re-using the existing one.

1. By trying to convert a huge amount of rows (30000+) into the xlsx

trimCell issue:
heap1

trimRow issue: (append)
heap-append

We can reusing the existing underlying array instead of always allocating a new one.

**Output of `go version`:**

```text
1.21
```

```text
v2.7.1
```

krstak:
# PR Details

Reduce memory consumption

## Description
By having a lot of rows and columns, memory consumption increases a lot. The reason is that we always allocate a new underlying array instead of re-using the existing one.

## Related Issue

https://github.com/qax-os/excelize/issues/1712

## Motivation and Context

By having a huge amount of rows, the library consumes a lot of memory for converting it to the xlsx file. After this change, a lot of memory would be saved.

## How Has This Been Tested

I tried to convert a huge amount of data (60000 rows) to the xlsx 

## Types of changes

## Checklist

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -177,9 +177,8 @@ func (f *File) workSheetWriter() {
 
 // trimRow provides a function to trim empty rows.
 func trimRow(sheetData *xlsxSheetData) []xlsxRow {
-	var (
-		row xlsxRow
-	)
+	var row xlsxRow
+
 	nextIndex := 0
 	for k := range sheetData.Row {
 		row = sheetData.Row[k]
--- a/sheet.go
+++ b/sheet.go
@@ -177,17 +177,19 @@ func (f *File) workSheetWriter() {
 
 // trimRow provides a function to trim empty rows.
 func trimRow(sheetData *xlsxSheetData) []xlsxRow {
-	var row xlsxRow
+	var (
+		row xlsxRow
+		i   int
+	)
 
-	nextIndex := 0
 	for k := range sheetData.Row {
 		row = sheetData.Row[k]
 		if row = trimCell(row); len(row.C) != 0 || row.hasAttr() {
-			sheetData.Row[nextIndex] = row
+			sheetData.Row[i] = row
 		}
-		nextIndex++
+		i++
 	}
-	return sheetData.Row[:nextIndex]
+	return sheetData.Row[:i]
 }
 
 // trimCell provides a function to trim blank cells which created by fillColumns.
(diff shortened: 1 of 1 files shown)


sheet.go
```diff
@@ -178,36 +178,38 @@ func (f *File) workSheetWriter() {
 // trimRow provides a function to trim empty rows.
 func trimRow(sheetData *xlsxSheetData) []xlsxRow {
 	var (
-		row  xlsxRow
-		rows []xlsxRow
+		row xlsxRow
```
xuri, line 181: This can be simplify to `var row xlsxRow`.

xuri (changes requested):

sheet.go
krstak: Correct. Done.

sheet.go
```diff
@@ -177,37 +177,38 @@ func (f *File) workSheetWriter() {
 
 // trimRow provides a function to trim empty rows.
 func trimRow(sheetData *xlsxSheetData) []xlsxRow {
-	var (
-		row  xlsxRow
-		rows []xlsxRow
-	)
-	for k, v := range sheetData.Row {
+	var row xlsxRow
```
xuri, line 180: Declare variable in the same way we be better (rename nextIndex to i):

```go
var (
    row xlsxRow
    i   int
)
```

xuri (changes requested):

sheet.go
krstak: Done.

xuri: Thanks for your great work. Could you run a new pprof figure to show how much memory was reduced in the optimization?

xuri (commented):
Thanks for your great work. Could you run a new pprof figure to show how much memory was reduced in the optimization?

krstak:
> Thanks for your great work. Could you run a new pprof figure to show how much memory was reduced in the optimization?

You are welcome.

I've already done it and pprof didn't show the problem anymore. It didn't show anything related to that problem. Since there are other potential memory consumption data issues, pprof pointed me to the new place that uses ~192MB. That place is in the function prepareSheetXML where we use append `ws.SheetData.Row = append(ws.SheetData.Row, ...)`

Since I am not 100% familiar with the code and it seems I would need to re-read a whole codebase to safely change that part, I didn't want to touch it. 

mem_problem

xuri (approved):

xuri:
Note that this PR introduce regression for issue #1383, fixed by issue #2335.

PR #1715, opened 2023-11-07
table-resize

gennesseaux:
# PR Details

Add function `ResizeTable` to resize a table instead of deleting and adding a table

## Motivation and Context

It seems easier to resize a table when needed instead of deleting the table and re-adding it

## How Has This Been Tested

I added the unit test for the methods I created.
I ran all the tests locally.

## Types of changes

## Checklist


gennesseaux:
Hello, is there any problem with this pull request or maybe you planned such a function ?

xuri (changes requested):
Thanks for your pull request. Sorry for my late reply. 

We needn't provide a sets of function for each properties of table, such `ReSizeTable`, `RenameTable`, `ReStyleTable`, etc., change `AddTable` to `SetTable` would be better approch, create table if table not exist, overwise update specified properties of table for the table: `Range`, `Name`, `StyleName`, etc., but this is a breaking changes, and we also need added `SourceName` and `TargetName` in the `Table` data type.

This library already support get table definiation and add table. Due to the principle of minimum availability, there is no plan to introduce new similar functions to keep the library core and easy.

PR #1816, opened 2024-02-14
Fix: Add OpenReader options to GetRows

vivekkairi:
# PR Details

The Options struct passed to OpenReader function is not passed in the GetRows which results in passing the options separately resulting in duplication 

## Description

Passing Options to OpenReader should automatically pass the same Options in the child function call as it is done initially. Currently the OpenReader options are not used in GetRows function, which causes inconsistency and leads to duplication. 

## Related Issue
https://github.com/qax-os/excelize/issues/1815

## Motivation and Context

It solves the duplicate need of passing Options if it's already passed in OpenReader

## How Has This Been Tested

Tested locally on my machine with multiple excel files and the output it gives satisfies the requirement

## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X ] I have read the **CONTRIBUTING** document.
- [ X] All new and existing tests passed.

### DIFF
--- a/col.go
+++ b/col.go
@@ -68,6 +68,7 @@ func (f *File) GetCols(sheet string, opts ...Options) ([][]string, error) {
 	}
 	cols, err := f.Cols(sheet)
 	results := make([][]string, 0, 64)
+	opts = append([]Options{*f.options}, opts...)
 	for cols.Next() {
 		col, _ := cols.Rows(opts...)
 		results = append(results, col)
@@ -92,6 +93,7 @@ func (cols *Cols) Rows(opts ...Options) ([]string, error) {
 	if cols.stashCol >= cols.curCol {
 		return rowIterator.cells, rowIterator.err
 	}
+	opts = append([]Options{*cols.f.options}, opts...)
 	cols.rawCellValue = getOptions(opts...).RawCellValue
 	if cols.sst, rowIterator.err = cols.f.sharedStringsReader(); rowIterator.err != nil {
 		return rowIterator.cells, rowIterator.err
--- a/col.go
+++ b/col.go
@@ -93,8 +93,7 @@ func (cols *Cols) Rows(opts ...Options) ([]string, error) {
 	if cols.stashCol >= cols.curCol {
 		return rowIterator.cells, rowIterator.err
 	}
-	opts = append([]Options{*cols.f.options}, opts...)
-	cols.rawCellValue = getOptions(opts...).RawCellValue
+	cols.rawCellValue = cols.f.getOptions(opts...).RawCellValue
 	if cols.sst, rowIterator.err = cols.f.sharedStringsReader(); rowIterator.err != nil {
 		return rowIterator.cells, rowIterator.err
 	}
--- a/col.go
+++ b/col.go
@@ -68,7 +68,6 @@ func (f *File) GetCols(sheet string, opts ...Options) ([][]string, error) {
 	}
 	cols, err := f.Cols(sheet)
 	results := make([][]string, 0, 64)
-	opts = append([]Options{*f.options}, opts...)
 	for cols.Next() {
 		col, _ := cols.Rows(opts...)
 		results = append(results, col)
--- a/rows.go
+++ b/rows.go
@@ -63,8 +63,6 @@ func (f *File) GetRows(sheet string, opts ...Options) ([][]string, error) {
 		return nil, err
 	}
 	rows, _ := f.Rows(sheet)
-	// Adding the OpenReader function options
-	opts = append([]Options{*f.options}, opts...)
 	results, cur, max := make([][]string, 0, 64), 0, 0
 	for rows.Next() {
 		cur++
--- a/rows.go
+++ b/rows.go
@@ -153,8 +153,7 @@ func (rows *Rows) Columns(opts ...Options) ([]string, error) {
 	}
 	var rowIterator rowXMLIterator
 	var token xml.Token
-	opts = append([]Options{*rows.f.options}, opts...)
-	rows.rawCellValue = getOptions(opts...).RawCellValue
+	rows.rawCellValue = rows.f.getOptions(opts...).RawCellValue
 	if rows.sst, rowIterator.err = rows.f.sharedStringsReader(); rowIterator.err != nil {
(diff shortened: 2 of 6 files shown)


xuri (changes requested):
Currently, the `OpenReader` options are not used in the `GetRows` function, the `GetRows` function just uses the same data structure `Options` as the `OpenReader` function to specify the `RawCellValue`. I agree with making the cell value reading functions inherit the `Options` settings of the `OpenReader` function, and we need to do some other modifications:

1. If `Options` have also been passed when calling the cell value reading functions, its priority should be higher than `OpenReader` and override its settings, so the code needs to be changed to:

```diff
-opts = append(opts, *f.options)
+opts = append([]Options{*rows.f.options}, opts...)
```

2. The following cell reading functions also need to inherit the Options settings:

- [`Columns`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/rows.go#L154)
- [`GetCols`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/col.go#L72)
- [`Rows`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/col.go#L95)
- [`GetCellValue`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/cell.go#L75)

3. The unit tests covered these changes

vivekkairi:
> ```diff
> -opts = append(opts, *f.options)
> ```

Hey

I've made the code changes, can you explain bit more on unit tests?

vivekkairi:
> Currently, the `OpenReader` options are not used in the `GetRows` function, the `GetRows` function just uses the same data structure `Options` as the `OpenReader` function to specify the `RawCellValue`. I agree with making the cell value reading functions inherit the `Options` settings of the `OpenReader` function, and we need to do some other modifications:
> 
> 1. If `Options` have also been passed when calling the cell value reading functions, its priority should be higher than `OpenReader` and override its settings, so the code needs to be changed to:
> 
> ```diff
> -opts = append(opts, *f.options)
> +opts = append([]Options{*rows.f.options}, opts...)
> ```
> 
> 2. The following cell reading functions also need to inherit the Options settings:
> 
> * [`Columns`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/rows.go#L154)
> * [`GetCols`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/col.go#L72)
> * [`Rows`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/col.go#L95)
> * [`GetCellValue`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/cell.go#L75)
> 
> 3. The unit tests covered these changes

Hey

I've made the code changes, can you explain bit more on unit tests?

vivekkairi:
> Currently, the `OpenReader` options are not used in the `GetRows` function, the `GetRows` function just uses the same data structure `Options` as the `OpenReader` function to specify the `RawCellValue`. I agree with making the cell value reading functions inherit the `Options` settings of the `OpenReader` function, and we need to consider following 2 conditions:
> 
> 1. If `Options` have also been passed when calling the cell value reading functions, its priority should be higher than `OpenReader` and override its settings, so the code needs to be changed to:
> 
> ```diff
> -opts = append(opts, *f.options)
> +opts = append([]Options{*rows.f.options}, opts...)
> ```
> 
> 2. The following cell reading functions also need to inherit the Options settings:
> 
> * [`Columns`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/rows.go#L154)
> * [`GetCols`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/col.go#L72)
> * [`Rows`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/col.go#L95)
> * [`GetCellValue`](https://github.com/qax-os/excelize/blob/9cbe3b6bd085a5b45ade85ba70ef578ab31a0b0f/cell.go#L75)
> 
> Based on the above two points, to avoid append options duplicated, I suggest changing the `getOptions` function to support inheriting the `Options` settings of the `OpenReader` function, just like this:
> 
> ```diff
> -func getOptions(opts ...Options) *Options {
> -    options := &Options{}
> +func (f *File) getOptions(opts ...Options) *Options {
> +    options := f.options
> ```
> 
> 3. The unit tests covered these changes

I've done the code changes, the current tests are passing. Could you guide me through the unit tests once?

col.go
```diff
@@ -68,6 +68,7 @@ func (f *File) GetCols(sheet string, opts ...Options) ([][]string, error) {
 	}
 	cols, err := f.Cols(sheet)
 	results := make([][]string, 0, 64)
+	opts = append([]Options{*f.options}, opts...)
```
xuri, line 71: This statement is not needed now.

rows.go
```diff
@@ -63,6 +63,8 @@ func (f *File) GetRows(sheet string, opts ...Options) ([][]string, error) {
 		return nil, err
 	}
 	rows, _ := f.Rows(sheet)
+	// Adding the OpenReader function options
+	opts = append([]Options{*f.options}, opts...)
```
xuri, line 67: Same as above. This statement is not needed now.

xuri (changes requested):

vivekkairi:
Fixed, also removed multiple times options fetch in CalcCellValue

col.go
vivekkairi: Done

xuri (approved):

vivekkairi:
Hey @xuri 

I am using v2 version of excelize, do I need to send a separate PR for that?

xuri:
The next version will be released based on the master branch. You can upgrade to the master branch code by `go get -u github.com/xuri/excelize/v2@master` before that.

Issue #1836, opened 2024-03-02
Flamegraph of `CalcCellValue` shows lots of time spent doing `runtime.duffcopy` inside `getSharedFormula`


paolobarbolini:
**Description**

`runtime.duffcopy` is showing up taking a substantial amount of time in flamegraphs of `CalcCellValue`.

![image](https://github.com/qax-os/excelize/assets/6215781/1b9b9e53-130e-4898-a743-31b617f9003c)

**Steps to reproduce the issue:**
1. Open an Excel containing many formulas
2. Call `CalcCellValue` on a cell containing a formula referencing many other cells
3. The flamegraph will show `runtime.duffcopy` taking a long time

**Output of `go version`:**

```text
go version go1.22.0 linux/amd64
```

**Excelize version or commit ID:**

```text
9d4c2e60f66b
```

xuri:
Thanks for your issue. The pull request has been merged, and this patch will be released in the next version.

PR #1837, opened 2024-03-02
Optimize `getSharedFormula` to avoid `runtime.duffcopy`

paolobarbolini:
# PR Details

Avoids large copies when dealing with `getSharedFormula`, which is called by `CalcCellValue`, increasing the performance of evaluating formulas in cells.

## Description

According to https://stackoverflow.com/a/45786922 uses of `range` in `getSharedFormula` caused large copies of objects, which would show-up in flamegraphs as `runtime.duffcopy`. Changing to a loop with indexing avoids the copying.

## Related Issue

#1836

## Motivation and Context

It cuts out about 40% of execution time from `CalcCellValue`.

Compared to the one in the issue, the flamegraph now looks like this:

image

## How Has This Been Tested

I've run the patched version inside my program and observed the calculations work as they did before.
I've also run `go test`.

## Types of changes

## Checklist

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -1655,8 +1655,10 @@ func parseSharedFormula(dCol, dRow int, orig []byte) (res string, start int) {
 // Note that this function not validate ref tag to check the cell whether in
 // allow range reference, and always return origin shared formula.
 func getSharedFormula(ws *xlsxWorksheet, si int, cell string) string {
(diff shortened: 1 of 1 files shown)


xuri (approved):
LGTM, thanks for your contribution and sponsorship.

PR #1857, opened 2024-03-20
Add support for getting cell-embedded pictures/images

msackman:
GetCellPicture gets a picture that is embedded within a cell. Excel itself (both standalone, and as part of Office 365) supports images being embedded within a cell, as well as images floating over one or more cells (supported by GetPictures and GetPictureCells).


msackman:
Thank you for looking at this. I'm not a massive fan of this API though: as a consumer of this API, I think it's important to know what type of image I'm accessing - I want to take different actions depending on whether the image is embedded within the cell, or whether it's floating. By combining all these APIs together, I'm not able to do that.

xuri (approved):
Thanks for your pull request. I have made some changes based on your code and made the existing functions `GetPictures` and `GetPictureCells` support for getting the embedded cell image created by Microsoft 365, instead of introducing a new exported function. After this PR is merged, I suggest that add a new field in the `Picture` data structure to indicate the type of the picture.

msackman:
> I suggest that add a new field in the `Picture` data structure to indicate the type of the picture.

Thank you - that would meet my use case well.

xuri:
There is another way to detect if the picture is an embedded cell image: get the picture cells by the GetPictureCells function at first, then get the cell's formula function by `GetCellFormula`, if you got the formula error `#VALUE!`, which represents the image was a cell image. We can avoid introducing new fields in this way to keep the library simple and core. What do you think about it?

msackman:
> What do you think about it?

In general, I believe code should be as explicit, obvious, and hard-to-misinterpret as possible. Using the magical `#VALUE!` seems non-obvious to me -- it requires plenty of prior-knowledge and is not particularly discoverable. So if it were me, I would introduce some sort of enum-like-thing to indicate the source of the image, and add a suitable field to the Picture struct. I think that would be more obvious and easier to work with.

xuri:
Thanks for your feedback. I will add support for that recently.

xuri:
The new field `InsertType` with `PictureInsertType` data type has been added in the `Picture` structure by pull request #1864, Please upgrade to the master branch code by `go get -u github.com/xuri/excelize/v2@master`, this feature will be released in the next version. In addition, the `GetPictures` and `GetPictureCells` functions still don't support getting cell images inserted by the `IMAGE` formula function currently, I'll consider adding support for that later.

xuri:
The Excelize library now support get the cell images inserted by IMAGE formula function.

msackman:
> The Excelize library now support get the cell images inserted by IMAGE formula function.

Awesome, thank you - yep, already using that feature. All working well.

PR #1935, opened 2024-07-02
Optimize ColumnNumberToName function performance, reduce about 50% memory usage and 50% time cost

zhayt:
# PR Details

**Optimizing memory usage by pre-allocation.**

## Description

1. **ColumnNumberToName Function**:
   - The `ColumnNumberToName` function previously concatenated rows using the `+` operator, leading to significant memory allocation when called frequently. 
   - The function now uses a pre-allocated slice to build the Excel column name. During construction, the converted number to letter is added to this slice. At the end, the slice of runes is converted to a string and returned.

2. **Sheet Pre-allocation**:
   - Added an option to create a Sheet with pre-allocated rows. 
   - The `xlsxWorksheet` structure now contains a field with a slice type. When calling the `SetCell...` function to add a row to the sheet, it in turn calls the `prepareSheetXML` function. This function adds an `xlsxRow` structure to the slice. 
   - Previously, multiple calls to `prepareSheetXML` caused the slice to grow, allocating a lot of memory and performing unnecessary copying. The new approach pre-allocates the maximum number of rows, reducing memory allocations and copying overhead.

## Related Issue

- Resolves Issue #XXX

## Motivation and Context

- When working with a lot of data, some functions create many objects and trigger garbage collection (GC). The GC then loads the processor.
- This optimization reduces memory allocations and the associated GC overhead, improving performance.

## How Has This Been Tested

- The changes have been tested in a development environment with datasets of varying sizes.
- Benchmark tests were added to measure performance improvements in memory usage and processing time.
- Existing unit tests were run to ensure no regressions were introduced.

## Types of changes

## Checklist

---

Feel free to modify any part of this to better fit your needs!

### DIFF
--- a/lib.go
+++ b/lib.go
@@ -237,10 +237,10 @@ func ColumnNumberToName(num int) (string, error) {
 		estimatedLength++
 	}
 
-	result := make([]rune, estimatedLength)
+	result := make([]byte, estimatedLength)
 	for num > 0 {
 		estimatedLength--
-		result[estimatedLength] = rune((num-1)%26 + 'A')
+		result[estimatedLength] = byte((num-1)%26 + 'A')
 		num = (num - 1) / 26
 	}
 	return string(result), nil
--- a/lib.go
+++ b/lib.go
@@ -232,12 +232,18 @@ func ColumnNumberToName(num int) (string, error) {
 	if num < MinColumns || num > MaxColumns {
 		return "", ErrColumnNumber
 	}
-	var col string
+	estimatedLength := 0
+	for n := num; n > 0; n = (n - 1) / 26 {
+		estimatedLength++
+	}
+
+	result := make([]rune, estimatedLength)
 	for num > 0 {
-		col = string(rune((num-1)%26+65)) + col
+		estimatedLength--
+		result[estimatedLength] = rune((num-1)%26 + 'A')
 		num = (num - 1) / 26
 	}
-	return col, nil
+	return string(result), nil
 }
 
 // CellNameToCoordinates converts alphanumeric cell name to [X, Y] coordinates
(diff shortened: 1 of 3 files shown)


lib.go
```diff
@@ -232,12 +232,18 @@ func ColumnNumberToName(num int) (string, error) {
 	if num < MinColumns || num > MaxColumns {
 		return "", ErrColumnNumber
 	}
-	var col string
+	estimatedLength := 0
```
xuri, line 235: Could you provides benchmark data for this performance improvements?

sheet.go
```diff
@@ -30,11 +30,21 @@ import (
...
+		}
+	}
+}
+
 // NewSheet provides the function to create a new sheet by given a worksheet
 // name and returns the index of the sheets in the workbook after it appended.
 // Note that when creating a new workbook, the default worksheet named
 // `Sheet1` will be created.
-func (f *File) NewSheet(sheet string) (int, error) {
+func (f *File) NewSheet(sheet string, options ...Option) (int, error) {
```
xuri, line 47: This function allows users to set cell values ​​randomly when they can't sure the total number of rows. If user know how many rows will be written and data is in ordered, they can using stream writer for get much lower cost. So I doesn't suggest changing the function signature.

xuri (changes requested):

lib.go
zhayt: I wrote a benchmark test for the old and new version of the function, and I also made some changes (replacing a slice of runes with a slice of bytes).

[bencgcmp_result.txt](https://github.com/user-attachments/files/16086822/bencgcmp_result.txt)

sheet.go
zhayt: I saw that using StreamWriter did not help the person who reported issue 876. Maybe I should use both options under the same conditions and compare the results?

xuri (changes requested):
I suggest that we separate these two improvements into 2 PRs, and I will take a reconsider about the design of reallocate rows for worksheets.

zhayt:
Sure, I will separate these two changes into different PRs. Do I need to create two new PRs, or can I keep the changes related to the ColumnNumberToName function in this PR?

Thanks for reviewing!

xuri:
You can keep the changes related to the `ColumnNumberToName` function in this PR. Thanks.

zhayt:
Done, I left in this PR only the changes related to the ColumnNumberToName function. Thanks.

xuri (approved):

xuri (approved):

PR #1953, opened 2024-07-16
This fix missing horizontal axis in scatter chart with negative values

pjh591029530:
This fix missing horizontal axis in scatter chart with negative values


xuri (approved):

Issue #1936, opened 2024-07-02
`CalcCellValue` silently fails when formula is `=OTHER_CELL` and OTHER_CELL has a formula evaluation error


paolobarbolini:
**Description**

`CalcCellValue` returns an empty string, with no error, if the evaluated `cell` references another cell (`=OTHER_CELL`) that contains a formula that fails to evaluate.

**Steps to reproduce the issue:**

1. Create a new Excel file
2. Write a number in `A1`
3. Write a string in `B1`
4. Write `=A1+B1` in `C1`
5. Write `=C1` in `D1`
6. Call `CalcCellValue` on `D1`

Repository: I'll edit the description in a few minutes with a link to the repro in the form of a repository

**Describe the results you received:**

An empty string is returned, with no error

**Describe the results you expected:**

A formula evaluation error

**Output of `go version`:**

```text
go version go1.22.4 linux/amd64
```

**Excelize version or commit ID:**

```text
4e6457accde38876ecdf8bb1c6f1e8bce8fd2139
```

**Environment details (OS, Microsoft Excel™ version, physical, etc.):**

Arch Linux with LibreOffice Calc v24.2.4-2

PR #1962, opened 2024-07-22
Fix: Issue #1936 CalcCellValue silently fails when formula is =OTHER_FORMULA

abdelaziz-ouhammou:
Fix: Issue #1936 CalcCellValue silently fails when formula is =OTHER_FORMULA

# PR Details

Return Error if it doesn't have "#N/A"

## Related Issue
 CalcCellValue silently fails when formula is =OTHER_CELL and OTHER_CELL has a formula evaluation error #1936 

## How Has This Been Tested

I added a unit test for the changes.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [X] I have added tests to cover my changes.


paolobarbolini:
What is missing for this to be merged?

xuri (changes requested):
Thanks for your contribution, sorry for my late reply. This fix can be mitigated for the issue, but can not resolve the root cause of this issue. I think we need to let the [`parseToken`](https://github.com/xuri/excelize/blob/d21b598235769617c4bfa3c4b845d764027ed348/calc.go#L1446) function return value to `formulaArg` data type, the error context was missing caused by `calcAdd`, `calcG`, `calcGe` series function return value data type conversion.

In addition, I don't suggest add more test workbook files, generate the workbook by unit tests code would be better.

Issue #1076, opened 2021-11-29
 How to adjust the order of sheets？


iseejustdoit:
Adjust Sheet2 before Sheet1

xuri:
Thanks for your issue, this library not support to adjust the order of sheets directly, you can rename sheet with [`SetSheetName`](https://xuri.me/excelize/en/sheet.html#SetSheetName), note that it won't be update the sheet name in the formula or reference associated with the cell. So there may be problem formula error or reference missing.

xuri:
New function `MoveSheet` has been added, this function will be available on the next released version.

xuri:
Hi @RichardHill, the [`MoveSheet`](https://pkg.go.dev/github.com/xuri/excelize/v2#File.MoveSheet) function already released in the [version 2.9.0](https://github.com/xuri/excelize/releases/tag/v2.9.0). Please upgrade to the lasted version.

PR #1996, opened 2024-09-24
#1076 Add MoveSheet and SwapSheets functions to support changing sheet order in the workbook 

Zncl2222:
# PR Details

Implement MoveSheet and SwapSheets function to change the order of sheet in the workbook.

- MoveSheet: Moves the sheet to the new position using its current index and the target index. The other sheets will be shifted to the left or the right.

- SwapSheets: Swaps the positions of two sheets in the workbook based on their indices.

## Description

Both `MoveSheet` and `SwapSheets` functions use the index as their input arguments instead of the `SheetName`. This is because `SheetName` is allowed to be duplicated within a workbook. To avoid potential confusion or conflicts caused by duplicate sheet names, the implementation relies on the sheet index.

Here is an example code snippet to reorder the sheet

```golang
package main

import (
	"fmt"
	"strconv"

	"github.com/xuri/excelize/v2"
)

func main() {
	f := excelize.NewFile()
	defer func() {
			fmt.Println(err)
		}
	}()
	for i := 2; i < 10; i++ {
		f.NewSheet("Sheet" + strconv.Itoa(i))
       }
	if err := f.SaveAs("BeforeOrder.xlsx"); err != nil {
		fmt.Println(err)
	}
	if err := f.MoveSheet(7, 0); err != nil {
		fmt.Println(err)
	}
	if err := f.SaveAs("AfterOrder.xlsx"); err != nil {
		fmt.Println(err)
	}
}
```

### Before Order
image

### After Order
image

## Related Issue
#1076

## Motivation and Context

Allow users to use this package more flexibly. Additional this can resolve the problem that [`SetSheetName`](https://xuri.me/excelize/en/sheet.html#SetSheetName) can't update the sheet name in the formula or reference associated with the cell. 

## Types of changes

## Checklist


workbook.go
```diff
@@ -130,6 +130,89 @@ func (f *File) UnprotectWorkbook(password ...string) error {
...
+
+	wb, err := f.workbookReader()
+	if err != nil {
+		return err
+	}
+	if sheet1Idx >= f.SheetCount || sheet2Idx >= f.SheetCount || sheet1Idx < 0 || sheet2Idx < 0 {
+		return ErrSheetIdx
+	}
+
+	// Unselect the tab of a sheet by index
```
xuri: Move sheet may break grouped sheets, so we just ungroup all sheets instead of unselecting each sheet, so the `unselectSheetTab` function no longer used.

PR #2018, opened 2024-11-03
Add a way to set graphics in a header/footer

imirkin:
The full functionality of the VML drawing is not addressed here. However this seems to work for embedding regular images.

# PR Details

Adds the ability to set header/footer graphics.

## Description

This is a somewhat restricted implementation out of all the options possible. But it's enough to get the basics going, and I doubt anyone out there wants too much more than this offers -- image + width/height for scaling.

If this function is called twice, it won't "update" the existing bits, but rather just leave them around and make a new drawing / media image. I don't think there's any process to "GC" the left-overs, but it's also not completely clear what to do about this. The media object may be shared by multiple drawings / sheets that have the same image.

## How Has This Been Tested

This is based on what MS Excel produces. I tried other variations but was mostly unsuccessful. I had some variations which produced a border (without using the formulas bit), and many which crashed Excel -- seems like the logic to processing these is a bit fragile. Not 100% sure what all is needed since there are a lot of permutations, but with everything I have, it seems to work OK.

## Types of changes

## Checklist

I wasn't too sure how to test the change, so I added an example.

### DIFF
--- a/picture.go
+++ b/picture.go
@@ -295,6 +295,16 @@ func (f *File) addSheetLegacyDrawing(sheet string, rID int) {
 	}
 }
 
+// addSheetLegacyDrawingHF provides a function to add legacy drawing
+// header/footer element to xl/worksheets/sheet%d.xml by given
+// worksheet name and relationship index.
+func (f *File) addSheetLegacyDrawingHF(sheet string, rID int) {
+	ws, _ := f.workSheetReader(sheet)
+	ws.LegacyDrawingHF = &xlsxLegacyDrawingHF{
+		RID: "rId" + strconv.Itoa(rID),
+	}
+}
+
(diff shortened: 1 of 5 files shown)


xuri (approved):
Thanks for your PR. I updated unit tests and simplify code based on your code. Much appreciated.

imirkin:
Great. I wasn't 100% sure about the API, and I couldn't come up with a good Get variant. But I think it's still useful without that. I thought about putting it into the SetHeaderFooter call, but since there's no good way to Get it, I decided against it. Happy to iterate on the API if you think it still needs work though.

xuri:
Thanks for your contribution. I merged this PR. If you'd like to add `GetLegacyDrawingHF` function, that would be great.

PR #2022, opened 2024-11-07
Add PageOrder setting as part of PageLayoutOptions

imirkin:
This controls whether page numbers are assigned in a zig-zag going across first, or down first.

## How Has This Been Tested

Verified that opening in Excel shows the "over then down" value, whereas "down then over" is used by default.

## Types of changes

## Checklist

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -88,6 +88,9 @@ var (
 	// ErrOutlineLevel defined the error message on receive an invalid outline
 	// level number.
 	ErrOutlineLevel = errors.New("invalid outline level")
+	// ErrPageSetupAdjustTo defined the error message for receiving a page setup
+	// adjust to value exceeds limit.
+	ErrPageSetupAdjustTo = errors.New("adjust to value must be between 10 and 400")
 	// ErrParameterInvalid defined the error message on receive the invalid
 	// parameter.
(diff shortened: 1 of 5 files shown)


xuri (commented):
Thanks for your PR. LGTM. Your last PR exported new function `SetLegacyDrawingHF`, I think rename it to `SetHeaderFooterImage` would be better, how about you think? You can change it in this PR if you think so.

imirkin:
> Thanks for your PR. LGTM. Your last PR exported new function `SetLegacyDrawingHF`, I think rename it to `SetHeaderFooterImage` would be better, how about you think? You can change it in this PR if you think so.

I think that's fine, but I'd prefer to keep the changes separate. Can I send a different PR to update it?

xuri:
> > Thanks for your PR. LGTM. Your last PR exported new function `SetLegacyDrawingHF`, I think rename it to `SetHeaderFooterImage` would be better, how about you think? You can change it in this PR if you think so.
> 
> I think that's fine, but I'd prefer to keep the changes separate. Can I send a different PR to update it?

Of course, sure. Thanks.

xuri (approved):
I made some changes based on your branch, to reduce cyclomatic complexity, added value check and error message for each option.

PR #2023, opened 2024-11-07
Rename SetLegacyDrawingHF to SetHeaderFooterImage

imirkin:
Question: should `HeaderFooterGraphics` be renamed to something, or OK as-is?

# PR Details

Rename SetLegacyDrawingHF to SetHeaderFooterImage.

## Motivation and Context

This is being done per request from @xuri. Probably fits in better with the overall naming convention.

## How Has This Been Tested

`go test` passes. Given that this was generated with `sed -i`, no further testing has been done.

## Types of changes

## Checklist

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -1239,7 +1239,7 @@ func attrValToBool(name string, attrs []xml.Attr) (val bool, err error) {
 //	                        |
 //	 &F                     | Current workbook's file name
 //	                        |
-//	 &G                     | Drawing object as background (Use SetLegacyDrawingHF)
+//	 &G                     | Drawing object as background (Use SetHeaderFooterImage)
 //	                        |
 //	 &H                     | Shadow text format
 //	                        |
--- a/vml.go
+++ b/vml.go
@@ -1071,12 +1071,12 @@ func extractVMLFont(font []decodeVMLFont) []RichTextRun {
 	return runs
 }
 
-// SetLegacyDrawingHF provides a mechanism to set the graphics that
+// SetHeaderFooterImage provides a mechanism to set the graphics that
 // can be referenced in the Header/Footer defitions via &G.
 //
 // The extension should be provided with a "." in front, e.g. ".png".
 // The width/height should have units in them, e.g. "100pt".
-func (f *File) SetLegacyDrawingHF(sheet string, g *HeaderFooterGraphics) error {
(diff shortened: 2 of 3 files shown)


imirkin:
BTW, it looks like it may be possible to set multiple images somehow:

https://github.com/jmcnamara/XlsxWriter/blob/main/xlsxwriter/worksheet.py#L4244

Perhaps multiple shapes in the VML? No idea, I wasn't able to follow this code. I don't need this myself, but perhaps an improvement for later.

imirkin:
Well there's still only one drawing allowed, so probably have to set them
all at once. Otherwise you have to edit the vml.

xuri:
Yep, support append VML shapes (just like form controls) to allow call `SetHeaderFooterImage` multiple times, that's would be better.

vmlDrawing.go
```diff
@@ -193,15 +193,19 @@ type decodeVmlDrawing struct {
...
 // the file xl/drawings/vmlDrawing%d.vml.
 type decodeShapeType struct {
-	ID        string `xml:"id,attr"`
-	CoordSize string `xml:"coordsize,attr"`
-	Spt       int    `xml:"spt,attr"`
-	Path      string `xml:"path,attr"`
+	ID             string `xml:"id,attr"`
+	CoordSize      string `xml:"coordsize,attr"`
+	Spt            int    `xml:"spt,attr"`
+	PreferRelative string `xml:"preferrelative,attr,omitempty"`
```
imirkin, line 199: `o:preferrelative`, I think?

And above, `o:spt`

vmlDrawing.go
```diff
@@ -193,15 +193,19 @@ type decodeVmlDrawing struct {
...
+	Path           string `xml:"path,attr"`
+	Filled         string `xml:"filled,attr,omitempty"`
+	Stroked        string `xml:"stroked,attr,omitempty"`
 }
 
 // decodeShape defines the structure used to parse the particular shape element.
 type decodeShape struct {
 	ID          string `xml:"id,attr"`
+	SpID        string `xml:"spid,attr,omitempty"`
```
imirkin, line 208: `o:spid`, no?

xuri: This library using two structs to resolve limitation XML name space support of Go encoding/xml, no need namespace in the `decode` prefix structures.

Issue #2038, opened 2024-12-11
Conditional Formatting: support new Icon Sets, custom breakpoints


imirkin:
There's currently functionailty to support an IconSet on a conditional format. However if one doesn't want to use a built-in scheme, Excel allows a custom scheme instead where icons are individually picked out. This is currently not supported in Excelize. A sample xml block for such a conditional format rule might be

```xml
          <x14:cfRule type="iconSet" priority="7" id="{EFADC4A8-BD1B-48CB-BBAC-95556150EFE4}">
            <x14:iconSet showValue="0" custom="1">
              <x14:cfvo type="percent">
                <xm:f>0</xm:f>
              </x14:cfvo>
              <x14:cfvo type="num">
                <xm:f>0</xm:f>
              </x14:cfvo>
              <x14:cfvo type="num" gte="0">
                <xm:f>0</xm:f>
              </x14:cfvo>
              <x14:cfIcon iconSet="3Triangles" iconId="0"/>
              <x14:cfIcon iconSet="3Triangles" iconId="1"/>
              <x14:cfIcon iconSet="3Triangles" iconId="2"/>
            </x14:iconSet>
          </x14:cfRule>
```

I may or may not end up needing this, but wanted to log it in case others find it useful as a starting point.

imirkin:
Actually it looks like the reason for this explicitly custom icon set is dubious? Just need to make "3Triangles" a recognized icon set? But I also need the cutoffs to be negative/zero/positive rather than 0/33/67 cutoffs... maybe that's why I need the icon set to be custom? Just need to be able to pass in a list of xlsxCfRule somehow?

Based on https://learn.microsoft.com/en-us/dotnet/api/microsoft.office.interop.excel.xliconset?view=excel-pia it looks like the following were added at some point:

* `3Stars`
* `3Triangles`
* `5Boxes`

imirkin:
I've done a bit more research on this. Looks like the "new" icon sets can't simply be added. They must appear in the x14 namespace, which has to go in the extLst. The full thing might be

```xml
  <conditionalFormatting sqref="B1:B5">
    <cfRule type="iconSet" priority="1">
      <iconSet iconSet="5Quarters" showValue="0">
        <cfvo type="percent" val="0"/>
        <cfvo type="percent" val="20"/>
        <cfvo type="percent" val="40"/>
        <cfvo type="percent" val="60"/>
        <cfvo type="percent" val="80"/>
      </iconSet>
    </cfRule>
[...]

PR #2042, opened 2024-12-11
WIP add support for new icon sets

imirkin:
# PR Details

This adds support for the "new" icon sets (3 triangles, 3 stars, 5 boxes). These were introduced in a later revision of the spec, and thus need different namespaces/etc.

## Description

For the new icon sets, we add an x14-namespaced conditional formatting entry + rules. Note that unlike the data bar, these are entirely in the "new" land, so no need to add it in the old location + ext.

Still WIP:
* Strategy for priorities
* GUID generation? The Excel-generated samples have GUIDs, but seem to work without it
* Decoding

Looking for some feedback/suggestions.

## Related Issue

#2038

## How Has This Been Tested

I added the conditional format with all 3 "new" icon sets (as well as a data bar with solid=true just to make sure I'm not breaking anything existing), and it displayed fine in Excel.

## Types of changes

## Checklist


imirkin:
The problem here, BTW, is that priority is determined based on analyzing the list of `ws.ConditionalFormatting`. But this won't count the "x14" cfrules. So we have to re-parse it. There's a bit of proliferation of that parsing code, so it should get centralized somehow. But re-parsing it from scratch every time we add a conditional format rule will lead to O(n^2) behavior. (I guess that's already there, but adding some numbers isn't so heavy as parsing a bunch of xml each time.)

Suggestions welcome.

xuri (approved):
Thanks for your PR. I have made some changes based on your branch.

PR #2080, opened 2025-01-30
Handle the format '?' in number formatting

kurtinge:
# Handle the format '?' in number formatting

## Description

When excelize tries to read a cell where the formatCode in the style is "?" it will return an empty string instead of the number.

## Motivation and Context

This handles that the formatCode "?" is valid and returns a the number correctly

## How Has This Been Tested

This has been tested in the unit test and with an actual file.

## Types of changes

## Checklist


numfmt_test.go
```diff
@@ -3635,6 +3635,7 @@ func TestNumFmt(t *testing.T) {
 		{"123", "[$x.-unknown]MM/DD/YYYY", "x.05/02/1900"},
 		{"1234.5678", "0.0xxx00", "1234.5xxx68"},
 		{"80145.899999999994", "[$¥-8004]\" \"#\" \"####\"\"", "¥ 8 0146"},
+		{"1", "?", "1"},
```
xuri, line 3638: TODP: support blank padding. For example: 
```text
{"1", "???", "  1"},
```

xuri (approved):

xuri:
Thanks for your PR. The case as I commented on above should be supported.

numfmt_test.go
xuri: This TODO has been finished by PR #2216.

PR #2097, opened 2025-03-01
fix: return nil instead of error directly

huochexizhan:
# PR Details

## Description

return nil instead of error directly

## Types of changes

## Checklist


xuri (changes requested):
Thanks for your PR. Generally, error should be handled and return, any reason to suppress error? In addition, unit tests is required for any pull request changes.

Issue #2056, opened 2024-12-27
Very slow SUMIF performance when using "Row Range Notation"


grahamorrell:
**Description**

I have an Excel workbook (minimal example attached) which has some SUMIF( ) formulae that use "Row Range Notation", for example : 

```text
=SUMIF(2:2, B11, 4:4)
```

These perform very poorly in Excelize despite there only being 25 columns in the worksheet.  To calculate the sum of 8 of these SUMIF( ) statements takes over 3 seconds.

If I change the SUMIFs to also specify the columns, for example : 

```text
=SUMIF(B2:Y2, B11, B4:Y4)
```

Performance is dramatically improved, taking just 8 milliseconds to calculate the sum of 8 of them.

**Steps to reproduce the issue:**
1.  Call CalcCellValue on a SUMIF( ) using Row Range Notation

**Describe the results you received:**

![image](https://github.com/user-attachments/assets/050a72fb-7bce-4572-9dd8-e57cb7419c17)

**Describe the results you expected:**

Performance should be almost identical in both cases because the sheet only has 25 columns.

**Output of `go version`:**

```text
N/A
```

**Excelize version or commit ID:**

```text
excelize-wasm 0.0.7
```

Windows
[SUMIF Performance.xlsx](https://github.com/user-attachments/files/18261192/SUMIF.Performance.xlsx)

paolobarbolini:
I'm trying to fix this in #2111. Not all tests pass yet but I'd expect the final implementation to fix your issue without having to manually restrict the range.

xuri:
I fixed incorrect formula calculation result for cell `"Test SUMIF"!D16` in workbook "SUMIF Performance.xlsx". Issue caused by shared formula parsed error. This patch will be released on next version, and excelize-wasm NPM package will be updated soon.

PR #2111, opened 2025-03-27
Avoid looking up absent cells in `rangeResolver`

paolobarbolini:
# PR Details

While benchmarking the performance of `SUMIFS` I've discovered that when a range references an entire column or an entire row, excelize scans the entire theoretical range even if the file is smaller than that. This was already discovered by #2056.

## Description

I've initially tried to reduce the scan range, but a lot of the code assume that `rangeResolver` is going to be scanning the entire range, so I've changed it to instead skip looking up cells that don't exist.

## Related Issue

#2056

## Motivation and Context

This makes a big performance difference.

## How Has This Been Tested

I've run tests and also tested this on my projects.

## Types of changes

## Checklist

---

PR developed together with @Aprax14.

### DIFF
--- a/calc.go
+++ b/calc.go
@@ -1690,16 +1690,31 @@ func (f *File) rangeResolver(ctx *calcContext, cellRefs, cellRanges *list.List) 
 	// extract value from ranges
 	if cellRanges.Len() > 0 {
 		arg.Type = ArgMatrix
+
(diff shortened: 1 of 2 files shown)


xuri (approved):
LGTM, thanks for your contribution. I added unit test based on your branch.

PR #2116, opened 2025-04-16
improve cell read performance by optimizing xml parsing

shcabin:
# PR Details

improve cell read performance by optimizing  unmarshal for xlsxC struct

## Motivation and Context

Improve the cell reading performance of large documents

## How Has This Been Tested

Performance testing was conducted on an AMD64 WSL using Go's benchmark tool. The benchmarks were run BenchmarkRows 10 times each to ensure statistical significance. The following results were observed:
goos: linux
goarch: amd64
pkg: github.com/xuri/excelize/v2
cpu: 12th Gen Intel(R) Core(TM) i5-12400F
```
        │ bashline_1.txt │              new_1.txt              │
        │     sec/op     │   sec/op     vs base                │
Rows-12      164.7µ ± 5%   131.1µ ± 2%  -20.41% (p=0.000 n=10)

        │ bashline_1.txt │              new_1.txt              │
        │      B/op      │     B/op      vs base               │
Rows-12     72.20Ki ± 0%   66.44Ki ± 0%  -7.99% (p=0.000 n=10)

        │ bashline_1.txt │              new_1.txt              │
        │   allocs/op    │  allocs/op   vs base                │
Rows-12      1.903k ± 0%   1.658k ± 0%  -12.87% (p=0.000 n=10)
```

Due to the small size of Book1.xlsx, I replaced it with TestStreamWriter.xlsx(Get after run TestStreamWriter) and re-executed it:
```
       │ bashline_2.txt │             new_2.txt              │
        │     sec/op     │   sec/op    vs base                │
Rows-12       5.662 ± 3%   3.970 ± 3%  -29.88% (p=0.000 n=10)

        │ bashline_2.txt │              new_2.txt               │
        │      B/op      │     B/op      vs base                │
Rows-12     2.387Gi ± 0%   2.044Gi ± 0%  -14.38% (p=0.000 n=10)

        │ bashline_2.txt │              new_2.txt              │
        │   allocs/op    │  allocs/op   vs base                │
Rows-12      64.82M ± 0%   52.02M ± 0%  -19.75% (p=0.000 n=10)
```
## Types of changes

## Checklist

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -251,11 +251,13 @@ func (cell *xlsxC) cellXMLAttrHandler(start *xml.StartElement) error {
 		case "r":
 			cell.R = attr.Value
 		case "s":
-			i64, err := strconv.ParseInt(attr.Value, 10, 64)
+			val, err := strconv.ParseInt(attr.Value, 10, 64)
 			if err != nil {
 				return err
 			}
-			cell.S = int(i64)
+			if int64(math.MinInt) <= val && val <= int64(math.MaxInt) {
+				cell.S = int(val)
+			}
 		case "t":
 			cell.T = attr.Value
 		default:
--- a/rows.go
+++ b/rows.go
@@ -244,40 +244,42 @@ func (rows *Rows) rowXMLHandler(rowIterator *rowXMLIterator, xmlElement *xml.Sta
 	}
 }
 
-func (cell *xlsxC) cellXMLHandler(decoder *xml.Decoder, start *xml.StartElement) (err error) {
-	cell.XMLName = start.Name
+func (cell *xlsxC) cellXMLAttrHandler(start *xml.StartElement) (err error) {
+	var res uint64
 	for _, attr := range start.Attr {
 		if attr.Name.Local == "space" {
 			cell.XMLSpace = attr
 		} else if attr.Name.Local == "r" {
 			cell.R = attr.Value
 		} else if attr.Name.Local == "s" {
-			cell.S, err = strconv.Atoi(attr.Value)
+			var i64 int64
+			i64, err = strconv.ParseInt(attr.Value, 10, 64)
+			cell.S = int(i64)
 		} else if attr.Name.Local == "t" {
 			cell.T = attr.Value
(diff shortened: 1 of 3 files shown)


rows.go
```diff
@@ -244,6 +244,74 @@ func (rows *Rows) rowXMLHandler(rowIterator *rowXMLIterator, xmlElement *xml.Sta
...
+	for _, attr := range start.Attr {
+		if attr.Name.Local == "space" {
+			cell.XMLSpace = attr
+		} else if attr.Name.Local == "r" {
+			cell.R = attr.Value
+		} else if attr.Name.Local == "s" {
+			cell.S, err = strconv.Atoi(attr.Value)
+		} else if attr.Name.Local == "t" {
+			cell.T = attr.Value
+		} else if attr.Name.Local == "cm" {
```
xuri, line 258: Unit tests is required, the line code coverage of changes should be 100%.

rows.go
```diff
@@ -244,6 +244,74 @@ func (rows *Rows) rowXMLHandler(rowIterator *rowXMLIterator, xmlElement *xml.Sta
...
+			}
+			vm := uint(res)
+			cell.Vm = &vm
+		} else if attr.Name.Local == "ph" {
+			r, err := strconv.ParseBool(attr.Value)
+			if err != nil {
+				return err
+			}
+			cell.Ph = &r
+		}
```
xuri, line 278: Should we handle the `space` XML attribute at here?

xuri (changes requested):

rows.go
shcabin: "Cm" "Ph" is not directly used, is it still needed here?
It seems that the streaming method does not support obtaining embedded image cells, so the "Vm" is not  cover

xuri: Yeah, after check the usage of these attributes, I removed these based from your code.

rows.go
xuri: Same as above, the `space` not required at here.

xuri (approved):
Thank you, excellent work! I made some changes based on your branch.

Issue #2121, opened 2025-04-27
Use mmap if possible (can save memory with big files)


tgulacsi:
### Description

ReadAll is used only in a few places, but most dominantly it is used to slurp up all the .xlsx we are opening.
It may be only a few MiBs, but mmap is an easy solution for this.
 

### Steps to reproduce the issue

1. open a big .xlsx
2. Check used memory and GC churn

### Describe the results you received

At least the size of .xlsx is used, then later may be released.

### Describe the results you expected

Not a lot of memory is used, GC is in peace

### Go version

1.24.1

### Excelize version or commit ID

2.9.0

### Environment

```shell
Debian Linux testing
```

### Validations

- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a minimal reproducible example of the bug.

shcabin:
Have you compared it with io.Copy?  It seems that io.Copy is more compatible and not fallback to io.ReadAll().

In my work environment there is only Fd() uintptr （not found Fd() int), and I am not sure why, so I am unable to conduct effective testing mmap.

PR #2122, opened 2025-04-27
Use mmap in ReadAll if possible

tgulacsi:
# PR Details

Use mmap instead of io.ReadAll, if possible

## Description

Introduce readAll function which uses mmap if possible, io.ReadAll otherwise.

## Related Issue

#2121 

## Motivation and Context

Memory inefficiency.

## How Has This Been Tested

go test

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist


shcabin:
Have you compared it with io.Copy? It seems that io.Copy is more compatible and not fallback to io.ReadAll().

In my work environment there is only Fd() uintptr （not found Fd() int), and I am not sure why, so I am unable to conduct effective testing mmap.

tgulacsi:
My bad, you're right, Fd() returns uintptr.

This mmap is not for speed, it's for decreasing RSS by using the kernel-supplied probably-in-filecache []byte of the file.

Issue #2001, opened 2024-09-26
Possible bugs in AddPictureFromBytes method.


Sirius2020:
**Code to reproduce the issue**
```
package main

import (
	_ "image/jpeg"
	"os"
	"path/filepath"

	"github.com/xuri/excelize/v2"
)

func addPictureToA1WithOffset(f *excelize.File, data []byte, ext string, offset int) error {
	enable := true
	disable := false
	if err := f.AddPictureFromBytes("Sheet1", "A1", &excelize.Picture{Extension: ext, File: data, Format: &excelize.GraphicOptions{
		PrintObject: &enable,
		Locked:      &disable,
		ScaleX:      1,
		ScaleY:      1,
		OffsetX:     offset,
		Positioning: "oneCell",
	}}); err != nil {
		panic(err)
	}
	return nil
}

func main() {
	f := excelize.NewFile()
	defer func() {
		if err := f.Close(); err != nil {
			panic(err)
		}
	}()
	f.SetColWidth("Sheet1", "A", "H", 20)
	imagePath := "./pic.jpg" // pixel 60*60
	ext := ".jpg"
	data, _ := os.ReadFile(filepath.Clean(imagePath))

	addPictureToA1WithOffset(f, data, ext, 0)
	addPictureToA1WithOffset(f, data, ext, 60)
	addPictureToA1WithOffset(f, data, ext, 120)
	addPictureToA1WithOffset(f, data, ext, 180)

	if err := f.SaveAs("result.xlsx"); err != nil {
		panic(err)
	}
}
```

**Description**
In the code example below, I try to add 4 60x60 sized pictures into cell A1, I'm expecting the pictures to maintain the size of 60x60 and be arranged horizontally end to end.
But there are two problems occur.
1. The size of the pictures is incorrect, the first and second picture size is 90x100, while the third picture size is 112x100.
2. The fourth picture is invisible.
<img width="380" alt="image" src="https://github.com/user-attachments/assets/b9661edf-1ec6-4af1-bade-c291d10bac88">

The fourth picture is misplaced and I believe it's a bug. As the debug screenshot show below, in method addDrawingPicture, from.Col is equal to to.Col while from.ColOff(1714500) is larger than to.ColOff(895350).

<img width="1318" alt="image2" src="https://github.com/user-attachments/assets/134b1636-7ec6-43c2-8e52-6bc56892cc8c">

**Output of go version:**
```
go version go1.20.6 linux/amd64
```

[...]

PR #2123, opened 2025-04-28
Fixed drawingResize dimensions calculation

R3dByt3:
# PR Details

The current implementation does not seem to be in accordance to the specification:
https://github.com/closedxml/closedxml/wiki/Cell-Dimensions
https://github.com/closedxml/closedxml/wiki/Cell-Dimensions#mdw
https://github.com/closedxml/closedxml/wiki/Cell-Dimensions#skiasharp-code

Also there seem to be logical issues as well. This commit solves an identical issue as linked below.

## Description

Looking at the C# ClosedXML Excel file Wiki:
- The conversion from pt to px was incorrect (row height).
- The default conversion from MDW to px were incorrect (column width - there is no margin) - also this calculation should depend on the used font; To be honest I have not checked if and how this could be achieved within the project.
- The aspect ratio calculation has been simplified
- The offset values were subtracted from the image dimensions which is simply not correct

## Related Issue

https://github.com/qax-os/excelize/issues/2001

## Motivation and Context

In my use case I encountered the problems mentioned in the issue #2001 and these changes applied locally fixed my issues.

## How Has This Been Tested

Try to add multiple images into a single cell using offsets as mentioned in #2001. Currently this will cause problems. Like mentioned earlier these changes solve these issues.

## Types of changes

## Checklist

### DIFF
--- a/adjust.go
+++ b/adjust.go
@@ -1034,7 +1034,7 @@ func (from *xlsxFrom) adjustDrawings(dir adjustDirection, num, offset int, editA
 
 // adjustDrawings updates the ending anchor of the two cell anchor pictures
 // and charts object when inserting or deleting rows or columns.
-func (to *xlsxTo) adjustDrawings(dir adjustDirection, num, offset int, editAs string, ok bool) error {
+func (to *xlsxTo) adjustDrawings(dir adjustDirection, num, offset int, ok bool) error {
 	if dir == columns && to.Col+1 >= num && to.Col+offset >= 0 && ok {
 		if to.Col+offset >= MaxColumns {
 			return ErrColumnNumber
@@ -1054,27 +1054,35 @@ func (to *xlsxTo) adjustDrawings(dir adjustDirection, num, offset int, editAs st
 // inserting or deleting rows or columns.
 func (a *xdrCellAnchor) adjustDrawings(dir adjustDirection, num, offset int) error {
 	editAs := a.EditAs
-	if a.From == nil || a.To == nil || editAs == "absolute" {
+	if (a.From == nil && (a.To == nil || a.Ext == nil)) || editAs == "absolute" {
 		return nil
 	}
 	ok, err := a.From.adjustDrawings(dir, num, offset, editAs)
 	if err != nil {
 		return err
 	}
-	return a.To.adjustDrawings(dir, num, offset, editAs, ok || editAs == "")
+	if a.To != nil {
+		return a.To.adjustDrawings(dir, num, offset, ok || editAs == "")
+	} else {
+		return nil
+	}
 }
 
 // adjustDrawings updates the existing two cell anchor pictures and charts
 // object when inserting or deleting rows or columns.
 func (a *xlsxCellAnchorPos) adjustDrawings(dir adjustDirection, num, offset int, editAs string) error {
-	if a.From == nil || a.To == nil || editAs == "absolute" {
+	if (a.From == nil && (a.To == nil || a.Ext == nil)) || editAs == "absolute" {
 		return nil
 	}
 	ok, err := a.From.adjustDrawings(dir, num, offset, editAs)
 	if err != nil {
 		return err
 	}
-	return a.To.adjustDrawings(dir, num, offset, editAs, ok || editAs == "")
+	if a.To != nil {
+		return a.To.adjustDrawings(dir, num, offset, ok || editAs == "")
+	} else {
+		return nil
+	}
 }
 
 // adjustDrawings updates the pictures and charts object when inserting or
@@ -1124,6 +1132,11 @@ func (f *File) adjustDrawings(ws *xlsxWorksheet, sheet string, dir adjustDirecti
 		return err
 	}
 	for _, anchor := range wsDr.TwoCellAnchor {
+		if err = anchorCb(anchor); err != nil {
+			return err
+		}
+	}
+	for _, anchor := range wsDr.OneCellAnchor {
 		if err = anchorCb(anchor); err != nil {
 			return err
 		}
--- a/adjust.go
+++ b/adjust.go
@@ -38,10 +38,10 @@ var adjustHelperFunc = [9]func(*File, *xlsxWorksheet, string, adjustDirection, i
 		return f.adjustDataValidations(ws, sheet, dir, num, offset, sheetID)
 	},
 	func(f *File, ws *xlsxWorksheet, sheet string, dir adjustDirection, num, offset, sheetID int) error {
-		return f.adjustDefinedNames(ws, sheet, dir, num, offset, sheetID)
+		return f.adjustDefinedNames(sheet, dir, num, offset)
 	},
 	func(f *File, ws *xlsxWorksheet, sheet string, dir adjustDirection, num, offset, sheetID int) error {
-		return f.adjustDrawings(ws, sheet, dir, num, offset, sheetID)
+		return f.adjustDrawings(ws, sheet, dir, num, offset)
 	},
 	func(f *File, ws *xlsxWorksheet, sheet string, dir adjustDirection, num, offset, sheetID int) error {
 		return f.adjustMergeCells(ws, sheet, dir, num, offset, sheetID)
@@ -1063,9 +1063,8 @@ func (a *xdrCellAnchor) adjustDrawings(dir adjustDirection, num, offset int) err
 	}
 	if a.To != nil {
 		return a.To.adjustDrawings(dir, num, offset, ok || editAs == "")
-	} else {
-		return nil
 	}
+	return err
 }
 
 // adjustDrawings updates the existing two cell anchor pictures and charts
@@ -1080,14 +1079,13 @@ func (a *xlsxCellAnchorPos) adjustDrawings(dir adjustDirection, num, offset int,
 	}
 	if a.To != nil {
 		return a.To.adjustDrawings(dir, num, offset, ok || editAs == "")
-	} else {
-		return nil
 	}
+	return err
 }
 
 // adjustDrawings updates the pictures and charts object when inserting or
 // deleting rows or columns.
-func (f *File) adjustDrawings(ws *xlsxWorksheet, sheet string, dir adjustDirection, num, offset, sheetID int) error {
+func (f *File) adjustDrawings(ws *xlsxWorksheet, sheet string, dir adjustDirection, num, offset int) error {
 	if ws.Drawing == nil {
 		return nil
 	}
@@ -1146,7 +1144,7 @@ func (f *File) adjustDrawings(ws *xlsxWorksheet, sheet string, dir adjustDirecti
 
 // adjustDefinedNames updates the cell reference of the defined names when
 // inserting or deleting rows or columns.
-func (f *File) adjustDefinedNames(ws *xlsxWorksheet, sheet string, dir adjustDirection, num, offset, sheetID int) error {
+func (f *File) adjustDefinedNames(sheet string, dir adjustDirection, num, offset int) error {
 	wb, err := f.workbookReader()
 	if err != nil {
 		return err
--- a/col.go
+++ b/col.go
@@ -800,16 +800,15 @@ func (f *File) RemoveCol(sheet, col string) error {
 // pixel. If the width hasn't been set by the user we use the default value.
 // If the column is hidden it has a value of zero.
 func convertColWidthToPixels(width float64) float64 {
-	var padding float64 = 5
 	var pixels float64
-	var maxDigitWidth float64 = 7
+	var maxDigitWidth float64 = 8
 	if width == 0 {
 		return pixels
 	}
 	if width < 1 {
 		pixels = (width * 12) + 0.5
-		return math.Ceil(pixels)
+		return float64(int(pixels))
 	}
-	pixels = (width*maxDigitWidth + 0.5) + padding
-	return math.Ceil(pixels)
+	pixels = (width*maxDigitWidth + 0.5)
+	return float64(int(pixels))
 }
--- a/col.go
+++ b/col.go
@@ -22,10 +22,10 @@ import (
 
 // Define the default cell size and EMU unit of measurement.
 const (
-	defaultColWidth        float64 = 9.140625
-	defaultColWidthPixels  float64 = 64
-	defaultRowHeight       float64 = 15
-	defaultRowHeightPixels float64 = 20
+	defaultColWidth        float64 = 10.5
+	defaultColWidthPixels  float64 = 84.0
+	defaultRowHeight       float64 = 15.6
+	defaultRowHeightPixels float64 = 20.8
 	EMU                    int     = 9525
 )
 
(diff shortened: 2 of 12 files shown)


col.go
```diff
@@ -800,16 +800,15 @@ func (f *File) RemoveCol(sheet, col string) error {
...
 	}
 	if width < 1 {
 		pixels = (width * 12) + 0.5
-		return math.Ceil(pixels)
+		return float64(int(pixels))
 	}
-	pixels = (width*maxDigitWidth + 0.5) + padding
-	return math.Ceil(pixels)
+	pixels = (width*maxDigitWidth + 0.5)
+	return float64(int(pixels))
```
xuri, line 813: After this change, the `math` import no longer required. Unit tests `TestChartSize` and `TestColWidth` should be update.

picture.go
```diff
@@ -753,18 +753,20 @@ func (f *File) drawingResize(sheet, cell string, width, height float64, opts *Gr
...
 		}
 	}
-	if float64(cellWidth) < width {
-		asp := float64(cellWidth) / width
-		width, height = float64(cellWidth), height*asp
-	}
-	if float64(cellHeight) < height {
-		asp := float64(cellHeight) / height
-		height, width = float64(cellHeight), width*asp
+	if float64(cellWidth) < width || float64(cellHeight) < height {
```
xuri, line 756: Did you test this changes in the Windows Excel? It seem the ratio of third picture in issue #2001 still not same with first and second one. But it works on Excel for Mac (Version 16.96.25041326).

xuri (changes requested):

col.go
R3dByt3: @xuri Fixed - thanks for the notice, I just pasted my local changes to the browser, sorry about this one

picture.go
R3dByt3: @xuri Good catch the issue only occurs if a "cell overflow" occurs, since the calculated overflow starting positions are being discarded. The provided offsets are being used instead which causes the offset to multiply rapidly in case of an overflow. Now I have tested locally the no overflow and overflow case using code like:

```
func RenderImage(f *excelize.File, k *KnownTemplateFields, row int, col int, value *ExcelPicture) error {
	cellName, err := excelize.CoordinatesToCellName(col+k.InitialCol, row+k.InitialRow)
	if err != nil {
		log.Println(err)
		return err
	}

	// 60 = 80px * 0.75 (HeightMagic)
	f.SetRowHeight(MainSheetName, row, 60)

	// extension example: .png
	enable := true
	err = f.AddPictureFromBytes(MainSheetName, cellName, &excelize.Picture{
		Extension: value.FileType,
		File:      value.Bytes,
		Format: &excelize.GraphicOptions{
			Positioning:     "oneCell",
			LockAspectRatio: true,
			AutoFit:         true,
			Locked:          &enable,
			PrintObject:     &enable,
			//AutoFitIgnoreAspect: false,
		},
	})
	if err != nil {
		log.Println(err)
		return err
	}

	return nil
}
```

And the resulting text excel looks like this:
grafik

Besides slight (propably) rounding errors the properties show +- 0.01cm the image properties windows shows the same scaling and width

rows.go
```diff
@@ -947,5 +947,5 @@ func convertRowHeightToPixels(height float64) float64 {
 	if height == 0 {
 		return 0
 	}
-	return math.Ceil(4.0 / 3.4 * height)
+	return height * 4.0 / 3.0
```
xuri, line 950: The `math` import of this file also no longer required. Please run test locally by `go test .` after changes.

xuri (changes requested):

col.go
xuri: The unit tests `TestChartSize` and `TestColWidth` test fail after this changes.

xuri (changes requested):
The image will overflow `Sheet1!B30` cell after this changes, the image with `AutoFit` inserted by [this unit test](https://github.com/xuri/excelize/blob/efdfc52252a3f9a840e47da0c8bd4017845eef7a/picture_test.go#L50).
Image position has been affect by the pull request 2123

xuri (changes requested):
The `Sheet2!I9` cell image has been affected (the ratio of width to height increases), the image is widened after this changes. The image inserted by this [unit test](https://github.com/xuri/excelize/blob/efdfc52252a3f9a840e47da0c8bd4017845eef7a/picture_test.go#L39-L40).
changes

picture.go
```diff
@@ -753,18 +753,20 @@ func (f *File) drawingResize(sheet, cell string, width, height float64, opts *Gr
...
+		aspWidth := float64(cellWidth) / width
+		aspHeight := float64(cellHeight) / height
+		var asp float64
```
xuri, line 759: This can be simplify with:

```go
aspWidth := float64(cellWidth) / width
aspHeight := float64(cellHeight) / height
asp := min(aspWidth, aspHeight)
width, height = width*asp, height*asp
```

xuri (changes requested):

col.go
R3dByt3: Sorry somehow I missed that there are affected tests - will be fixed

R3dByt3:
> The image will overflow `Sheet1!B30` cell after this changes, the image with `AutoFit` inserted by [this unit test](https://github.com/xuri/excelize/blob/efdfc52252a3f9a840e47da0c8bd4017845eef7a/picture_test.go#L50). Image position has been affect by the pull request 2123

I fear that this is a very complex subject which needs to be discussed, please hear me out.

First of all I just realized another issue of the current implementation and fork (at least from my point of view). If I understand you correctly you suggest, that AutoFit should add pictures into the target cell and resize the image in a way, that it is not rendered beyond the cell borders.

However the current implementation changes the from anchor if the offset leaves the area of the cell. This means that a cell insertion targeting A1 can end up starting in A2. While the initial visible result seems to be correct there is an issue: The image will be moved if Cell A1 is being enlarged. This is not correct and I think that I should fix this as well.

The starting anchor point can be outside of the cell dimensions which leads to correct anchoring and simply hides the image as long as the cell is not expanded enough. In the end this will still break your expectation, but just in another way.

AutoFit would resize the image respecting the aspect ratio to the current cell dimensions and if an offset is used which causes the image to overflow the cell, it would look cut off, but this could be changed by the user, by expading the row / col size.

However, from my understanding your expected behaviour comes along with various problematic edge cases and this is why I would like to suggest considering to review this feature:

1. If the same image is added twice with the same settings except the offset, it might be possible, that both images end up with different sizes. Should varying offsets really affect the image size?
2. Consider the case of adding multiple images or dynamically calculating cell sizes and images positions - Would it really be the expected behaviour, if there are only a few pixels left from the start anchor, that an image is added with for example 5x5, 1x1 pixels?
3. When adding multiple images dynamically, maybe even with varying native sizes, this may be not the desired behaviour like in my case and can only be fixed by calculating the required col / row size for all rows / cols in advance.

What are your thoughts about this?

picture.go
R3dByt3: Will be changed, thanks for the info

xuri:
Any update for other remaining code review issues?

R3dByt3:
Im working on it hoping to complete it today, from what I have noticed the anchoring wasnt fully correct and confused me a bit.

R3dByt3:
@xuri I think its finally done

Some explanations:

- oneCellAnchors (<=> InsertionMode Picture shall move with cell) were not fully implemented, which could lead to images not beeing shown - fixed

- oneCellAnchors have not been adjusted on col / row insertion - fixed

- insertionPositionCalculation does not modify anymore the targeted starting cell => from my point of view the targeted starting cell must not be modified, since the resulting behaviour will otherwise not be equal to the requested one:

grafik
(Images get seperated by cell overflows)

- updated some excel default values and unit conversion formulas => at least for me (a german office 365 user) these numbers seem to work a little better, but the best case would be to include some font based calculations

- row and col functions to retrieve the height / width did seemed to handle the hidden row / col case incorrectly - fixed

Please notice regarding the anchoring behaviour:
Using a twoCellAnchor (default) the starting cell and its size are the most top left point of the area of insertion; Using a twoCellAnchor with offsets **will** cause headaches, since the image will start scaling based on its own size and cell sizes etc. Also it is not possible to have an offset > the area of the starting cell.

xuri (changes requested):
After these changes, add 60px x 60px image to test issue #2001, the size of the last 3 images is incorrect.

You can download this image attachment: 



The result:

adjust.go
```diff
@@ -1054,27 +1054,35 @@ func (to *xlsxTo) adjustDrawings(dir adjustDirection, num, offset int, editAs st
...
-	if a.From == nil || a.To == nil || editAs == "absolute" {
+	if (a.From == nil && (a.To == nil || a.Ext == nil)) || editAs == "absolute" {
 		return nil
 	}
 	ok, err := a.From.adjustDrawings(dir, num, offset, editAs)
 	if err != nil {
 		return err
 	}
-	return a.To.adjustDrawings(dir, num, offset, editAs, ok || editAs == "")
+	if a.To != nil {
```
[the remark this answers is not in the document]
R3dByt3: Will do this 👍

adjust.go
```diff
@@ -1128,6 +1136,11 @@ func (f *File) adjustDrawings(ws *xlsxWorksheet, sheet string, dir adjustDirecti
 			return err
 		}
 	}
+	for _, anchor := range wsDr.OneCellAnchor {
+		if err = anchorCb(anchor); err != nil {
+			return err
```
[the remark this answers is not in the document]
R3dByt3: There is at least already an integration test for this, which causes a failure if it is missing, adjust_test.go l. 1209 and following would fail now without this, since the col / row insertions would not occur for oneCellAnchors - if this is sufficient for you.

R3dByt3:
> After these changes, add 60px x 60px image to test issue #2001, the size of the last 3 images is incorrect.
> 
> You can download this image attachment:
> 
> 
> 
> The result:
> 

I will try to reproduce this issue and add a test case for it 👍 (Propably on monday)

adjust.go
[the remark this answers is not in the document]
R3dByt3: Fixed

adjust.go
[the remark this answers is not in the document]
R3dByt3: @xuri Is the already existing test which showed me this missing piece of code sufficient for you?

R3dByt3:
I fixed the absolute anchoring issue - the issue arised from updating the anchors since oneCellAnchors have absolute sizes.

xuri (changes requested):
Thanks for your update. I test issue #2001, all the size of the images is correct, but the position of the last image is incorrect. If the value of `OffsetX` and `OffsetY` is over than anchor cell column width or height, it won't work.

image

Expected:

image

R3dByt3:
Hopefully this covers all your questions or issues: If we are talking about oneCellAnchors, Excel starts overlapping images from last to first if the images do not fit into the column. This can be observed by resizing the column. Implementing an overflow by hand creates other issues like I have mentioned earlier (image separation; incorrect anchor cell, which causes unexpected movements on column resizing/hiding). At least currently I do not know about a property or similar to control or override this behavior of excel regarding too many images for column size X.

Does this answer your question or did I miss something?

adjust.go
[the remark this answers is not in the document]
xuri: I pushed a commit on your branch, added unit test for coverage this line.

xuri (approved):
The `positionObjectPixels` function has too many parameters, I pushed a commit on your branch, combine some parameters to made it shorten and simplicity.

xuri (commented):

xuri:
Hi @R3dByt3, this changes will cause generated workbook corrupted on Office 2007. The [`EditAs`](https://github.com/qax-os/excelize/blob/d230ecdb48b77450baa5fe664f5681ef3ad38360/picture.go#L375) should be empty if `Positioning` is `oneCell`. We need fix it like this:

```diff
cellAnchor := xdrCellAnchor{}
-cellAnchor.EditAs = opts.Positioning
from := xlsxFrom{}
from.Col = colStart
from.ColOff = x1 * EMU
from.Row = rowStart
from.RowOff = y1 * EMU
cellAnchor.From = &from

if opts.Positioning != "oneCell" {
    to := xlsxTo{}
    to.Col = colEnd
    to.ColOff = x2 * EMU
    to.Row = rowEnd
    to.RowOff = y2 * EMU
    cellAnchor.To = &to
+   cellAnchor.EditAs = opts.Positioning
}
```

Now-Shimmer:
Hi @xuri, @R3dByt3, I created a PR #2128 to fix this. Please take a review for that, thanks!

xuri:
I found new issue about PR #2123. If open a workbook which contain shape with one cell anchor, and which has following element:

```xml
<xdr:ext cx="4495333" cy="937629"/>
<xdr:sp macro="" textlink="">
```

After this changes, open the workbook and insert a column, the generated workbook corrupted, some attribute of one cell anchor shape is missing:

```xml
<xdr:ext></xdr:ext>
<xdr:sp>
```

The next version 2.9.1 will release on May 13th, 2025. I will revert this PR #2123 if we not resolve this issue.

xuri:
Issue 3: if adding drawing objects such as pictures, charts, shapes, and form controls with the `Positioning` value as `absolute`, the drawing object size will fit the cell. Just like enable `LockAspectRatio`.

R3dByt3:
I will take a look at these issues

crush-wu:
In this PR:

- Change default column width from 9.140625 to 10.5
- Change default column width in pixels from 64 to 84
- Change default row height from 15 to 15.6
- Change default row height in pixels from 20 to 20.8
- Change column width and row height to pixels conversion calculation logic

That's a really a huge breaking change, all drawing objects such as pictures, charts, shapes, and form controls size which depends on width and height would be impacted. Users need to adjust all sizes for their code. Can we keep the origin default value?

R3dByt3:
@crush-wu if I remember correctly, during implementation and testing these changes caused the images of AddPictureTests.xlsx to keep their aspect ratios, if a column has not been modified / has default dimensions. Partially as mentioned earlier during this PR some of the units used by excel would actually require calculations based on font and font size, which makes any hard coded values incorrect. On top of that I can not test this for various excel versions.

If possible I will try to roll back this change too, if not - im unsure. Best case would be another PR with font calculations or hardcoded factors?

Now-Shimmer:
> I found new issue about PR #2123. If open a workbook which contain shape with one cell anchor, and which has following element:
> 
> ```
> <xdr:ext cx="4495333" cy="937629"/>
> <xdr:sp macro="" textlink="">
> ```
> 
> After this changes, open the workbook and insert a column, the generated workbook corrupted, some attribute of one cell anchor shape is missing:
> 
> ```
> <xdr:ext></xdr:ext>
> <xdr:sp>
> ```
> 
> The next version 2.9.1 will release on May 13th, 2025. I will revert this PR #2123 if we not resolve this issue.

Hi @xuri, @R3dByt3, I fixed this issue on this commit [`1b9e19b`](https://github.com/xuri/excelize/pull/2128/commits/1b9e19b74a7f16d0f2fcd8ebca4ea5af1e12215c)

Now-Shimmer:
The 

> Issue 3: if adding drawing objects such as pictures, charts, shapes, and form controls with the `Positioning` value as `absolute`, the drawing object size will fit the cell. Just like enable `LockAspectRatio`.

Hi @R3dByt3, @xuri, this issue fixed by this commit [`cbcf7f9`](https://github.com/xuri/excelize/pull/2128/commits/cbcf7f95711c6cc5e485b5e1ff0b29f15f7c5601).

R3dByt3:
> > Issue 3: if adding drawing objects such as pictures, charts, shapes, and form controls with the `Positioning` value as `absolute`, the drawing object size will fit the cell. Just like enable `LockAspectRatio`.
> 
> Hi @R3dByt3, @xuri, this issue fixed by this commit [`cbcf7f9`](https://github.com/xuri/excelize/pull/2128/commits/cbcf7f95711c6cc5e485b5e1ff0b29f15f7c5601). All issues has been fixed, any suggestion about this?

LGTM - thank you very much for fixing these issues so fast

R3dByt3:
I guess that might have been a pitfall I have fallen into... Good job @Now-Shimmer and good research.

@Now-Shimmer @xuri What do you think about making these values configurable / overrideable on opening a file?

R3dByt3:
@Now-Shimmer I could not help but to wonder a bit more so I researched further:

Can you verify this with ur tests?
https://answers.microsoft.com/en-us/msoffice/forum/all/excel-new-blank-book-default-row-height-problem/9df7c580-9020-4b51-ae66-34e86a01dae3

Maybe it would be possible to fix this by a precomputed table, since even this topic seems to depend on font settings.
https://github.com/closedxml/closedxml/wiki/Cell-Dimensions#skiasharp-code

xuri:
Good job! I will take look for the PR #2128. Yeah, I agree with made it configurable by add related fields in the [`Options`](https://github.com/xuri/excelize/blob/d230ecdb48b77450baa5fe664f5681ef3ad38360/excelize.go#L67-L111) data type in future version.

Issue #2126, opened 2025-05-07
DefinedNames SheetNames are not escaped correctly, if a sheet gets renamed


R3dByt3:
### Description

Renaming a sheet might cause Microsoft Excel to show a warning that the file contains broken links. This happens if the sheet name contains whitespaces for example and definedNames referencing this sheet. This is also the reason why some sheet names are escaped and others not. This PR checks via regex for non word characters and escapes sheet names in definedNames if required. The current implementation only preserves already present escaping, even if it is not required anymore.
Currently "Sheet 2!$A$1:$A$2"
Required "'Sheet 2'!$A$1:$A$2"

### Steps to reproduce the issue

DefinedName = "Sheet1!$A$1:$A$2"
```
	if err := f.SetSheetName("Sheet1", "Requires escaping"); err != nil {
		log.Println(err)
		return err
	}
```

### Describe the results you received

DefinedName = "Requires escaping!$A$1:$A$2"

### Describe the results you expected

DefinedName = "'Requires escaping'!$A$1:$A$2" 

### Go version

1.24.2

### Excelize version or commit ID

2.9.0 / master

### Environment

```shell
Alpine, Microsoft Excel 365
```

### Validations

- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a minimal reproducible example of the bug.

PR #2127, opened 2025-05-07
DefinedNames SheetNames are not escaped correctly, if a sheet gets renamed

R3dByt3:
# PR Details

DefinedNames SheetNames are not escaped correctly, if a sheet gets renamed

## Description

Renaming a sheet might cause Microsoft Excel to show a warning that the file contains broken links. This happens if the sheet name contains whitespaces for example and definedNames referencing this sheet. This is also the reason why some sheet names are escaped and others not. This PR checks via regex for non word characters and escapes sheet names in definedNames if required. The current implementation only preserves already present escaping, even if it is not required anymore.
Currently "Sheet 2!$A$1:$A$2"
Required "'Sheet 2'!$A$1:$A$2"

## Related Issue

#2126 

## Motivation and Context

Fixes "broken links".

## How Has This Been Tested

Tested in a local project.

## Types of changes

## Checklist

### DIFF
--- a/adjust.go
+++ b/adjust.go
@@ -15,6 +15,7 @@ import (
 	"bytes"
 	"encoding/xml"
 	"io"
+	"regexp"
 	"strconv"
 	"strings"
 	"unicode"
@@ -38,10 +39,10 @@ var adjustHelperFunc = [9]func(*File, *xlsxWorksheet, string, adjustDirection, i
 		return f.adjustDataValidations(ws, sheet, dir, num, offset, sheetID)
 	},
 	func(f *File, ws *xlsxWorksheet, sheet string, dir adjustDirection, num, offset, sheetID int) error {
-		return f.adjustDefinedNames(ws, sheet, dir, num, offset, sheetID)
+		return f.adjustDefinedNames(sheet, dir, num, offset)
 	},
 	func(f *File, ws *xlsxWorksheet, sheet string, dir adjustDirection, num, offset, sheetID int) error {
-		return f.adjustDrawings(ws, sheet, dir, num, offset, sheetID)
+		return f.adjustDrawings(ws, sheet, dir, num, offset)
 	},
 	func(f *File, ws *xlsxWorksheet, sheet string, dir adjustDirection, num, offset, sheetID int) error {
 		return f.adjustMergeCells(ws, sheet, dir, num, offset, sheetID)
@@ -1014,7 +1015,7 @@ func (from *xlsxFrom) adjustDrawings(dir adjustDirection, num, offset int, editA
 
 // adjustDrawings updates the ending anchor of the two cell anchor pictures
 // and charts object when inserting or deleting rows or columns.
-func (to *xlsxTo) adjustDrawings(dir adjustDirection, num, offset int, editAs string, ok bool) error {
+func (to *xlsxTo) adjustDrawings(dir adjustDirection, num, offset int, ok bool) error {
 	if dir == columns && to.Col+1 >= num && to.Col+offset >= 0 && ok {
 		if to.Col+offset >= MaxColumns {
 			return ErrColumnNumber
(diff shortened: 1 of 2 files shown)


adjust.go
```diff
@@ -467,9 +468,10 @@ func adjustRangeSheetName(rng, source, target string) string {
 				if singleQuote {
 					part = strings.TrimPrefix(strings.TrimSuffix(part, "'"), "'")
 				}
+				needsQuoting := regexp.MustCompile(`\W`).MatchString(target)
 				if part == source {
-					if part = target; singleQuote {
-						part = "'" + part + "'"
+					if part = target; needsQuoting {
```
xuri, line 473: Using `string.Contains` instead of regexp for better performance. Just simplify like this:

```diff
for k, part := range parts {
    singleQuote := strings.HasPrefix(part, "'") && strings.HasSuffix(part, "'")
    if singleQuote {
        part = strings.TrimPrefix(strings.TrimSuffix(part, "'"), "'")
    }
    if part == source {
-       if part = target; singleQuote {
+       if part = target; singleQuote || strings.Contains(target, " ") {
            part = "'" + part + "'"
        }
    }
    parts[k] = part
}
```

xuri (changes requested):
Thanks for your PR. I've left a comment.

adjust.go
R3dByt3: Thank you very much for your feedback, but I think this regex is required. As written in the description _non word characters_ probably all have to be escaped, especially characters excel syntactically uses:

'!' '$' ':' '-' '_' ...

So I thought using a regex instead of checking for each special character individually would be the preferred solution.

xuri: Yep, I forget we already have [`escapeSheetName`](https://github.com/xuri/excelize/blob/8fdc26e286734b808dcab1546760768b60045e67/adjust.go#L308-L317) function, use it would be better approach:

```diff
for k, part := range parts {
    singleQuote := strings.HasPrefix(part, "'") && strings.HasSuffix(part, "'")
    if singleQuote {
        part = strings.TrimPrefix(strings.TrimSuffix(part, "'"), "'")
    }
    if part == source {
-       if part = target; singleQuote {
+       if part = escapeSheetName(target); singleQuote {
            part = "'" + part + "'"
        }
    }
    parts[k] = part
}
```

R3dByt3: Fixed - sorry my git somehow messed up during refactoring

xuri (changes requested):
Hi @R3dByt3, I found new issue on update defined name with multiple worksheets. Please consider this case:

```go
package main

import (
    "fmt"

    "github.com/xuri/excelize/v2"
)

func main() {
    f := excelize.NewFile()
    defer func() {
            fmt.Println(err)
        }
    }()
    if _, err := f.NewSheet("Sheet 3"); err != nil {
        fmt.Println(err)
        return
    }
    if err := f.SetDefinedName(&excelize.DefinedName{
        Name:     "Amount3",
        RefersTo: "'Sheet 3'!$A$2:$D$5",
        Comment:  "defined name comment",
        Scope:    "Sheet3",
    }); err != nil {
        fmt.Println(err)
        return
    }
    if err := f.SetDefinedName(&excelize.DefinedName{
        Name:     "Amount",
        RefersTo: "Sheet1!$A$2:$D$5",
        Comment:  "defined name comment",
        Scope:    "Sheet2",
    }); err != nil {
        fmt.Println(err)
        return
    }
    if err := f.SetSheetName("Sheet1", "Requires escaping"); err != nil {
        fmt.Println(err)
        return
    }
        fmt.Println(err)
    }
}
```

The defined name "Amount3" shouldn't be changed.

- Expected: `'Sheet 3'!$A$2:$D$5`
- Actual:   `Sheet '3'!$A$2:$D$5`

The fix could be like this:

```go
func adjustRangeSheetName(rng, source, target string) string {
    source = escapeSheetName(source)
    cellRefs := strings.Split(rng, ",")
    for i, cellRef := range cellRefs {
        rangeRefs := strings.Split(cellRef, ":")
        for j, rangeRef := range rangeRefs {
            parts := strings.Split(rangeRef, "!")
            for k, part := range parts {
                if strings.TrimPrefix(strings.TrimSuffix(part, "'"), "'") == source {
                    part = escapeSheetName(target)
                }
                parts[k] = part
            }
            rangeRefs[j] = strings.Join(parts, "!")
        }
        cellRefs[i] = strings.Join(rangeRefs, ":")
    }
    return strings.Join(cellRefs, ",")
}
```

xuri (approved):
Great work! Thanks for your contribution.

xuri (approved):

Issue #1961, opened 2024-07-22
Is there any possible GetCellFormula faster?


ShowerBandV:
**Description**
A sheet which has 4600 rows and 55 cols, and some cols have formulas,I want to get them,it seems no problem but the time cost is very large,I want to know whether excelize can solve this.

here is demo code and file
```golang
func TestGetFormulas(t *testing.T) {
	file, err := excelize.OpenFile("4600Formulas.xlsx")
	if err != nil {
		panic(err)
	}
	begin := time.Now()
	for col := 0; col < 55; col++ {
		for row := 0; row < 4600; row++ {
			axis := AxisEncode(col, row)
			_, err := file.GetCellFormula("Sheet1", axis)
			if err != nil {
				panic(err)
			}
		}
	}
	fmt.Printf("GetCellFormula time cost:%dms", time.Since(begin).Milliseconds())
}
```
[4600Formulas.xlsx](https://github.com/user-attachments/files/16330178/4600Formulas.xlsx)

xuri:
Thanks for your issue. Currently, there are some formula functions still working in progress, I think we can optimize performance for formula calculation engine after formula functions implemented.

HZL3151904214:
共享函数的实现有问题，理论上共享函数应该单独存储一个共享函数的索引结构吧，不应该为了获取共享函数的表达式在多遍历一次m*n，这样一旦是共享函数就需要m*n*m*n，性能太差了。作者这块有规划优化下？

xuri:
Thanks for the feedback. If you would like to create a pull request for this, that would be great.

xuri:
This issue has been fixed by PR #2118. Please upgrade to the master branch code by `go get github.com/xuri/excelize/v2@master`, and this feature will be released on the next version.

xuri:
This issue has been fixed by PR #2118. Please upgrade to the master branch code by `go get github.com/xuri/excelize/v2@master`, and this patch will be released on the next version.

PR #2129, opened 2025-05-11
Using binary search instead of sequential search for rowData

shcabin:
# PR Details

Using binary search instead of sequential search for rowData

## Related Issue
#1961 

## How Has This Been Tested

Performance testing was conducted on an AMD64 WSL using Go's benchmark tool. The benchmarks were run the test mentioned in #1961 8 times . The following results were observed:

```
goos: linux
goarch: amd64
pkg: github.com/xuri/excelize/v2
cpu: 12th Gen Intel(R) Core(TM) i5-12400F
               │ base.txt   │             new.txt               │
               │   sec/op   │   sec/op    vs base               │
GetFormulas-12   3.951 ± 4%   1.382 ± 3%  -65.02% (p=0.000 n=8)

               │  base.txt    │           new.txt             │
               │     B/op     │     B/op      vs base         │
GetFormulas-12   815.1Mi ± 0%   815.1Mi ± 0%  ~ (p=0.721 n=8)

               │  base.txt   │           new.txt            │
               │  allocs/op  │  allocs/op   vs base         │
GetFormulas-12   10.87M ± 0%   10.87M ± 0%  ~ (p=0.940 n=8)
(The version of the issue submitted is 27 seconds in the same environment)
```
code:
```

func BenchmarkGetFormulas(b *testing.B) {
	for i := 0; i < b.N; i++ {
		file, err := OpenFile("4600Formulas.xlsx")
		if err != nil {
			panic(err)
		}
		begin := time.Now()
		cc := 0
		for row := 1; row <= 4600; row++ { // 4600
			for col := 1; col <= 55; col++ { //55
				axis, _ := CoordinatesToCellName(col, row)
				str, err := file.GetCellFormula("Sheet1", axis)
				if err != nil {
					panic(err)
				}
				if str != "" {
					cc++
				}
			}
		}
		assert.Equal(b, 45990, cc)
		b.Logf("GetCellFormula time cost:%dms cc:%v", time.Since(begin).Milliseconds(), cc)
	}
}
```

## Types of changes

## Checklist

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -1502,11 +1502,11 @@ func (f *File) getCellStringFunc(sheet, cell string, fn func(x *xlsxWorksheet, c
 	idx, found := sort.Find(len(ws.SheetData.Row), func(i int) int {
 		if ws.SheetData.Row[i].R == row {
 			return 0
-		} else if ws.SheetData.Row[i].R > row {
-			return -1
-		} else {
-			return 1
 		}
+		if ws.SheetData.Row[i].R > row {
+			return -1
+		}
+		return 1
 	})
 	if !found {
 		return "", nil
--- a/cell.go
+++ b/cell.go
@@ -17,6 +17,7 @@ import (
 	"math"
 	"os"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


shcabin:
this PR dependency that 'SheetData.Row[i].R ' is ordered.  
the first commit does not have this dependency and can reduce the time by 55% compared to the previous version (3.95seconds vs 1.762seconds).

cell.go
```diff
@@ -1498,23 +1499,30 @@ func (f *File) getCellStringFunc(sheet, cell string, fn func(x *xlsxWorksheet, c
...
+			return 1
 		}
-		for colIdx := range rowData.C {
-			colData := &rowData.C[colIdx]
-			if cell != colData.R {
-				continue
-			}
+	})
+	if !found {
+		return "", nil
```
xuri, line 1512: Need add unit test for coverage this case.

xuri (changes requested):

xuri:
> this PR dependency that 'SheetData.Row[i].R ' is ordered. the first commit does not have this dependency and can reduce the time by 55% compared to the previous version (3.95seconds vs 1.762seconds).

This is awesome! Does these changes increase memory usage? If so, how much memories will be increased?

cell.go
shcabin: I have tried, but I don't seem to have found a suitable way to construct this test case(how to construct discontinuous rows.).

shcabin:
> This is awesome! Does these changes increase memory usage? If so, how much memories will be increased?

 no change in memory usage

cell.go
xuri: I add test case for this based on your branch.

xuri (approved):

PR #2141, opened 2025-05-23
feat(font): add charset support to font definitions

aliavd1:
Excel requires the charset attribute to render certain non-ASCII fonts correctly, especially for East Asian languages. This commit introduces support for setting the `charset` field when defining fonts.

Changes:
- Added `Charset` field to Font struct
- Modified XML marshaling logic to include `charset` when present
- Added tests covering different charset values

This allows developers to explicitly specify font encodings when generating spreadsheets.


xuri (approved):
LGTM, thanks for your contribution. I added get charset support based on your branch.

PR #2144, opened 2025-05-25
Add calculation cache to improve CalcCellValue performance

### LINKED ISSUE #2057: Performance enhancement - cache calculated values
**Description**

My use case is as follows.  I open an Excel file.  I inject around 10 values into various cells (using SetCellStr) and then extract around 100 values from formulas on the sheet (using CalcCellValue) and then exit.

It seems that Excelize doesn't cache any values it has calculated via CalcCellValue, and recalculates them over and over if they are used in other cells.  If these formulae are time consuming this is very wasteful.

Example : 

A1:  40
A2:  =VERYEXPENSIVEFUNCTION( )
A3:  =A1+A2

CalcCellValue("Sheet1", "A1")  // takes 0ms
CalcCellValue("Sheet1", "A2")  // takes 100ms (for example)
CalcCellValue("Sheet1", "A3")  // takes 100ms (for example)

CalcCellValue("Sheet1", "A3") could be practically instant, because Excelize could/should already know the value of A1 and A2 from its previous calculations.  It also knows that no SetCellX( ) functions have been called and therefore the values A1 and A2 can't have changed.  So it shouldn't have to recalculate A2 when calculating A3.

As a first pass Excelize could cache the results of CalcCellValue and invalidate whole cache when any SetCellX( ) is called.

A better version could only invalidate the parts of the cache that are affected by a SetCellX( ) by building a formula dependency tree (as previously suggested [here](https://github.com/qax-os/excelize/issues/757))

**Output of `go version`:**

```text
go version go1.23.1 windows/amd64
```

```text
2.9.0
```

Windows

DengY11:
## PR Details
Add calculation cache to improve CalcCellValue performance

## Description
Implements a calculation cache mechanism for `CalcCellValue` to significantly improve performance when calculating formulas repeatedly. The cache stores calculated results and automatically invalidates when cells are modified.

## Related Issue
Fixes #2057 - Performance enhancement - cache calculated values

## Motivation and Context
As described in #2057, Excelize doesn't cache calculated values from `CalcCellValue`, causing expensive formulas to be recalculated every time they are referenced by other cells. This leads to poor performance when dealing with complex formula dependencies.

**Example scenario:**
```
A1: 40
A2: =VERYEXPENSIVEFUNCTION()  // takes 100ms
A3: =A1+A2                    // takes 100ms (recalculates A2)
```

**Solution implemented:**
- Cache calculated results in thread-safe `sync.Map`
- Clear entire cache when any `SetCellX()` function is called
- Cache key format: `"sheet!cell"`

**Performance improvement:**
- First calculation: ~150µs (calculated and cached)
- Second calculation: ~600ns (from cache, **250x faster**)

## How Has This Been Tested
```bash
go test -v -run "TestCalc.*Cache|TestSetFunctionsClearCache"
go test ./...
```

**Test coverage includes:**
- ☑️ Basic cache hit/miss scenarios
- ☑️ Cache invalidation on cell modifications
- ☑️ Multi-cell formula dependencies
- ☑️ All 12 Set functions properly clear cache
- ☑️ Thread safety 
- ☑️ 3 new comprehensive test functions added

## Types of changes
☑️ New feature (non-breaking change which adds functionality)
☐ Docs change / refactoring / dependency upgrade
☐ Bug fix (non-breaking change which fixes an issue)
☐ Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
☑️ My code follows the code style of this project.
☐ My change requires a change to the documentation.
☐ I have updated the documentation accordingly.
☑️ I have read the CONTRIBUTING document.
☑️ I have added tests to cover my changes.
☑️ All new and existing tests passed.

## Implementation Details

**Modified Files:**
- `excelize.go`: Added cache fields (`calcCache sync.Map`, `calcCacheMu sync.RWMutex`) to File struct
- `calc.go`: Implemented caching logic in `CalcCellValue()` + added `clearCalcCache()` helper
- `cell.go`: Added cache clearing to all 12 Set functions
- `calc_test.go`: Added 3 comprehensive cache test functions

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -634,7 +634,6 @@ func (f *File) GetRowOutlineLevel(sheet string, row int) (uint8, error) {
 // worksheet, it will cause a file error when you open it. The excelize only
 // partially updates these references currently.
 func (f *File) RemoveRow(sheet string, row int) error {
-	f.clearCalcCache()
 	if row < 1 {
 		return newInvalidRowNumberError(row)
 	}
--- a/merge.go
+++ b/merge.go
@@ -66,7 +66,6 @@ func (f *File) MergeCell(sheet, topLeftCell, bottomRightCell string) error {
 	}
 	ws.mu.Lock()
 	defer ws.mu.Unlock()
-	f.clearCalcCache()
 	for col := rect[0]; col <= rect[2]; col++ {
 		for row := rect[1]; row <= rect[3]; row++ {
 			if col == rect[0] && row == rect[1] {


xuri (changes requested):
Code conflicts need resovle.

DengY11:
> Code conflicts need resolve.

so sorry, the conflict was caused by my local auto-formatting. I've resolved it by keeping the original content from master

ar090:
This is really exciting! I have some very large workbooks this should help out on a ton.

ar090:
Hi @DengY11 and @xuri thanks so much for working on this! What is left to get this merged and released?

paolobarbolini:
Ping @xuri

xuri:
Will take review for this.

sheet.go
```diff
@@ -570,6 +573,7 @@ func (f *File) setSheetBackground(sheet, extension string, file []byte) error {
 // value of the deleted worksheet, it will cause a file error when you open
 // it. This function will be invalid when only one worksheet is left.
 func (f *File) DeleteSheet(sheet string) error {
+	f.clearCalcCache()
```
xuri: If the given sheet name was invalid, we needn't clear calc cache, I suggest move this statement after `f.deleteCalcChain`.

sheet.go
```diff
@@ -624,6 +628,7 @@ func (f *File) DeleteSheet(sheet string) error {
 //
 //	err := f.MoveSheet("Sheet2", "Sheet1")
 func (f *File) MoveSheet(source, target string) error {
+	f.clearCalcCache()
 	if strings.EqualFold(source, target) {
```
xuri, line 631: Does clear calc cache on move sheet order is required?

sheet.go
```diff
@@ -371,6 +373,7 @@ func (f *File) getActiveSheetID() int {
 // sheet name in the formula or reference associated with the cell. So there
 // may be problem formula error or reference missing.
 func (f *File) SetSheetName(source, target string) error {
+	f.clearCalcCache()
```
xuri: Clear calc cache after sheet name checking.

xuri (changes requested):
Sorry for late reply, I've left some comments.

rows.go
```diff
@@ -700,6 +702,7 @@ func (f *File) InsertRows(sheet string, row, n int) error {
 // worksheet, it will cause a file error when you open it. The excelize only
 // partially updates these references currently.
 func (f *File) DuplicateRow(sheet string, row int) error {
```
xuri, line 702: For the `DuplicateRow`, `DuplicateRowTo`, `InsertCol` and `InsertRow` functions, we can clear calc cache in `f.adjustHelper` function instead of call `f.clearCalcCache` on each functions.

xuri (changes requested):

cell.go
```diff
@@ -1043,6 +1052,7 @@ func (f *File) removeHyperLink(ws *xlsxWorksheet, sheet, cell string) error {
 //
 //	err := f.SetCellHyperLink("Sheet1", "A3", "Sheet1!A40", "Location")
 func (f *File) SetCellHyperLink(sheet, cell, link, linkType string, opts ...HyperlinkOpts) error {
+	f.clearCalcCache()
```
xuri: Need to clear calc cache on set cell hyperlink.

cell.go
```diff
@@ -1389,6 +1400,7 @@ func (f *File) SetCellRichText(sheet, cell string, runs []RichTextRun) error {
 //
 //	err := f.SetSheetRow("Sheet1", "B6", &[]interface{}{"1", nil, 2})
 func (f *File) SetSheetRow(sheet, cell string, slice interface{}) error {
+	f.clearCalcCache()
```
xuri: The `SetSheetCol` and `SetSheetRow` will call `SetCellValue`, so we needn't clear calc cache at here.

col.go
```diff
@@ -748,6 +748,7 @@ func (f *File) GetColWidth(sheet, col string) (float64, error) {
 // worksheet, it will cause a file error when you open it. The excelize only
 // partially updates these references currently.
 func (f *File) InsertCols(sheet, col string, n int) error {
+	f.clearCalcCache()
```
xuri: Clear calc cache in `f.adjustHelper` function would be better approach.

col.go
```diff
@@ -768,6 +769,7 @@ func (f *File) InsertCols(sheet, col string, n int) error {
 // worksheet, it will cause a file error when you open it. The excelize only
 // partially updates these references currently.
 func (f *File) RemoveCol(sheet, col string) error {
+	f.clearCalcCache()
```
xuri: Same as above, clear calc cache in f.adjustHelper function.

rows.go
```diff
@@ -679,6 +680,7 @@ func (f *File) RemoveRow(sheet string, row int) error {
 // worksheet, it will cause a file error when you open it. The excelize only
 // partially updates these references currently.
 func (f *File) InsertRows(sheet string, row, n int) error {
+	f.clearCalcCache()
```
xuri: Clear calc cache in `f.adjustHelper`.

sheet.go
```diff
@@ -751,6 +756,7 @@ func (f *File) getSheetRelationshipsTargetByID(sheet, rID string) string {
 //	}
 //	err = f.CopySheet(0, index)
 func (f *File) CopySheet(from, to int) error {
+	f.clearCalcCache()
```
xuri: Clear calc in the `f.copySheet` function.

xuri (changes requested):
Due to formula function may reference defined name or table name, so we need also clear calc cache in `SetDefinedName`, `DeleteDefinedName`, `AddPivotTable`, `DeletePivotTable`, `AddTable` and `DeleteTable` functions.

calc.go
```diff
@@ -18984,3 +18865,12 @@ func (fn *formulaFuncs) DISPIMG(argsList *list.List) formulaArg {
 	}
 	return argsList.Front().Value.(formulaArg)
 }
+
+func (f *File) clearCalcCache() {
+	f.calcCacheMu.Lock()
+	defer f.calcCacheMu.Unlock()
+	f.calcCache.Range(func(key, value interface{}) bool {
```
xuri: Can we using `f.calcCache.Clear()` instead of range each key?

xuri:
Notice that, we needn't add `calcCacheMu` in `sync.RWMutex` data type around `calcCache`, because `calcCache` is a `sync.Map`, which already concurrency-safe by design. We need to create a separate PR to fix that.

PR #2152, opened 2025-06-09
streaming writer set default value like non-streaming one

mengpromax:
# PR Details
1. make receiver name of xlsxC unified.
2. streaming writer can set cell default value like non-streaming one.

## Related Issue
https://github.com/qax-os/excelize/issues/2151

## Motivation and Context
We have a legacy impl using the non-streamingly writer to write excel, in which we use setCellDefault func to set raw value.
Now i want to migrate some logic to streaming writer impl, i found no a corresponding impl setCellDefault for steaming write. :)

## How Has This Been Tested
- UT test for pass SetCellDefaultValue to setCellValFunc.
- Test in our business scenarios.

## Types of changes

## Checklist

### DIFF
--- a/stream_test.go
+++ b/stream_test.go
@@ -78,8 +78,14 @@ func TestStreamWriter(t *testing.T) {
 	assert.Equal(t, ErrMaxRowHeight, streamWriter.SetRow("A8", nil, RowOpts{Height: MaxRowHeight + 1}))
 
 	assert.NoError(t, streamWriter.SetRow("A9", []interface{}{math.NaN(), math.Inf(0), math.Inf(-1)}))
+	assert.NoError(t, streamWriter.SetRow("A10", []interface{}{
+		SetCellDefaultValue("1.0"), "1.0", 1.0,
+	}))
+	assert.NoError(t, streamWriter.SetRow("A11", []interface{}{
+		SetCellDefaultValue("2.0"), "2.0", 2.0,
+	}))
 
-	for rowID := 10; rowID <= 51200; rowID++ {
+	for rowID := 12; rowID <= 51200; rowID++ {
 		row := make([]interface{}, 50)
 		for colID := 0; colID < 50; colID++ {
 			row[colID] = rand.Intn(640000)
@@ -148,7 +154,7 @@ func TestStreamWriter(t *testing.T) {
 		cells += len(row)
 	}
 	assert.NoError(t, rows.Close())
-	assert.Equal(t, 2559562, cells)
+	assert.Equal(t, 2559468, cells)
 	// Save spreadsheet with password.
 	assert.NoError(t, file.SaveAs(filepath.Join("test", "EncryptionTestStreamWriter.xlsx"), Options{Password: "password"}))
 	assert.NoError(t, file.Close())


mengpromax:
@xuri ptal， this is test suite for our biz already.

xuri:
Sure, I need to take some time to review this PR.

stream_test.go
```diff
@@ -408,6 +414,8 @@ func TestStreamSetCellValFunc(t *testing.T) {
 		true,
 		nil,
 		complex64(5 + 10i),
+		SetCellDefaultValue("100.1588"),
+		SetCellDefaultValue(" Hello"),
```
xuri, line 418: If set specific characters, it will caused corrupted workbook. For example:

```go
err := sw.SetRow("A1", []interface{}{excelize.SetCellDefaultValue("<>"))
```

xuri (commented):
Thanks for your PR. I don't think this changes is required.

The cell value with the `string` data type will be storage as `inlineStr` in the normal mode `SetCellDefault` function, the `SetCellValue` and `SetCellStr` function will storage cell value as shared string table. That's the major different between them.

But in the stream writer, the `SetRow` function always storage `string` data type cell value as `inlineStr`, the behavior same as normal mode `SetCellDefault` function, so we needn't introduce new `SetCellDefaultValue` for this.

stream_test.go
mengpromax:
```go
func (c *xlsxC) setCellDefault(value string) {
	if ok, _, _ := isNumeric(value); !ok {
		if value != "" {
			c.setInlineStr(value)
			c.IS.T.Val = value
			return
		}

		c.T, c.V, c.IS = "", "", nil // I modify this to set c.T = "" explicitly to avoid misunderstanding of c.Type can be value
		return
	}
	c.T, c.V = "", value
}


I think the difference between setCellDefault and setCellString is the former can set c.T to a empty string, i didn't find the path to do that using streaming writer.

Comment (xuri) [reply stream_test.go:418]: The cell type is empty only for no-numeric cell values, here an example for different data type of set cell values with `SetCellDefault` and stream mode function:

```go
package main

import (
    "fmt"

    "github.com/xuri/excelize/v2"
)

func main() {
    f := excelize.NewFile()
    defer func() {
            fmt.Println(err)
        }
    }()
    if err := f.SetCellDefault("Sheet1", "A1", "100.1588"); err != nil {
        fmt.Println(err)
        return
    }
    if err := f.SetCellDefault("Sheet1", "B1", "<>"); err != nil {
        fmt.Println(err)
        return
    }

    _, err := f.NewSheet("Sheet2")
    if err != nil {
        fmt.Println(err)
        return
    }
    sw, err := f.NewStreamWriter("Sheet2")
    if err != nil {
        fmt.Println(err)
        return
    }
    if err := sw.SetRow("A1", []interface{}{100.1588, "<>"}); err != nil {
        fmt.Println(err)
        return
    }
        fmt.Println(err)
        return
    }
        fmt.Println(err)
    }
}
```

the generated XML element  of Sheet1!A1 is same as Sheet2!A1, and Sheet2!B1 is same as Sheet2!B1:

```xml
<row r="1">
    <c r="A1">
        <v>100.1588</v>
    </c>
    <c r="B1" t="inlineStr">
        <is>
            <t>&lt;&gt;</t>
        </is>
    </c>
</row>
```

So you can using stream writer to generats same things that same as `SetCellDefault` function.

mengpromax: Thanks. Got it, you are right. I will check it again.

xuri: I've closed this, if you have any questions, please let me know, and I can reopen this anytime.

PR #2163, opened 2025-06-23
This closes #2024, add TempDir field in the Options data type (#2024)

### LINKED ISSUE #2024: Flexability in temporary folder in StreamWriter files
The StreamWriter feature in Excelize is great for handling large XLSX file generation. Currently, temporary files are created using Golang's os.CreateTemp, as seen in [stream.go#L740](https://github.com/qax-os/excelize/blob/30d3561d0e9bceb3a669ee864e5dcc672eff5341/stream.go#L740).

In some cases, relying on os.CreateTemp is not flexible enough. Specifically, there are scenarios where it's preferable to direct these temporary files to different or dynamically determined storage locations based on the request context. This restriction can limit the usage of StreamWriter in more complex environments.

I propose adding an option in StreamWriter to allow users to provide a custom temporary file creation method. This would follow the options pattern, maintaining the current behavior as the default. The user would be able to specify their own TempFileFactory, allowing for dynamic control over the temporary file path.

For example:
```golang
type TempFileFactory func() (*os.File, error)

var DefaultTempFileFactory TempFileFactory = func() {
    return os.CreateTemp(os.TempDir(), "excelize-")
}


// Line 740:
bw.tmp, err = newTempFile();
```

I have no problem to contribue this solution for the library if there is an agreement on a solution.

mengpromax:
Change-Id: I96e24738982be893c8a52783069221b9abeff2cf

# PR Details
1. add specific temp dir option to excelize file options.
2. when writing excel using streaming writer or read excel to load data to tmp file, it will use specific temp dir or default os.TempDir() to place temp file.

## Description
Close https://github.com/qax-os/excelize/issues/2024, add temp dir file option.

## Related Issue
https://github.com/qax-os/excelize/issues/2024

## Motivation and Context
1. when we use excelize streaming writer to write excel in our production pod, we meet that /tmp mountion is small, so we need this feature to change the temp file directory.

## How Has This Been Tested
1. UT.
2. Test in our prod env.

## Types of changes

## Checklist

### DIFF
--- a/lib.go
+++ b/lib.go
@@ -77,10 +77,17 @@ func (f *File) ReadZipReader(r *zip.Reader) (map[string][]byte, int, error) {
 	return fileList, worksheets, nil
 }
 
+func (f *File) getTempDir() string {
+	if f.options.TmpDir != "" {
+		return f.options.TmpDir
+	}
+	return "" // return "" means use the system temporary directory
+}
+
 // unzipToTemp unzip the zip entity to the system temporary directory and
 // returned the unzipped file path.
 func (f *File) unzipToTemp(zipFile *zip.File) (string, error) {
-	tmp, err := os.CreateTemp("", "excelize-")
+	tmp, err := os.CreateTemp(f.getTempDir(), "excelize-")
 	if err != nil {
 		return "", err
 	}
--- a/lib.go
+++ b/lib.go
@@ -77,17 +77,10 @@ func (f *File) ReadZipReader(r *zip.Reader) (map[string][]byte, int, error) {
 	return fileList, worksheets, nil
 }
 
-func (f *File) getTempDir() string {
-	if f.options.TmpDir != "" {
-		return f.options.TmpDir
-	}
-	return "" // return "" means use the system temporary directory
-}
-
 // unzipToTemp unzip the zip entity to the system temporary directory and
 // returned the unzipped file path.
 func (f *File) unzipToTemp(zipFile *zip.File) (string, error) {
-	tmp, err := os.CreateTemp(f.getTempDir(), "excelize-")
+	tmp, err := os.CreateTemp(f.options.TmpDir, "excelize-")
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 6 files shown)


lib.go
```diff
@@ -77,10 +77,17 @@ func (f *File) ReadZipReader(r *zip.Reader) (map[string][]byte, int, error) {
 	return fileList, worksheets, nil
 }
 
+func (f *File) getTempDir() string {
```
xuri: This function is not required, just using `os.CreateTemp(f.options.TmpDir, "excelize-")`.

mengpromax: Get, i will modify that

stream.go
```diff
@@ -737,10 +742,22 @@ func bulkAppendFields(w io.Writer, ws *xlsxWorksheet, from, to int) {
...
+	tmpDir string
+
 	tmp *os.File
 	buf bytes.Buffer
 }
 
+// getTempDir returns the temporary directory for creating temporary files, if
+// the value is empty, the system default temporary directory which is calculated
+// by os.TempDir() will be used.
+func (bw *bufferedWriter) getTempDir() string {
```
xuri: This function also can be removed. Hold on, I will make some updates based on your branch.

stream_test.go
```diff
@@ -487,3 +488,89 @@ func TestStreamWriterGetRowElement(t *testing.T) {
 		assert.False(t, ok)
 	}
 }
+
+func TestSpecificTempDirStreamingWriter(t *testing.T) {
+	checkTempFileExist := func(testName, tempDir string, fileNames []string, exist bool) {
```
xuri: The `filenames` are always empty, this test case can be removed, update the `TestStreamWriter` function to cover the custom temporary directory.

lib_test.go
```diff
@@ -390,3 +391,50 @@ func TestUnzipToTemp(t *testing.T) {
 	_, err = f.unzipToTemp(z.File[0])
 	assert.EqualError(t, err, "EOF")
 }
+
+func TestUnzipToSpecificTempDir(t *testing.T) {
```
xuri: This test function doesn't impact the code coverage.

xuri (changes requested):

xuri (approved):
Thanks for your contribution.

PR #2188, opened 2025-07-30
Optimize the GetSheetDimension implemention, make it read by stream, more fast.

Neugls:
…so not load the entire sheet into memory

# PR Details

make GetSheetDimension read xml tag by stream, using xmlDecoder which used in Rows() method.

## Description

Make GetSheetDimension read the xml file in stream, no need to load the entire xml file into memory, so the performance is much better for large Excel files.

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -2121,68 +2121,47 @@ func (f *File) SetSheetDimension(sheet, rangeRef string) error {
 
 // GetSheetDimension provides the method to get the used range of the worksheet.
 func (f *File) GetSheetDimension(sheet string) (string, error) {
-	var (
-		ref  string
-		name string
-		ok   bool
-	)
-
+	var ref string
 	if err := checkSheetName(sheet); err != nil {
 		return ref, err
 	}
-
-	if name, ok = f.getSheetXMLPath(sheet); !ok {
+	name, ok := f.getSheetXMLPath(sheet)
+	if !ok {
 		return ref, ErrSheetNotExist{sheet}
 	}
-
-	// Check if worksheet is already loaded in memory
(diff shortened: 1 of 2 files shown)


sheet.go
```diff
@@ -2121,15 +2121,68 @@ func (f *File) SetSheetDimension(sheet, rangeRef string) error {
...
+				// Found dimension element, extract ref attribute
+				for _, attr := range t.Attr {
+					if attr.Name.Local == "ref" {
+						return attr.Value, nil
+					}
+				}
+			}
+			// If we encounter sheetData, we've gone too far
+			if t.Name.Local == "sheetData" {
+				return "", fmt.Errorf("dimension not found before sheetData")
```
xuri: Error message should be defined in `errors.go`.

sheet.go
```diff
@@ -2121,15 +2121,68 @@ func (f *File) SetSheetDimension(sheet, rangeRef string) error {
...
+			}
+		}
+	}
+
+	return "", fmt.Errorf("dimension element not found")
```
xuri: The `dimension` element is not required, so we shouldn't return the error if the element not exits.

sheet.go
```diff
@@ -2121,15 +2121,68 @@ func (f *File) SetSheetDimension(sheet, rangeRef string) error {
...
+		ok   bool
+	)
+
+	if err := checkSheetName(sheet); err != nil {
 		return ref, err
 	}
-	if ws.Dimension != nil {
-		ref = ws.Dimension.Ref
+
+	if name, ok = f.getSheetXMLPath(sheet); !ok {
```
xuri: Unit tests should be added for this changes.

xuri (approved):
Thanks for your PR. I added unit tests for this changes based on your branch.

PR #2196, opened 2025-08-07
feat: add IterateColumns method for improved performance

### LINKED ISSUE #2191: support Rows.IterateColumns
### Description

Rows.Columns return []string, the performance is not good, because user usuall want to iterate []string

the difference is: user want iterate each string, not want []string

so the []string can be removed, because it cost performance, espacially big []string

maybe could support Rows.IterateColumns(), user use it like below

```go
iterateFunc := func(index int, cell string) {

}

rows.IterateColumns(iterateFunc)
```

### Additional context

_No response_

### Validations

- [x] Check that there isn't already an issue that requests the same feature to avoid creating a duplicate.

DengY11:
# PR Details

Add rows.IterateColumns(iterateFunc), iterate it with a callback function instead create a huge []string.And that will enhance memory performance

## Related Issue
close #2191 

## Example:
```go
err := rows.IterateColumns(func(index int, cell string) error {
    fmt.Printf("column %d: %s\n", index, cell)
    return nil
})
```

## How Has This Been Tested
go test

## Types of changes

## Checklist

- [] My change requires a change to the documentation.


ankisme:
Is this pr still adding? Wish to add the feature

PR #2200, opened 2025-08-16
Add custom ZIP writer support for WriteToBuffer function

a2659802:
# PR Details

## Description

Added `WriteToBufferCustomZIP` method, which allows using a custom ZIP library (e.g., `github.com/klauspost/compress/zip`) for better performance while keeping the original `WriteToBuffer` method backward-compatible.

- Introduced `ZipWriter` interface to abstract standard and custom ZIP writers.  
- `WriteToBuffer` now calls `WriteToBufferCustomZIP` with the standard library `zip.NewWriter` by default.  
- `writeToZip` method updated to accept `ZipWriter` interface.  

## Related Issue

https://github.com/qax-os/excelize/issues/2199

## Motivation and Context

To improve in-memory write performance and support high-performance third-party ZIP libraries while maintaining compatibility with the standard library.

## How Has This Been Tested

write a benchmark test for testing performance 

```go
package excel

import (
	"archive/zip"
	"bytes"
	"fmt"
	"io"
	"testing"

	kzip "github.com/klauspost/compress/zip"

	"github.com/stretchr/testify/require"
	"github.com/xuri/excelize/v2"
)

const (
	columnNum = 10
	columnLen = 100
	rowNum    = 50000
)

var data [rowNum][columnNum]string
var columnName []string
var columnNameAny []any

func init() {
	for r := range rowNum {
		for col := range columnNum {
			data[r][col] = string(bytes.Repeat([]byte{'a'}, columnLen))
		}
	}
	for range columnNum {
		columnName = append(columnName, "name")
		columnNameAny = append(columnNameAny, "name")
	}
}

func libExcelizeStreamImpl(t testing.TB, factory func(io.Writer) excelize.ZipWriter) []byte {
	file := excelize.NewFile()
	defer file.Close()
	sw, err := file.NewStreamWriter("Sheet1")
	require.NoError(t, err)

	style := &excelize.Style{Font: &excelize.Font{
		Bold: true,
	}}
	styleId, err := file.NewStyle(style)
	require.NoError(t, err)
	err = sw.SetRow("A1", columnNameAny, excelize.RowOpts{
		StyleID: styleId,
	})
	require.NoError(t, err)

	for i := range data {
		row := fmt.Sprintf("A%d", i+2)
		err := sw.SetRow(row, []any{data[i][0], data[i][1], data[i][2],
			data[i][3], data[i][4], data[i][5], data[i][6], data[i][7], data[i][8], data[i][9]})
		require.NoError(t, err)
	}

	require.NoError(t, sw.Flush())
	// buffer, err := file.WriteToBuffer()
	buffer, err := file.WriteToBufferCustomZIP(factory)
	require.NoError(t, err)
	return buffer.Bytes()
}

func BenchmarkExcelizeStreamStd(b *testing.B) {
	for b.Loop() {
		_ = libExcelizeStreamImpl(b, func(w io.Writer) excelize.ZipWriter { return zip.NewWriter(w) })
	}
}

func BenchmarkExcelizeStreamKzip(b *testing.B) {
	for b.Loop() {
		_ = libExcelizeStreamImpl(b, func(w io.Writer) excelize.ZipWriter { return kzip.NewWriter(w) })
	}
}

```

```
Running tool: D:\soft\Go125\bin\go.exe test -benchmem -run=^$ -bench ^(BenchmarkExcelizeStreamStd|BenchmarkExcelizeStreamKzip)$ github.com/a2659802/go-example/pkg/excel -count=1

goos: windows
goarch: amd64
pkg: github.com/a2659802/go-example/pkg/excel
cpu: 12th Gen Intel(R) Core(TM) i5-12500
=== RUN   BenchmarkExcelizeStreamStd
BenchmarkExcelizeStreamStd
BenchmarkExcelizeStreamStd-12
       2         599544500 ns/op        272403848 B/op   4153804 allocs/op
=== RUN   BenchmarkExcelizeStreamKzip
BenchmarkExcelizeStreamKzip
BenchmarkExcelizeStreamKzip-12
       3         468739967 ns/op        272623032 B/op   4152989 allocs/op
PASS
ok      github.com/a2659802/go-example/pkg/excel        3.077s
```

## Types of changes

## Checklist

### DIFF
--- a/file.go
+++ b/file.go
@@ -12,12 +12,10 @@
 package excelize
 
 import (
-	"archive/zip"
 	"bytes"
 	"encoding/binary"
 	"encoding/xml"
 	"io"
-	"io/fs"
 	"math"
 	"os"
 	"path/filepath"
@@ -137,22 +135,8 @@ func (f *File) WriteTo(w io.Writer, opts ...Options) (int64, error) {
 // WriteToBuffer provides a function to get bytes.Buffer from the saved file,
 // and it allocates space in memory. Be careful when the file size is large.
 func (f *File) WriteToBuffer() (*bytes.Buffer, error) {
-	return f.WriteToBufferCustomZIP(func(w io.Writer) ZipWriter { return zip.NewWriter(w) })
-}
-
-type ZipWriter interface {
-	Create(name string) (io.Writer, error)
-	AddFS(fsys fs.FS) error
-	Close() error
-}
-
-// WriteToBufferCustomZIP provides a function to write to a buffer using a custom ZIP writer
-// factory. This allows for custom ZIP compression implementations to be used when saving
-// the spreadsheet to memory. The factory function should return a zipWriter interface
-// that will be used to create the final ZIP archive.
-func (f *File) WriteToBufferCustomZIP(factory func(io.Writer) ZipWriter) (*bytes.Buffer, error) {
 	buf := new(bytes.Buffer)
-	zw := factory(buf)
+	zw := f.ZipWriter(buf)
 
 	if err := f.writeToZip(zw); err != nil {
 		_ = zw.Close()
@@ -173,7 +157,7 @@ func (f *File) WriteToBufferCustomZIP(factory func(io.Writer) ZipWriter) (*bytes
 	return buf, nil
 }
 
-// writeToZip provides a function to write to zipWriter
+// writeToZip provides a function to write to ZipWriter.
 func (f *File) writeToZip(zw ZipWriter) error {
 	f.calcChainWriter()
 	f.commentsWriter()
--- a/file.go
+++ b/file.go
@@ -17,6 +17,7 @@ import (
 	"encoding/binary"
 	"encoding/xml"
 	"io"
+	"io/fs"
 	"math"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


file.go
```diff
@@ -136,8 +137,22 @@ func (f *File) WriteTo(w io.Writer, opts ...Options) (int64, error) {
...
+	Create(name string) (io.Writer, error)
+	AddFS(fsys fs.FS) error
+	Close() error
+}
+
+// WriteToBufferCustomZIP provides a function to write to a buffer using a custom ZIP writer
+// factory. This allows for custom ZIP compression implementations to be used when saving
+// the spreadsheet to memory. The factory function should return a zipWriter interface
+// that will be used to create the final ZIP archive.
+func (f *File) WriteToBufferCustomZIP(factory func(io.Writer) ZipWriter) (*bytes.Buffer, error) {
```
xuri: The `Save` and `SaveAs` function also should be support save workbook with custom zip writer. Introduce new function `SetZipWriter` just like `CharsetTranscoder` instead of this would be better approach.

xuri (changes requested):

xuri (commented):
Note that the third-party archive package kzip you mentioned reduce 25% time cost, but the  size of generated workbook will be increase over 100%.

xuri (approved):

Issue #2205, opened 2025-08-31
Issue with cell character count limitation


torotake:
### Description

In Excel, both the 32,767-character cell limit and the 31-character sheet name limit are based on UTF-16 units. Since surrogate pairs count as two units, strings containing such characters may exceed these limits even if they appear shorter in rune length. In excelize, truncation is done by rune length, which can produce invalid files that trigger a “Problem with some content” alert in Excel.

<img width="260" height="260" alt="Image" src="https://github.com/user-attachments/assets/0f7f52a1-986d-43f7-9851-7502a4b7eb02" />

### Steps to reproduce the issue

```go
package main

import (
	"log"
	"strings"

	"github.com/xuri/excelize/v2"
)

func main() {
	f := excelize.NewFile()

	// "😀" and "𩸽" are both represented as surrogate pairs.
	// The text length is 32,767 runes in Go, but 32,769 units in UTF-16.
	text := strings.Repeat("A", 32767-2) + "😀𩸽"

	f.SetCellValue("Sheet1", "A1", text)
	if err := f.SaveAs("cell_limit_error.xlsx"); err != nil {
		log.Fatalf("failed to save file: %v", err)
	}
}
```

Run the code above, then open the generated xlsx file in Excel.

### Describe the results you received

“Problem with some content” alert in Excel.

### Describe the results you expected

The expected result is that when the string to be written to a cell is too long, it is properly truncated and a valid xlsx file is generated without errors.

### Go version

1.25.0

### Excelize version or commit ID

2.9.1

### Environment

```shell
macOS 15.5 / Excel for Mac 16.93.1 (25011917)
Windows 11 Home 24H / Excel 2508
```

### Validations

- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a minimal reproducible example of the bug.

PR #2206, opened 2025-08-31
Fix string counting and truncation for cell length limitation #2205

torotake:
# PR Details

Add UTF-16 code unit aware helpers (utf16UnitCountInString, truncateUTF16Units) and update sheet name validation to use code unit length.

## Description

This change introduces two new internal helpers:

- utf16UnitCountInString: counts UTF-16 code units (BMP runes =1, supplementary runes =2).
- truncateUTF16Units: safely truncates a string to a maximum number of UTF-16 code units without splitting surrogate pairs.

Sheet name validation (checkSheetName) now enforces the 31 unit limit using UTF-16 code units instead of plain rune count, aligning behavior with Excel’s internal UTF-16 length semantics (emoji / supplementary characters consume two units). This prevents accepting names that Excel would later reject or truncating inside a surrogate pair in downstream consumers.

## Related Issue

https://github.com/qax-os/excelize/issues/2205

## Motivation and Context

The Excel limit is effectively a UTF-16 unit, not a Unicode scalar number.
The previous length check did not properly handle strings containing surrogate pairs, resulting in problematic Excel files.
This will be fixed so that correct Excel files are output.

## How Has This Been Tested

- Count correctness for BMP-only strings.
- Mixed BMP + supplementary (emoji) sequences.
- Boundary where max falls exactly before / inside a surrogate pair.
- Zero and negative max cases (empty result).
- No-op when already within limit.
- Truncation preserves whole characters (no half surrogate).

Testing Environment:
Go version: go version go1.24.2 darwin/arm64
OS: macOS 15.5

## Types of changes

Stricter rejection of sheet names that previously slipped through when they exceeded the true UTF-16 unit limit via supplementary characters. This aligns behavior with Excel and is considered a correctness fix.

## Checklist

### DIFF
--- a/lib.go
+++ b/lib.go
@@ -941,22 +941,12 @@ func setPtrFieldsVal(fields []string, immutable, mutable reflect.Value) {
 	}
 }
 
-// utf16UnitCountInString returns the number of UTF-16 code units in a string.
-func utf16UnitCountInString(s string) int {
-	var count int
-	for _, r := range s {
-		count += utf16.RuneLen(r)
-	}
-	return count
-}
-
-// truncateUTF16Units truncates a string to a maximum number of UTF-16 code units.
-func truncateUTF16Units(s string, maxLength int) string {
-	var count int
+// truncateUTF16Units truncates a string to a maximum number of UTF-16 code
+// units.
+func truncateUTF16Units(s string, length int) string {
+	var cnt int
 	for i, r := range s {
-		count += utf16.RuneLen(r)
-		if count > maxLength {
-			// If s[maxLength-1] is a high surrogate, it is also truncated.
+		if cnt += utf16.RuneLen(r); cnt > length {
 			return s[:i]
 		}
 	}
--- a/lib.go
+++ b/lib.go
@@ -25,6 +25,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"unicode/utf16"
 )
 
 // ReadZipReader extract spreadsheet with given options.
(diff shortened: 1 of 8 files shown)


lib.go
```diff
@@ -940,6 +941,28 @@ func setPtrFieldsVal(fields []string, immutable, mutable reflect.Value) {
 	}
 }
 
+// utf16UnitCountInString returns the number of UTF-16 code units in a string.
```
xuri: This function can be simplify by `len(utf16.Encode([]rune(s)))`.

lib.go
```diff
@@ -940,6 +941,28 @@ func setPtrFieldsVal(fields []string, immutable, mutable reflect.Value) {
...
+	}
+	return count
+}
+
+// truncateUTF16Units truncates a string to a maximum number of UTF-16 code units.
+func truncateUTF16Units(s string, maxLength int) string {
+	var count int
+	for i, r := range s {
+		count += utf16.RuneLen(r)
+		if count > maxLength {
```
xuri: This code block can be simplify like this:

```go
if count += utf16.RuneLen(r); count > maxLength {
    return s[:i]
}
```

lib_test.go
```diff
@@ -390,3 +390,45 @@ func TestUnzipToTemp(t *testing.T) {
 	_, err = f.unzipToTemp(z.File[0])
 	assert.EqualError(t, err, "EOF")
 }
+
+func TestUTF16UnitCountInString(t *testing.T) {
+	cases := map[string]int{
+		"":    0,
+		"ABC": 3, // all Basic Multilingual Plane 1 unit each
+		"你好":  2, // Chinese Basic Multilingual Plane chars
+		"😀":   2, // single surrogate pair
```
xuri: Encode emoji with unicode `\U0001F600`.

xuri (changes requested):

lib.go
torotake: Thank you for the review and merge!

I think this is fine if the priority is to keep the code simple.

Just to add a note: since both []rune and []uint16 are allocated in memory at the same time, the original code was written to minimize memory usage in case a long string is passed in.

xuri: Thanks for your notice, I missed that point, will back port for it.

torotake: Thanks a lot for noticing and making the change! I really appreciate it.

PR #2229, opened 2025-10-28
feat: add example to setDefinedName

kenny-not-dead:
# PR Details

add example to setDefinedName

## Description

I'm actively using your project, including implementing a full-featured Excel document import and export, but I was missing an example for this particular function. The first example mentions the option to omit the scope, but the example itself does. I also thought it would be helpful to add an example with the function, since it works.

## Motivation and Context

as I indicated above, it is in this function that I would like to see more examples, especially with the function

## Types of changes

## Checklist

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -1733,7 +1733,6 @@ func (f *File) GetPageLayout(sheet string) (PageLayoutOptions, error) {
 //	    Name:     "Amount",
 //	    RefersTo: "Sheet1!$A$2:$D$5",
 //	    Comment:  "defined name comment",
-//	    Scope:    "Sheet2",
 //	})
 //
 // If you fill the RefersTo property with only one columns range without a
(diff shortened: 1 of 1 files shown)


sheet.go
```diff
@@ -1753,6 +1752,14 @@ func (f *File) GetPageLayout(sheet string) (PageLayoutOptions, error) {
 //	    RefersTo: "Sheet1!$1:$1",
 //	    Scope:    "Sheet1",
 //	})
+//
+// You can also use the function in RefersTo. For example:
+//
+//	err := f.SetDefinedName(&excelize.DefinedName{
```
xuri: Remove double quote after plus sign, and using new Name different with other examples would be better. I suggest adding example like this:

```text
err := f.SetDefinedName(&excelize.DefinedName{
    Name:     "CustomRange",
    RefersTo: "Sheet1!$A$2+Sheet1!$D$5",
    Scope:    "Sheet1",
})
```

xuri (changes requested):

kenny-not-dead:
@xuri 
Thanks for your comment, I've corrected it

kenny-not-dead:
@xuri 

Could you please clarify the policy for merging this same commit into v2? After merging this, can I create a separate PR for the v2 branch?

kenny-not-dead:
Should I be worried about the dropped dough?

sheet.go
```diff
@@ -1753,6 +1752,14 @@ func (f *File) GetPageLayout(sheet string) (PageLayoutOptions, error) {
 //	    RefersTo: "Sheet1!$1:$1",
 //	    Scope:    "Sheet1",
 //	})
+//
+// You can also use the function in RefersTo. For example:
+//
+// 	err := f.SetDefinedName(&excelize.DefinedName{
```
xuri, line 1758: Note that we need to use tab instead of space indent in doc code block. The space indent may cause [godoc](https://pkg.go.dev/golang.org/x/tools/cmd/godoc) tool cannot generate code block correctly.

kenny-not-dead: Thanks for the clarification, but I see you've already merged. Should I make changes to address that?

xuri: Sure, thanks! Sorry for missing that before merging this PR.

kenny-not-dead: I'll do it later this evening, thanks

PR #2232, opened 2025-10-30
Fix: AddPictureFromBytes fails on SVG (image: unknown format)

pegasscience-cyber:
Summary

This patch fixes a regression where AddPictureFromBytes() fails to insert .svg images with the error:
**image: unknown format**

The issue occurs because the standard Go image library does not support SVG, and the previous implementation always called image.DecodeConfig for every image type.

This change adds SVG-specific handling to bypass raster decoding and derive dimensions directly from the SVG metadata.


go.mod
```diff
@@ -1,17 +1,17 @@
 module github.com/xuri/excelize/v2
 
-go 1.24.0
+go 1.23.3
```
xuri: Don't change go.mod file and downgrade dependencies package version.

picture.go
```diff
@@ -286,6 +298,109 @@ func (f *File) AddPictureFromBytes(sheet, cell string, pic *Picture) error {
 	return err
 }
 
+// --- SVG helpers ---
```
xuri: Parse image file is out of scope of this library, user can register custom decoder for multiple image formats, for example:

```diff
package main

import (
    "fmt"
    "image"
    "io"
    "os"

    "github.com/xuri/excelize/v2"
)

func main() {
    f := excelize.NewFile()
    defer func() {
        if err := f.Close(); err != nil {
            fmt.Println(err)
        }
    }()
    file, err := os.ReadFile("img.svg")
    if err != nil {
        fmt.Println(err)
        return
    }
+   decode := func(r io.Reader) (image.Image, error) { return nil, nil }
+   decodeConfig := func(r io.Reader) (image.Config, error) { return image.Config{Height: 100, Width: 90}, nil }
+   image.RegisterFormat(".svg", "", decode, decodeConfig)
    if err := f.AddPictureFromBytes("Sheet1", "A2", &excelize.Picture{
        Extension: ".svg",
        File:      file,
        Format:    &excelize.GraphicOptions{AltText: "Excel Logo"},
    }); err != nil {
        fmt.Println(err)
        return
    }
    if err := f.SaveAs("Book1.xlsx"); err != nil {
        fmt.Println(err)
    }
}
```

So, I think this changes is not required.

xuri (changes requested):

PR #2235, opened 2025-11-11
Fix: skipped column style bug in StreamWriter's SetRow

radam9:
# PR Details

There was a small bug in `StreamWriter`'s `SetRow`, where the column styles where not applied due to a missing col increment in `prepareCellStyle`

## Description

Properly added the index value `i` to the column number `col` before passing to `prepareCellStyle`

## Motivation and Context

It solves missing column styles when there are no cell or row styles.

## How Has This Been Tested

updated the `TestStreamSetRowWithStyle` test to include a check for the column style

## Types of changes

## Checklist

### DIFF
--- a/stream_test.go
+++ b/stream_test.go
@@ -363,26 +363,23 @@ func TestStreamSetRowWithStyle(t *testing.T) {
 	assert.NoError(t, err)
 
 	sheetName := "Sheet1"
-	err = file.SetColStyle(sheetName, "A", grayStyleID)
-	assert.NoError(t, err)
-	err = file.SetColStyle(sheetName, "C", blueStyleID)
-	assert.NoError(t, err)
-
 	streamWriter, err := file.NewStreamWriter(sheetName)
 	assert.NoError(t, err)
+	assert.NoError(t, streamWriter.SetColStyle(1, 1, grayStyleID))
+	assert.NoError(t, streamWriter.SetColStyle(3, 3, blueStyleID))
 	assert.NoError(t, streamWriter.SetRow("A1", []interface{}{
-		"value11",
-		Cell{Value: "value12"},
-		&Cell{Value: "value12"},
-		Cell{StyleID: blueStyleID, Value: "value13"},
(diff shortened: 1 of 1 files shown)


xuri (changes requested):

stream_test.go
```diff
@@ -362,22 +362,38 @@ func TestStreamSetRowWithStyle(t *testing.T) {
 	blueStyleID, err := file.NewStyle(&Style{Font: &Font{Color: "0000FF"}})
 	assert.NoError(t, err)
 
-	streamWriter, err := file.NewStreamWriter("Sheet1")
+	sheetName := "Sheet1"
+	err = file.SetColStyle(sheetName, "A", grayStyleID)
```
xuri: As documentation says, this library doesn't allow mix using normal mode function and stream mode function in a worksheet, so we need to use `streamWriter.SetColStyle` instead of `file.SetColStyle` at here:

```go
streamWriter, err := file.NewStreamWriter(sheetName)
assert.NoError(t, err)
assert.NoError(t, streamWriter.SetColStyle(1, 1, grayStyleID))
assert.NoError(t, streamWriter.SetColStyle(3, 3, blueStyleID))
```

xuri (changes requested):

stream_test.go
xuri: I've made some changes based on your branch to fixed this.

radam9: oops, my bad, mixed those up.
Thanks for fixing it.

radam9:
> Thanks for your PR. I've left some comments.

I only see a single comment, which you made changes to fix.

xuri:
Yeah, another suggestion was change `valueXX` to cell coordinates name in test to made verification convenience. I also updated this in the last commit.

xuri (approved):
LGTM. Thanks for your contribution.

radam9:
Thanks for quickly fixing and merging this 👍🏻

PR #2244, opened 2025-12-11
feat: get rows and cols count

kenny-not-dead:
# PR Details

 get rows and cols count

## Description

While using it, I noticed that I wanted to get the maximum number of rows and columns used in an Excel file, but I couldn't find a method that could return me exactly this result... So I think this might be useful.

## Motivation and Context

as i indicated above this is a faster way to get this number

## Types of changes

## Checklist


kenny-not-dead:
@xuri Please tell me, do you think it's possible to include this code? If so, then I would add tests.

PR #2267, opened 2026-02-27
Fix SUMIF/COUNTIF exact text matching to not match substrings

jpoz:
# PR Details

Fix `formulaCriteriaParser` to anchor regex patterns so that exact text criteria in SUMIF/COUNTIF do not incorrectly match substrings.

## Description

Modified `formulaCriteriaParser` to wrap regex patterns with `^` and `$` anchors. Previously, criteria like `"administrative"` would match substrings such as `"cyclical_flat_administrative"`, producing incorrect results. With this fix, exact text criteria only match full cell values, while wildcard patterns (`*`, `?`) continue to work as expected.

## Related Issue

N/A

## Motivation and Context

When using `SUMIF(A2:A7,"administrative",B2:B7)`, the formula was matching cells containing "cyclical_flat_administrative" and "pre_administrative_post" in addition to "administrative". This is incorrect behavior — Excel only matches exact values when no wildcards are present. The fix ensures parity with Excel's behavior.

## How Has This Been Tested

Added `TestCalcSUMIFExactMatch` which verifies:
- Exact text match (`"administrative"`) only matches full cell values (sum = 250, count = 2)
- Wildcard `*administrative*` matches all cells containing the substring (sum = 750, count = 4)
- Wildcard `administrative*` matches cells starting with the term
- Wildcard `*administrative` matches cells ending with the term
- Single char wildcard `othe?` matches as expected
- All existing tests continue to pass

## Types of changes

## Checklist

### DIFF
--- a/calc_test.go
+++ b/calc_test.go
@@ -5727,17 +5727,26 @@ func TestCalcSUMIFExactMatch(t *testing.T) {
 	}
 	f := prepareCalcData(cellData)
 	formulaList := map[string]string{
-		`SUMIF(A2:A7,"text",B2:B7)`:   "300",
-		`SUMIF(A2:A7,"*text*",B2:B7)`: "800",
-		`SUMIF(A2:A7,"text*",B2:B7)`:  "300",
-		`SUMIF(A2:A7,"*text",B2:B7)`:  "500",
-		`SUMIF(A2:A7,"othe?",B2:B7)`:  "400",
-		`SUMIF(A2:A7,"*",B2:B7)`:      "1200",
-		`SUMIF(A2:A7,".",B2:B7)`:      "0",
-		`SUMIF(A2:A7,".*",B2:B7)`:     "0",
-		`SUMIF(A2:A7,".?",B2:B7)`:     "0",
-		`COUNTIF(A2:A7,"text")`:       "3",
-		`COUNTIF(A2:A7,"*text*")`:     "5",
+		`SUMIF(A2:A7,"text",B2:B7)`:           "300",
+		`COUNTIF(A2:A7,"text")`:               "3",
+		`SUMIF(A2:A7,".",B2:B7)`:              "0",
+		`SUMIF(A2:A7,".*",B2:B7)`:             "0",
+		`SUMIF(A2:A7,".?",B2:B7)`:             "0",
+		`SUMIF(A2:A7,"*text*",B2:B7)`:         "800",
+		`SUMIF(A2:A7,"text*",B2:B7)`:          "300",
+		`SUMIF(A2:A7,"*text",B2:B7)`:          "500",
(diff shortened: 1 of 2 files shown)


calc_test.go
```diff
@@ -5739,6 +5715,41 @@ func TestCalcSUMIFSAndAVERAGEIFS(t *testing.T) {
...
+		{"cyclical_flat_administrative", 200},
+		{"administrative", 150},
+		{"pre_administrative_post", 300},
+		{"ADMINISTRATIVE", 50},
+		{"other", 400},
+	}
+	f := prepareCalcData(cellData)
+	formulaList := map[string]string{
+		// Exact text match should not match substrings
+		`SUMIF(A2:A7,"administrative",B2:B7)`: "250",
```
xuri: The expected value is not consistent with Excel, it should be 300.

calc_test.go
```diff
@@ -5739,6 +5715,41 @@ func TestCalcSUMIFSAndAVERAGEIFS(t *testing.T) {
...
+		// Wildcard * should match substrings (case-sensitive)
+		`SUMIF(A2:A7,"*administrative*",B2:B7)`: "750",
```
xuri: Value should be 800.

calc_test.go
```diff
@@ -5739,6 +5715,41 @@ func TestCalcSUMIFSAndAVERAGEIFS(t *testing.T) {
...
+		// Wildcard * at end only
+		`SUMIF(A2:A7,"administrative*",B2:B7)`: "250",
```
xuri: Value should be 300.

calc_test.go
```diff
@@ -5739,6 +5715,41 @@ func TestCalcSUMIFSAndAVERAGEIFS(t *testing.T) {
...
+		// Wildcard * at start only
+		`SUMIF(A2:A7,"*administrative",B2:B7)`: "450",
```
xuri: Value should be 500.

calc_test.go
```diff
@@ -5739,6 +5715,41 @@ func TestCalcSUMIFSAndAVERAGEIFS(t *testing.T) {
...
+		// Single char wildcard ?
+		`SUMIF(A2:A7,"othe?",B2:B7)`: "400",
+		// COUNTIF exact match should not match substrings
+		`COUNTIF(A2:A7,"administrative")`: "2",
```
xuri: Value should be 3.

calc_test.go
```diff
@@ -5739,6 +5715,41 @@ func TestCalcSUMIFSAndAVERAGEIFS(t *testing.T) {
...
+		// COUNTIF wildcard match
+		`COUNTIF(A2:A7,"*administrative*")`: "4",
```
xuri: Value should be 4.

PR #2268, opened 2026-02-27
feat(calc): add implicit intersection and fix regex criteria anchoring

jpoz:
# PR Details

## Description

This PR implements two related fixes for the calculation engine:

1. **Implicit Intersection**: Adds `implicitIntersect` method that resolves matrix arguments to scalar values based on the formula cell's row position. This matches Excel's behavior where range references in non-array formulas resolve to the cell in the same row as the formula. Applied to `ABS`, `ISNUMBER`, and `IF` functions.

2. **Regex Criteria Anchoring Fix**: Moves `^`/`$` anchor insertion from `formulaCriteriaParser` to `formulaCriteriaEval`. The previous approach (anchoring at parse time) interfered with numeric criteria parsing since `"^5$"` cannot convert to a number. Anchoring at eval time preserves correct `ToNumber()` behavior while still enforcing exact matching in SUMIF/COUNTIF.

3. **ISNUMBER matrix bug fix**: The previous matrix handling code in `ISNUMBER` had a missing `else` branch, causing it to append both `true` and `false` for number cells, corrupting the output matrix. Replaced with `implicitIntersect` for correct scalar resolution.

## Related Issue

Fixes implicit intersection behavior for scalar functions receiving range/matrix arguments.

## Motivation and Context

When a whole-column or range reference is passed to a scalar function like `ABS(A1:A5)`, Excel resolves it to the single cell in the same row as the formula cell (implicit intersection). Without this fix, the calculation engine could not handle these common spreadsheet patterns correctly. The regex anchoring fix ensures criteria-based functions (SUMIF, COUNTIF, etc.) correctly match exact values without breaking numeric criteria handling.

## How Has This Been Tested

- Added `TestCalcImplicitIntersect`: validates ABS, ISNUMBER, and IF with range references across multiple rows, confirming each resolves to the correct row via implicit intersection (11 assertions)
- Added `TestCalcCriteriaRegexpAnchoring`: validates exact matching, wildcard patterns (`*`, `?`), and numeric criteria for SUMIF/COUNTIF
- All existing tests pass including `TestCalcSUMIFExactMatch`, `TestCalcAVERAGEIF`, and `TestCalcCellValue`

## Types of changes

## Checklist

### DIFF
--- a/calc_test.go
+++ b/calc_test.go
@@ -5778,6 +5778,28 @@ func TestCalcSUMIFExactMatch(t *testing.T) {
 		assert.NoError(t, err, formula)
 		assert.Equal(t, expected, result, formula)
 	}
+	cellData = [][]interface{}{
+		{"Category", "Amount"},
+		{"text", 10},
+		{"***text", 20},
+		{"text ***", 30},
+		{"TEXT", 40},
+		{5, 50},
+		{5.5, 60},
+	}
+	f = prepareCalcData(cellData)
+	for formula, expected := range map[string]string{
+		`SUMIF(A2:A7,"text",B2:B7)`:  "50",
+		`SUMIF(A2:A7,"*text",B2:B7)`: "70",
+		`SUMIF(A2:A7,"text*",B2:B7)`: "80",
+		`SUMIF(A2:A7,5,B2:B7)`:       "50",
+		`COUNTIF(A2:A7,"tex?")`:      "2",
+	} {
+		assert.NoError(t, f.SetCellFormula("Sheet1", "C1", formula))
+		result, err := f.CalcCellValue("Sheet1", "C1")
+		assert.NoError(t, err, formula)
+		assert.Equal(t, expected, result, formula)
+	}
 }
 
 func TestCalcXIRR(t *testing.T) {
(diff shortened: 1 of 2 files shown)


calc_test.go
```diff
@@ -7006,3 +7006,113 @@ func TestCalcTrendGrowthRegression(t *testing.T) {
...
+		{"pineapple", 20},
+		{"apple pie", 30},
+		{"APPLE", 40},
+		{5, 50},
+		{5.5, 60},
+	}
+	f := prepareCalcData(cellData)
+	for formula, expected := range map[string]string{
+		// Exact text match: only "apple" rows (case-insensitive in Excel)
+		`SUMIF(A2:A7,"apple",B2:B7)`: "10",
```
xuri: Expected result should be 50, not 10.

calc_test.go
```diff
@@ -7006,3 +7006,113 @@ func TestCalcTrendGrowthRegression(t *testing.T) {
...
+		// Wildcard at start: matches "pineapple" and "apple"
+		`SUMIF(A2:A7,"*apple",B2:B7)`: "30",
```
xuri: Expected result should be 70.

calc_test.go
```diff
@@ -7006,3 +7006,113 @@ func TestCalcTrendGrowthRegression(t *testing.T) {
...
+		// Wildcard at end: matches "apple" and "apple pie"
+		`SUMIF(A2:A7,"apple*",B2:B7)`: "40",
```
xuri: Expected result should be 80.

calc_test.go
```diff
@@ -7006,3 +7006,113 @@ func TestCalcTrendGrowthRegression(t *testing.T) {
...
+		// Single char wildcard
+		`COUNTIF(A2:A7,"appl?")`: "1",
```
xuri: Expected result should be 2.

xuri (approved):
Thanks for your contribution.

PR #2269, opened 2026-02-28
fix: preserve matrix data for range refs in function infix expressions

jpoz:
## Description

When a range token (e.g. `Ledger!H:H`) is the first operand in an infix expression inside a function argument (e.g. `IF(Ledger!H:H="TRANSFER",...)`), there was no handler in `evalInfixExp` for the case where the next token is an infix operator. The range fell through to `parseToken` which converted the `ArgMatrix` to a scalar string via `formulaArgToToken`, losing all matrix data. The entire column reference was reduced to just its first cell value before the comparison was evaluated.

This caused formulas like `MAX(IF(Sheet!Col:Col="X", ABS(Sheet!Col:Col)))/100` to always return 0.

Changes:
- Add element-wise array operations for all comparison and arithmetic operators with broadcasting support
- Add implicit intersection for whole-column/row references in scalar function contexts
- Add condition in `evalInfixExp` to parse range references directly when followed by infix operators inside functions, preserving matrix data

## Related Issue

N/A - discovered via user-reported incorrect calculation results.

## Motivation and Context

Formulas using the common `MAX(IF(range=criteria, values))` pattern with whole-column references (e.g. `H:H`, `F:F`) returned 0 instead of the correct value. This is because the column range matrix was being flattened to a scalar before the comparison operator could perform element-wise evaluation.

## How Has This Been Tested

- Added `TestCalcRangeInFunctionInfixOp` test covering:
  - `IF` with whole-column range comparison and implicit intersection across multiple rows
  - `MAX(IF(...))` wrapping with matching and non-matching categories
  - `ABS` and division within the pattern
  - `<>` operator with range as first operand
- All existing `TestCalc*` tests pass
- Full test suite passes (`go test ./...`)
- Verified with an xlsx file containing `MAX(IF(Ledger!H:H="TRANSFER",ABS(Ledger!F:F)))/100` — now returns correct value instead of `$0.00`

## Types of changes

## Checklist

### DIFF
--- a/calc.go
+++ b/calc.go
@@ -2200,14 +2200,7 @@ func formulaCriteriaEval(val formulaArg, criteria *formulaCriteria) (result bool
 			}
 		}
 	case criteriaRegexp:
-		pattern := criteria.Condition.Value()
-		if !strings.HasPrefix(pattern, "^") {
-			pattern = "^" + pattern
-		}
-		if !strings.HasSuffix(pattern, "$") {
-			pattern = pattern + "$"
-		}
-		return regexp.MatchString(pattern, val.Value())
+		return regexp.MatchString(criteria.Condition.Value(), val.Value())
 	}
 	return
 }
--- a/calc.go
+++ b/calc.go
@@ -377,6 +377,27 @@ type formulaFuncs struct {
 	sheet, cell string
 }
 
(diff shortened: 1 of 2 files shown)


xuri:
Thanks for the PR! If you'd like to reopen this, please go ahead. Just keep in mind I’ll need some time to review it since it’s quite large.

PR #2291, opened 2026-04-10
fix(comment): assign correct authorID in addComment when author already exists

### LINKED ISSUE #2289: AddComment assigns authorID 0 for all comments when author already exists in authors list
## Description

`addComment()` in `vml.go` has a bug where the `authorID` variable defaults to `0` when the specified author already exists in the comment authors list. Only the *first* `AddComment` call for a new author correctly assigns the author index; all subsequent calls for the same author silently fall through to `authorID = 0`.

This means every comment after the first is attributed to whichever author happens to be at index 0 in `xl/comments*.xml`, rather than the intended author.

## Steps to reproduce the issue

```go
package main

import (
    "fmt"
    "github.com/xuri/excelize/v2"
)

func main() {
    f := excelize.NewFile()

    // First comment creates "ExistingUser" as authorId=0
    f.AddComment("Sheet1", excelize.Comment{
        Cell: "A1", Author: "ExistingUser",
        Paragraph: []excelize.RichTextRun{{Text: "existing"}},
    })

    // First "Agent" comment — correctly creates authorId=1
    f.AddComment("Sheet1", excelize.Comment{
        Cell: "B1", Author: "Agent",
        Paragraph: []excelize.RichTextRun{{Text: "first"}},
    })

    // Second "Agent" comment — BUG: gets authorId=0 ("ExistingUser")
    f.AddComment("Sheet1", excelize.Comment{
        Cell: "C1", Author: "Agent",
        Paragraph: []excelize.RichTextRun{{Text: "second"}},
    })

    comments, _ := f.GetComments("Sheet1")
    for _, c := range comments {
        fmt.Printf("Cell %s: Author=%q\n", c.Cell, c.Author)
    }
    // Output:
    //   Cell A1: Author="ExistingUser"
    //   Cell B1: Author="Agent"         <- correct
    //   Cell C1: Author="ExistingUser"  <- WRONG, should be "Agent"
}
```

## Describe the results you received

All `AddComment` calls after the first for a given author produce comments attributed to the author at index 0, regardless of the actual author specified.

```
Cell A1: Author="ExistingUser"
Cell B1: Author="Agent"
Cell C1: Author="ExistingUser"   <- wrong
```

## Describe the results you expected

```
Cell A1: Author="ExistingUser"
Cell B1: Author="Agent"
Cell C1: Author="Agent"          <- correct
```

## Root cause

In `vml.go`, function `addComment` (line ~271):

```go
var authorID int   // defaults to 0
if cmts == nil {
    cmts = &xlsxComments{Authors: xlsxAuthor{Author: []string{opts.Author}}}
}
if inStrSlice(cmts.Authors.Author, opts.Author, true) == -1 {
    cmts.Authors.Author = append(cmts.Authors.Author, opts.Author)
    authorID = len(cmts.Authors.Author) - 1
}
```

When `inStrSlice` finds the author (returns >= 0), the code skips the `if` block but never assigns the found index to `authorID`. It stays at its zero-value `0`.

## Go version

1.25.4

## Excelize version or commit ID

v2.10.0

## Environment

Linux (x86_64), tested programmatically — no Excel version dependency.

narasaka:
# PR Details

## Description

`addComment()` in `vml.go` has a bug where `authorID` defaults to `0` when the author already exists in the comment authors list. only the first `AddComment` call for a new author correctly creates the author entry and assigns the right index. all subsequent calls with the same author skip the append but never set `authorID` to the found index — it stays at its zero-value `0`.

the fix stores the return value of `inStrSlice` and uses it as `authorID` in the `else` branch:

```go
// before
var authorID int
if inStrSlice(cmts.Authors.Author, opts.Author, true) == -1 {
    cmts.Authors.Author = append(cmts.Authors.Author, opts.Author)
    authorID = len(cmts.Authors.Author) - 1
}

// after
var authorID int
if idx := inStrSlice(cmts.Authors.Author, opts.Author, true); idx == -1 {
    cmts.Authors.Author = append(cmts.Authors.Author, opts.Author)
    authorID = len(cmts.Authors.Author) - 1
} else {
    authorID = idx
}
```

## Related Issue

Closes #2289

## Motivation and Context

when a workbook already contains comments from other authors (e.g. threaded comments created by Excel), every new comment written by a second author gets silently attributed to index 0 instead. in practice this turns yellow sticky-note comments into threaded discussion comments (because the author at index 0 may be a `tc={GUID}` threaded-comment author), which changes their behavior and length limits in Excel.

## How Has This Been Tested

- added `TestAddCommentExistingAuthorID` regression test
  - creates a comment with author "Alice" (authorId=0)
  - adds 3 comments with author "Bob"
  - asserts all 3 "Bob" comments have `Author == "Bob"`, not "Alice"
- confirmed the test **fails** on `v2.10.0` (C1 and D1 get "Alice") and **passes** with this fix
- all existing tests pass (`go test ./...`)

## Types of changes

## Checklist

### DIFF
--- a/vml.go
+++ b/vml.go
@@ -275,18 +275,16 @@ func (f *File) addComment(commentsXML string, opts vmlOptions) error {
 	if err != nil {
 		return err
 	}
-	var authorID int
 	if cmts == nil {
 		cmts = &xlsxComments{Authors: xlsxAuthor{Author: []string{opts.Author}}}
 	}
 	if inStrSlice(cmts.cells, opts.Comment.Cell, true) != -1 {
 		return newAddCommentError(opts.Comment.Cell)
 	}
-	if idx := inStrSlice(cmts.Authors.Author, opts.Author, true); idx == -1 {
+	authorID := inStrSlice(cmts.Authors.Author, opts.Author, true)
+	if authorID == -1 {
 		cmts.Authors.Author = append(cmts.Authors.Author, opts.Author)
 		authorID = len(cmts.Authors.Author) - 1
-	} else {
-		authorID = idx
 	}
 	defaultFont, err := f.GetDefaultFont()
 	if err != nil {
--- a/vml.go
+++ b/vml.go
@@ -282,9 +282,11 @@ func (f *File) addComment(commentsXML string, opts vmlOptions) error {
 	if inStrSlice(cmts.cells, opts.Comment.Cell, true) != -1 {
 		return newAddCommentError(opts.Comment.Cell)
 	}
-	if inStrSlice(cmts.Authors.Author, opts.Author, true) == -1 {
+	if idx := inStrSlice(cmts.Authors.Author, opts.Author, true); idx == -1 {
 		cmts.Authors.Author = append(cmts.Authors.Author, opts.Author)
 		authorID = len(cmts.Authors.Author) - 1
+	} else {
+		authorID = idx
 	}
 	defaultFont, err := f.GetDefaultFont()
 	if err != nil {
--- a/vml_test.go
+++ b/vml_test.go
@@ -83,6 +83,30 @@ func TestAddComment(t *testing.T) {
 	assert.EqualError(t, err, "sheet SheetN does not exist")
 }
 
+func TestAddCommentExistingAuthorID(t *testing.T) {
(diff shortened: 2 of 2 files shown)


xuri (approved):
Good catch! Thanks for your PR, I've made some changes based on your branch.

PR #2294, opened 2026-04-13
fix: prevent panic on add chart with no fill series

### LINKED ISSUE #2289: AddComment assigns authorID 0 for all comments when author already exists in authors list

lawrence3699:
# PR Details

## Description

When `AddComment` is called with an author that already exists in the authors list, `authorID` was left at its zero-value (`0`) instead of being set to the existing author's index. This caused all comments after the first for a given author to be misattributed to whichever author occupies index 0.

The fix captures the return value of `inStrSlice` and assigns it in the `else` branch so that existing authors are correctly resolved.

## Related Issue

Fixes #2289

## Motivation and Context

When multiple comments share the same author, only the first comment gets the correct author attribution. All subsequent comments by that author are incorrectly attributed to the author at index 0. This is because `authorID` defaults to `0` and is only assigned when a *new* author is appended — the existing-author path leaves it unchanged.

**Before (bug):**
```
Cell A1: Author="ExistingUser"  ← correct
Cell B1: Author="Agent"         ← correct (first comment by Agent)
Cell C1: Author="ExistingUser"  ← wrong, should be "Agent"
```

**After (fix):**
```
Cell A1: Author="ExistingUser"  ← correct
Cell B1: Author="Agent"         ← correct
Cell C1: Author="Agent"         ← correct
```

## How Has This Been Tested

- Added `TestAddCommentMultipleSameAuthor` covering the exact reproduction scenario
- All existing `TestAddComment` and `TestDeleteComment` tests pass
- `go vet ./...` clean

## Types of changes

## Checklist

### DIFF
--- a/drawing.go
+++ b/drawing.go
@@ -836,11 +836,12 @@ func (f *File) drawChartSeries(opts *Chart) *[]cSer {
 // drawShapeFill provides a function to draw the a:solidFill element by given
 // fill format sets.
 func (fill *Fill) drawShapeFill(spPr *cSpPr) *cSpPr {
-	if fill.Type == "pattern" && fill.Pattern == 1 {
-		if spPr == nil {
-			spPr = &cSpPr{}
-		}
-		if len(fill.Color) == 1 {
+	if spPr == nil {
+		spPr = &cSpPr{}
+	}
+	spPr.SolidFill = nil
+	if fill.Type == "pattern" {
+		if fill.Pattern == 1 && len(fill.Color) == 1 {
 			spPr.SolidFill = &aSolidFill{SrgbClr: &aSrgbClr{Val: stringPtr(strings.TrimPrefix(fill.Color[0], "#"))}}
 			if fill.Transparency > 0 {
 				val := (100 - fill.Transparency) * 1000
@@ -848,7 +849,6 @@ func (fill *Fill) drawShapeFill(spPr *cSpPr) *cSpPr {
 			}
 			return spPr
 		}
-		spPr.SolidFill = nil
 		spPr.NoFill = stringPtr("")
 	}
 	return spPr
@@ -878,7 +878,7 @@ func (f *File) drawChartSeriesSpPr(i int, opts *Chart) *cSpPr {
 	}[opts.Type]; ok {
 		return chartSeriesSpPr[opts.Series[i].Line.Type]
 	}
-	if spPr.SolidFill.SrgbClr != nil {
+	if spPr.SolidFill != nil && spPr.SolidFill.SrgbClr != nil || spPr.NoFill != nil {
 		return spPr
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


xuri (approved):
Thanks for your PR. The issue #2289 already fixed by PR #2291, I made some changes based on your branch for fix another problem related with https://github.com/xuri/excelize/discussions/1273.

PR #2300, opened 2026-04-20
Fix nil pointer panic in getOptions when File.options is nil and no opts are passed

### LINKED ISSUE #2299: (*Cols).Rows() panics with nil pointer dereference on malformed XLSX input
### Description

 When parsing a malformed (fuzz-generated) XLSX file, calling Cols() followed by Rows() causes a nil pointer dereference panic in (*File).getOptions(). The OpenReader call succeeds
  and returns a non-nil *File, and Cols() also succeeds and returns a non-nil *Cols. However, when iterating with Next() and then calling Rows(), the internal call to getOptions()
  panics because it dereferences a nil pointer.

  This is a crash on untrusted input — any application that parses user-supplied XLSX files (e.g. a document processing service) will crash without a recover() guard.

  Stack trace:
  panic: runtime error: invalid memory address or nil pointer dereference

  goroutine 14 [running]:
  github.com/xuri/excelize/v2.(*File).getOptions(...)
      excelize.go:238
  github.com/xuri/excelize/v2.(*Cols).Rows(0xc001320d30, {0x0?, 0xa?, 0x41ea8f?})
      col.go:94

 1. Save the following Go program as main.go:

  package main

  import (
      "fmt"
      "os"

      "github.com/xuri/excelize/v2"
  )

  func main() {
      f, err := excelize.OpenReader(os.Stdin)
      if err != nil {
          fmt.Println("OpenReader error:", err)
          return
      }
      defer f.Close()

      for _, sheet := range f.GetSheetList() {
          cols, err := f.Cols(sheet)
          if err != nil || cols == nil {
              continue
          }
          for cols.Next() {
              rows, err := cols.Rows() // panics here
              if err != nil {
                  fmt.Println("Rows error:", err)
                  continue
              }
              fmt.Println(rows)
          }
      }
  }

  2. Feed it the attached malformed XLSX file (found via Go's built-in fuzzer):
  go run main.go < fuzzInput.xlsx

  3. Observe the nil pointer dereference panic.

  Environment:
  - excelize version: v2.8.1 (also reproduced with v2.10.1)
  - Go version: go1.25.6 linux/amd64
  - OS: Ubuntu Linux (6.8.0-106-generic)

[fuzzInput.xlsx](https://github.com/user-attachments/files/26869988/fuzzInput.xlsx)

Parsing the file and not panic.

### Go version

go1.25.6 linux/amd64

 v2.8.1 (also reproduced with v2.10.1)

### Environment

```shell
- OS: Ubuntu Linux (6.8.0-106-generic)
```

### Validations

SAY-5:
## What

`getOptions` returned `File.options` verbatim when the caller didn't override with explicit opts. On the `OpenReader` path, `File.options` is populated via the same function:

```go
f := newFile()
f.options = f.getOptions(opts...)
```

So when `OpenReader` is invoked with no opts, `getOptions` sees the zero-value `f.options` (`nil`), returns `nil`, and `f.options` stays `nil` for the lifetime of the `*File`. Later readers — `(*Cols).Rows` at `col.go:94`:

```go
cols.rawCellValue = cols.f.getOptions(opts...).RawCellValue
```

then nil-deref:

```
panic: runtime error: invalid memory address or nil pointer dereference
github.com/xuri/excelize/v2.(*File).getOptions(...)
    excelize.go:238
github.com/xuri/excelize/v2.(*Cols).Rows(0xc001320d30, ...)
    col.go:94
```

This is reachable by feeding fuzz-generated XLSX data into `OpenReader` (the reporter in #2299 triggers it with Go's built-in fuzzer), which makes it a crash-on-untrusted-input for any service that parses user-supplied spreadsheets.

## Fix

Anchor `getOptions` to a fresh empty `*Options` when `f.options` is `nil` so its callers always get a safe pointer back. Explicit `opts` still override the resulting pointer as before:

```go
func (f *File) getOptions(opts ...Options) *Options {
    options := f.options
    if options == nil {
        options = &Options{}
    }
    for _, opt := range opts {
        options = &opt
    }
    return options
}
```

## Test plan

- Repro on `master` using the reporter's Go program and the attached `fuzzInput.xlsx` in #2299: `(*Cols).Rows` panics inside `getOptions`.
- Apply patch: same program returns the ORM-level error (`Rows error:`) for any invalid sheet instead of crashing, matching the documented "return an error" contract for `Rows()`. No explicit-opts path changes behaviour.

Fixes #2299

### DIFF
--- a/col.go
+++ b/col.go
@@ -214,8 +214,7 @@ func (f *File) Cols(sheet string) (*Cols, error) {
 		output, _ := xml.Marshal(ws)
 		f.saveFileList(name, f.replaceNameSpaceBytes(name, output))
 	}
-	var colIterator columnXMLIterator
-	colIterator.cols.sheetXML = f.readBytes(name)
+	colIterator := columnXMLIterator{cols: Cols{f: f, sheetXML: f.readBytes(name)}}
 	decoder := f.xmlNewDecoder(bytes.NewReader(colIterator.cols.sheetXML))
 	for {
 		token, _ := decoder.Token()
@@ -230,7 +229,6 @@ func (f *File) Cols(sheet string) (*Cols, error) {
 			}
 		case xml.EndElement:
 			if xmlElement.Name.Local == "sheetData" {
-				colIterator.cols.f = f
 				colIterator.cols.sheet = sheet
 				return &colIterator.cols, nil
 			}
--- a/col_test.go
+++ b/col_test.go
@@ -164,6 +164,38 @@ func TestGetColsError(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestColsIteratorRetainsFilePointer(t *testing.T) {
+	// Regression for #2299: when worksheet XML is truncated before the
+	// </sheetData> close element, the SAX loop in Cols falls through to
+	// the final return. Prior to #2300 the *File pointer was only wired
+	// into the iterator inside the sheetData EndElement branch, so this
+	// path returned a Cols whose f field was nil, and a subsequent
+	// Cols.Rows call would nil-dereference on cols.f.getOptions. Make
+	// sure the iterator carries a live *File regardless of where the
+	// parse exits.
+	f := NewFile()
+	f.Sheet.Delete("xl/worksheets/sheet1.xml")
+	// Missing </sheetData> and </worksheet> close tags simulate the
+	// truncation seen by the fuzzer in #2299.
+	f.Pkg.Store("xl/worksheets/sheet1.xml", []byte(fmt.Sprintf(
+		`<worksheet xmlns="%s"><sheetData><row r="1"><c r="A1" t="inlineStr"><is><t>v</t></is></c></row>`,
+		NameSpaceSpreadSheet.Value)))
+	f.checked = sync.Map{}
+
+	cols, err := f.Cols("Sheet1")
+	require.NoError(t, err)
+	require.NotNil(t, cols)
+	assert.NotNil(t, cols.f, "cols.f must be set even when sheetData close element is missing")
+
+	// Walking the iterator must not panic, even though the XML is truncated.
+	assert.NotPanics(t, func() {
+		for cols.Next() {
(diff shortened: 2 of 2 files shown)


excelize.go
```diff
@@ -229,7 +229,15 @@ func OpenReader(r io.Reader, opts ...Options) (*File, error) {
 // getOptions provides a function to parse the optional settings for open
 // and reading spreadsheet.
 func (f *File) getOptions(opts ...Options) *Options {
+	// options can be nil when newFile() has just run and no caller has yet
```
xuri: The root cause is corrupted workbook contains invalid worksheet XML, the SAX reader can not get `sheetData` close element, file pointer not assigned to `colIterator.cols.f`. Fix like this would be better approach (the Row's iterator also should be updated in same way):

```diff
diff --git a/col.go b/col.go
index 9c854bc..ee7f6e6 100644
--- a/col.go
+++ b/col.go
@@ -214,8 +214,7 @@ func (f *File) Cols(sheet string) (*Cols, error) {
 		output, _ := xml.Marshal(ws)
 		f.saveFileList(name, f.replaceNameSpaceBytes(name, output))
 	}
-	var colIterator columnXMLIterator
-	colIterator.cols.sheetXML = f.readBytes(name)
+	colIterator := columnXMLIterator{cols: Cols{f: f, sheetXML: f.readBytes(name)}}
 	decoder := f.xmlNewDecoder(bytes.NewReader(colIterator.cols.sheetXML))
 	for {
 		token, _ := decoder.Token()
@@ -230,7 +229,6 @@ func (f *File) Cols(sheet string) (*Cols, error) {
 			}
 		case xml.EndElement:
 			if xmlElement.Name.Local == "sheetData" {
-				colIterator.cols.f = f
 				colIterator.cols.sheet = sheet
 				return &colIterator.cols, nil
 			}
```

xuri (changes requested):

xuri (approved):
I simplify unit tests code based on your branch. Thanks for your contribution!

PR #2301, opened 2026-04-20
Fix external formula references during adjustment

rootsec1:
## What Changed

Formula adjustment now distinguishes external workbook sheet references from structured references that also contain square brackets. Structured references remain skipped, while external workbook references go through the normal adjustment path so required single quotes are restored after formula tokenization.

## Reproduction

A synthetic workbook can reproduce this:

1. Set `B1` to `SUMIFS('[1]External Data'!$L:$L,'[1]External Data'!$B:$B,A1)`.
2. Call `InsertRows("Sheet1", 1, 1)`.
3. Before this fix, the formula became `SUMIFS([1]External Data!$L:$L,[1]External Data!$B:$B,A2)`, which is invalid because the external sheet name with a space is no longer quoted.

The regression test uses only this synthetic formula and also saves and reopens the workbook.

## Validation

- `go test ./... -run TestAdjustFormula -count=1`
- `go test ./... -count=1`

### DIFF
--- a/adjust.go
+++ b/adjust.go
@@ -414,6 +414,17 @@ func (f *File) adjustFormulaOperand(sheet, sheetN string, keepRelative bool, tok
 	return operand, err
 }
 
+// isExternalSheetReference returns true for workbook-qualified sheet
+// references such as [Book1.xlsx]Sheet1!A1.
+func isExternalSheetReference(ref string) bool {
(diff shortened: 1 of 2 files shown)


xuri (approved):
Thanks for your PR! I've made some changes based on your branch.

PR #2304, opened 2026-04-20
This closes #2303, support 3D references across sheet ranges

krystophny:
## Description

`SUM(Sheet1:Sheet3!A1)` and other 3D aggregates previously failed with
`#NAME? invalid reference`: `parseReference` split on `:` and handed the
leading sheet token to `parseRef`, which treats it as a cell name.

This change teaches `parseReference` to recognise references of the
shape `<sheet1>:<sheet2>!<ref>`, expand them across the workbook-order
sheet range, and evaluate each sheet's cell or range recursively. The
resulting matrix is consumed unchanged by the existing aggregates
(`SUM`, `AVERAGE`, `MIN`, `MAX`, `COUNT`, `COUNTA`, `PRODUCT`). Non-ASCII
unquoted sheet names work too, e.g. `SUM(Jänner:Dezember!$M$40)`.

Out of scope: the quoted form `SUM('A':'B'!ref)` is mis-tokenised by
`github.com/xuri/efp` before reaching `parseReference` and still
evaluates only the last sheet. That fix belongs in efp and will be
filed there. `TestCalc3DRefQuotedKnownLimitation` pins the current
behaviour with `t.Skip` so a future efp fix lands here visibly.

## Related Issue

Closes #2303.

## Motivation and Context

Workbooks with 3D aggregate totals (annual summaries across monthly
sheets) currently can't be recalculated through excelize — every such
formula resolves to `#NAME?`. This is the narrowest fix that removes
that failure without touching other code paths.

## How Has This Been Tested

- `TestCalc3DRef` (table-driven) covers single-cell, range, degenerate
  single-sheet, multi-column, reversed range (`#REF!`), and missing
  sheet cases across all seven aggregates listed above.
- `TestCalc3DRefSplit` exercises the shape parser on malformed inputs
  (missing `!`, trailing `:`, empty sheet token, three-way ranges) to
  confirm they fall through to the existing 2D path.
- `TestCalc3DRef/non-ascii_sheet_names` covers `Jänner:Dezember`.
- `TestCalc3DRefQuotedKnownLimitation` is `t.Skip`'d as documented above.
- `go test -timeout 10m .` passes (`ok ... 13.960s`); `gofmt -s` and
  `go vet ./...` clean; ARM cross-builds from `.github/workflows/go.yml`
  (`GOARM=5|6|7`, `arm64`, `android/arm64`) all build.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
[...]

PR #2305, opened 2026-04-20
This closes #2303, add File.Recalc to evaluate and persist formula results

### LINKED ISSUE #2303: Add (*File).Recalc to evaluate formulas and persist cached values
## Problem

excelize has no "open → recalc → save" workflow. A consumer that wants it hits three walls:

1. **No cached-value writer.** `SetCellFloat` / `SetCellStr` / etc. call `removeFormula`, which deletes `<f>` and wipes shared-formula children.
2. **Round-tripping.** `CalcCellValue` returns a `string`; writing it back via `SetCellValue` records `t="s"`. Aggregates over that shared-string blob return 0.
3. **3D references:**
   - `SUM(Sheet1:Sheet3!A1)` unquoted → `#NAME? invalid reference`.
   - `SUM('Sheet1':'Sheet3'!A1)` quoted → parses, but `rangeResolver` overwrites `sheet` per iteration and evaluates only the last sheet.

## Proposal

Add two exported methods:

```go
func (f *File) RecalcCell(sheet, cell string) error
func (f *File) Recalc(opts ...RecalcOptions) error
```

`RecalcCell` evaluates one formula cell and writes the typed result into `<v>`/`<t>`, preserving `<f>` and shared-formula grouping. `Recalc` walks formula cells and calls `RecalcCell` on each. Dependency resolution reuses the existing recursive `calcCellValue` evaluator; circular references are bounded by `MaxCalcIterations`.

## Scope

Four narrow PRs split under this umbrella, each with its own tracking issue where appropriate:

1. **3D references, unquoted form.** `parseReference` recognises `<sheetA>:<sheetB>!<ref>` and expands in workbook order; aggregates consume the expanded matrix. Covers `SUM(Jänner:Dezember!M40)` etc. Closes this issue (partial).
2. **`RecalcCell` + private `setCellCachedValue` helper + `ErrCellNoFormula` sentinel.** Tracked in #2307.
3. **`File.Recalc` whole-workbook sweep built on `RecalcCell`, aggregating failures via `errors.Join`.** Closes this issue.
4. **`RecalcOptions{Sheet, Ref}` scoping for `File.Recalc`.** Tracked in #2308.

## Out of scope (tracked elsewhere)

- Quoted 3D references (`SUM('A':'B'!ref)`) — mis-tokenised by `github.com/xuri/efp` upstream of `parseReference`. Tracked in #2309.

## Not needed (verified)

- `_xlfn.` prefix stripping already works (`calc.go:210`).
- `ISOWEEKNUM` and `SHEET` are registered and correct.

krystophny:
## Description

Add `(*File).Recalc(opts ...RecalcOptions) error`, the second half of the
umbrella tracked in #2303. `Recalc` walks every formula cell in scope,
evaluates it with the existing recursive engine, and persists the result
into the cell's cached `<v>` / `<t>` without disturbing `<f>`, the
shared-formula master, `ref`, or `si`. A private `setCellCachedValue`
helper does the XML-level write; it is the counterpart to the public
`SetCell*` writers, which all strip the formula on write and are
therefore unusable for a recalc pipeline.

Public surface introduced by this PR:

- `(*File).Recalc(opts ...RecalcOptions) error`
- `RecalcOptions{Sheet, Ref}` for workbook / sheet / range scoping
- `*RecalcError` aggregating per-cell failures, with `Unwrap() []error`
- `RecalcCellError{Sheet, Cell, Err}` implementing `error` and `Unwrap()`
  so `errors.As` can target a specific address
- `ErrCellNoFormula` sentinel returned when the target cell has no `<f>`
  element

No existing API is changed or removed.

## Related Issue

Umbrella: #2303.

This is the second of two PRs under that issue. The first PR, #2304
(3D references across sheet ranges), is still open in draft. This PR's
branch is stacked on top of the #2304 branch so CI runs against the
combined state; `TestRecalcAcross3DRefs` exercises the 3D-ref path and
relies on #2304 being merged first (or landing together).

## Motivation and Context

Users who open a workbook, mutate inputs, and save it back currently
have no way to refresh the cached values that downstream consumers
rely on. `CalcCellValue` returns a formatted string but does not
persist it, and every `SetCell*` writer removes the formula while
writing a new value, which makes them useless for this purpose. The
result is that LibreOffice-headless, Numbers, and other readers that
trust `<v>` without recomputing see stale results after a round trip
through excelize. `Recalc` closes that gap with a single entry point
and zero cost to callers who do not use it.

Design notes reviewers are likely to ask about:

- **Fixed-point iteration.** Not needed. `calcCellValue` already
  recurses into referenced cells during evaluation, so a single pass
  converges chained dependencies. The implementation clears the calc
  cache before and after the pass to avoid cross-call contamination.
- **Circular references.** Bounded by `f.options.MaxCalcIterations`,
  the same option that governs the existing calc engine. No new knob.
  `TestRecalcCircularReferenceBounded` locks termination.
- **Shared-formula groups.** `setCellCachedValue` touches only `<v>`
  and `<t>`; `<f>` with its `t="shared"`, `ref`, and `si` attributes is
  left untouched, and sibling children in the same band are not
  modified by the master's rewrite.
  `TestRecalcSharedFormulaBandUntouched` locks this.
- **Error model.** Failures never abort the pass. Each failing cell
  contributes a `RecalcCellError`; cells that did compute are still
  written. The aggregate `*RecalcError` implements `Unwrap() []error`
  so `errors.Is` / `errors.As` descend naturally.
- **Scope.** `RecalcOptions{}` recalculates the whole workbook.
  `Sheet` narrows to one sheet; `Ref` (A1 range) narrows further and
  requires `Sheet`. `TestRecalcRejectsRefWithoutSheet` locks the
  contract.

## How Has This Been Tested

New tests covering the private helper (`cell_cached_test.go`, 8 tests)
and the public entry point (`recalc_test.go`, 17 tests) across:

- Simple sums, chained dependencies, round-trip through save/reopen
- 3D references (`TestRecalcAcross3DRefs`)
- Shared-formula row 40 band (master + children untouched)
- Idempotency (second Recalc is a no-op on value)
- `t="s"` regression guard (numeric result must not store as shared
  string)
- Sheet and range scope, rejection of `Ref` without `Sheet`
- Boolean and error result types
- Circular reference termination
- Non-ASCII sheet names

Full suite:

```
$ go test -skip TestZip64 -timeout 10m .
ok  	github.com/xuri/excelize/v2	13.97s
```

Static checks clean: `gofmt -s -l .`, `go vet ./...`, and ARM cross-builds
(`GOARM=5|6|7`, `arm64`, `android/arm64`) from
`.github/workflows/go.yml`.

## Types of changes

## Checklist

Documentation for the new exported surface lives in the doc comments on
`Recalc`, `RecalcOptions`, `RecalcError`, `RecalcCellError`, and
`ErrCellNoFormula`.


krystophny:
Superseded. The ~1000-line original is being re-filed as three independently reviewable PRs so each does one thing:

- #2307 — `RecalcCell` + private `setCellCachedValue` helper + `ErrCellNoFormula` sentinel (~255 LoC)
- #2303 — `File.Recalc` whole-workbook sweep built on `RecalcCell`, aggregating failures via `errors.Join` (~120 LoC)
- #2308 — `RecalcOptions{Sheet, Ref}` scoping (~100 LoC)

Dropped in the restructure: typed `RecalcError` / `RecalcCellError` (replaced by `errors.Join` so `errors.Is`/`errors.As` still descend); `TestRecalcAcross3DRefs` (removed the stacking dependency on the 3D-refs PR).

PR #2306, opened 2026-04-21
fix setCellStyle fill with pattern 1 without color

noahchiu22:
# Summary
- Fix `setCellStyle` fill handling when `pattern = 1` and no fill color is provided.
- Prevent unintended black `bgColor` output for non-default patterns that omit color.
- Keep color inheritance behavior only for the default pattern, so format color is not incorrectly overridden.
# Problem
When a style uses fill `pattern = 1` without color, Excel may render a black background due to generated `bgColor` behavior.
This causes unexpected formatting output for pattern-based fills that do not explicitly set a color.

# Root Cause
`newFills` created `patternFill` even when no color existed, regardless of pattern type.
For non-default patterns, this can trigger fallback rendering that appears as black background.

# Fix
Update `newFills` in `styles.go` so that when `len(style.Fill.Color) < 1`, `patternFill` is only emitted for the default pattern (`pattern == 0`).
This avoids forcing background color behavior for non-default patterns with no color.

# Impact
- Preserves expected formatting for styles using pattern fills without explicit color.
- Avoids accidental black background output.
- Keeps existing behavior for default pattern fill.
# Test Plan
- Create style with `Fill.Pattern = 1` and empty `Fill.Color`; apply via `SetCellStyle`; verify no unexpected black background.
- Verify default fill pattern path still behaves as before.
- Run style-related tests (especially fill/style serialization paths).

### DIFF
--- a/styles.go
+++ b/styles.go
@@ -2086,7 +2086,9 @@ func newFills(style *Style, fg bool) *xlsxFill {
 		var pattern xlsxPatternFill
 		pattern.PatternType = styleFillPatterns[style.Fill.Pattern]
 		if len(style.Fill.Color) < 1 {
-			fill.PatternFill = &pattern
+			if style.Fill.Pattern == 0 {
+				fill.PatternFill = &pattern
+			}
 			break
 		}
 		if fg {
--- a/styles.go
+++ b/styles.go
@@ -1523,33 +1523,45 @@ func (f *File) extractFills(fl *xlsxFill, s *xlsxStyleSheet, style *Style) {
 	if fl != nil {
 		var fill Fill
 		if fl.GradientFill != nil {
-			fill.Type = "gradient"
-			for shading, variants := range styleFillVariants() {
-				if fl.GradientFill.Bottom == variants.Bottom &&
-					fl.GradientFill.Degree == variants.Degree &&
-					fl.GradientFill.Left == variants.Left &&
-					fl.GradientFill.Right == variants.Right &&
-					fl.GradientFill.Top == variants.Top &&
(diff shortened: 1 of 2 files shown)


styles.go
```diff
@@ -2086,7 +2086,9 @@ func newFills(style *Style, fg bool) *xlsxFill {
 		var pattern xlsxPatternFill
 		pattern.PatternType = styleFillPatterns[style.Fill.Pattern]
 		if len(style.Fill.Color) < 1 {
-			fill.PatternFill = &pattern
+			if style.Fill.Pattern == 0 {
```
xuri: This changes will cause custom pattern setting not work if `Pattern` value not 0. For example, the following style with "50% Gray" (`mediumGray`) pattern style  will not work:

```go
    Fill: excelize.Fill{Pattern: 2, Type: "pattern"},
})
```

xuri (changes requested):

xuri (approved):
I made some changes based on your branch, set auto foreground and background color when creating solid fill style when no custom color value specified would be better approach.

noahchiu22:
@xuri Sorry—I’m wondering when this change can be released, or whether my project can’t use the remote Excelize package right now.

xuri:
Thanks for your contribution. The next version 2.11.0 will be released on fall of this year. Please use the Go module to lock the version before the new version is released. Please upgrade your Excelize module to the master branch to apply this patch by following command: `go get -u github.com/xuri/excelize/v2@master`.

noahchiu22:
@xuri No problem, Thanks

PR #2320, opened 2026-05-12
Optimize the coordinates conversion and some internal functions performance

AdamDrewsTR:
## Summary

This PR optimizes several frequently-called functions to reduce heap allocations and improve throughput for large spreadsheet operations. Each change targets a specific hot path identified through profiling.

### `ColumnNumberToName` — O(1) lookup table (~16KB init cost)

Precomputes all 16,384 valid column names at package init into a flat `[]string` slice. Subsequent calls become a bounds check + slice index — zero allocations.

**Before:** Each call allocated a `[]byte` and computed the column name via division loop.
**After:** Single table lookup.

### `CoordinatesToCellName` — avoid empty-prefix concatenation

When `abs` is not set (the common case), the old code concatenated `"" + colName + "" + rowStr`, producing unnecessary string copies. The new code returns `colName + strconv.Itoa(row)` directly and early-returns for the absolute case.

### `namespaceStrictToTransitional` — fast path for Transitional files

The vast majority of XLSX files use the Transitional namespace. All Strict namespace URIs contain `"purl.oclc.org"`, so a single `bytes.Contains` check can skip the entire replacement loop and avoid allocating a copy of the sheet XML. This applies to &gt;99% of real-world files.

### `isNumeric` — replace `math/big.Float` with `strconv.ParseFloat`

The `big.Float` parser allocates significantly more than `strconv.ParseFloat` for the same input. This also removes the `math/big` import entirely. The digit-counting step uses `strings.Count` instead of `strings.ReplaceAll` to avoid allocating a modified copy of the string.

### `bstrUnmarshal` — skip regex when no escape sequences present

Over 99% of cell values contain no `_x` escape sequences. A simple `strings.Contains(s, "_x")` guard skips the regex entirely for these cells. In profiling, this eliminated ~54 MB of regex-related allocations per 100K rows.

### `workSheetReader` — avoid double-reading sheet data

The original code called `f.readBytes(name)` twice — once for `getRootElement` and once for `Decode`. Caching the result in a local variable halves the I/O for worksheets read from temp files.

## Benchmark Impact

These are foundational optimizations — the impact compounds with sheet size since `ColumnNumberToName`, `CoordinatesToCellName`, `bstrUnmarshal`, and `namespaceStrictToTransitional` are called per-cell or per-sheet. Individually each saves microseconds; collectively they reduce GC pressure significantly for large workbooks.

### DIFF
--- a/chart.go
+++ b/chart.go
@@ -956,7 +956,7 @@ func (opts *Chart) parseTitle() error {
 // Set chart offset, scale, aspect ratio setting and print settings by 'Format',
 // same as function 'AddPicture'.
 //
-// Set the position of the chart plot area by 'PlotArea' field with
+// Set the properties of the chart plot area by 'PlotArea' field with
 // 'ChartPlotArea' data type. The properties that can be set are:
 //
 //	SecondPlotValues
--- a/excelize.go
+++ b/excelize.go
@@ -311,15 +311,16 @@ func (f *File) workSheetReader(sheet string) (ws *xlsxWorksheet, err error) {
 		}
 	}
 	ws = new(xlsxWorksheet)
+	data := f.readBytes(name)
 	if attrs, ok := f.xmlAttr.Load(name); !ok {
-		d := f.xmlNewDecoder(bytes.NewReader(namespaceStrictToTransitional(f.readBytes(name))))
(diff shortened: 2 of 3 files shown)


xuri (approved):
Thanks for your contribution. I've made some update based on your branch.

PR #2322, opened 2026-05-12
Optimize StreamWriter cell writing to reduce per-cell allocations (B2)

AdamDrewsTR:
## Summary

Optimizes the `StreamWriter.SetRow` cell writing path to eliminate per-cell heap allocations for the most common value types (numbers and plain strings). This builds on `perf/buffered-writer` (#2321).

**Depends on:** #2321 (buffered-writer enhancements)

### Fast path: numeric cells (`writeNumericCell`)

For `int`, `int8`–`int64`, `uint`–`uint64`, `float32`, `float64`, and `bool` values, writes the complete `<c r="..." s="..."><v>...</v></c>` element directly to the buffer using `WriteInt`/`WriteFloat` from the enhanced `bufferedWriter`. This bypasses:
- `xlsxC.V` string allocation (`strconv.FormatInt` → string → `xml.EscapeText` → `[]byte`)
- `CoordinatesToCellName` per-cell call
- `prepareCellStyle` per-cell scan

### Fast path: string cells (`writeStringCell`)

For plain `string` and `[]byte` values, writes inline-string XML directly:
```xml
<c r="A1" t="inlineStr"><is><t>value</t></is></c>
```
This bypasses `xlsxC`, `xlsxSI`, `trimCellValue`, `bstrMarshal`, and `xml.EscapeText` — eliminating ~6 heap allocations per string cell. Falls through to the slow path for strings containing bstr escape sequences (`_xHHHH_`) or exceeding `TotalCellChars`.

### `writeEscaped` — allocation-free XML escaping

Scans for the 5 XML special characters (`<`, `>`, `&`, `"`, `\r`) and writes escape sequences inline. For strings with no special characters (the common case), this is a single `WriteString` call with zero allocations, vs `xml.EscapeText` which always allocates a `[]byte` copy.

### `streamCellStyle` + `colStyles` cache

Builds a `[]int` array mapping column index → style ID when `writeSheetData` processes the `<cols>` element. Per-cell style resolution becomes an array index instead of scanning `worksheet.Cols.Col` entries.

### `marshalAttrs` → direct buffer write

Split the old `marshalAttrs() (strings.Builder, error)` into `validateRowOpts() error` (called once) and `marshalAttrs(w *bufferedWriter)` (writes directly). Eliminates the intermediate `strings.Builder` and its `.String()` copy.

### `setCellValFunc` expansion

Each integer size (`int8`, `int16`, etc.) gets its own case with direct `strconv.FormatInt`/`FormatUint`, avoiding the extra interface boxing through `setCellIntFunc`.

### `SetRow` restructuring

- Precomputes `rowStr` once per row
- Uses `ColumnNumberToName` directly instead of `CoordinatesToCellName` (avoids redundant row formatting)
- Reuses a single `xlsxC` struct for the slow path instead of allocating one per cell

## Compatibility

All existing tests pass. The output XML is byte-identical for all cell types — the optimizations only change how the XML is constructed, not what is produced.

### DIFF
--- a/excelize.go
+++ b/excelize.go
@@ -122,6 +122,20 @@ type Options struct {
 	LongDatePattern   string
 	LongTimePattern   string
 	CultureInfo       CultureName
+	// StreamingChunkSize is the number of bytes of XML data accumulated in
+	// memory before a streaming worksheet spills to a temp file. A smaller
+	// value reduces peak memory usage at the cost of more disk I/O. Zero
+	// means use the default (StreamChunkSize = 16 MiB). Set to -1 to
+	// disable temp files entirely (all data stays in memory); this
+	// eliminates disk I/O overhead and can be significantly faster when
+	// sufficient memory is available.
+	StreamingChunkSize int
+	// StreamingBufSize is the size of the bufio.Writer used for all disk
+	// writes after the StreamingChunkSize threshold is crossed. Larger values
+	// reduce write syscall counts at the cost of slightly more memory. The
+	// measured inflection point on NVMe and HDD alike is 128 KiB. Zero means
+	// use the default (StreamingBufSizeDefault = 128 KiB).
+	StreamingBufSize int
 }
 
 // OpenFile take the name of a spreadsheet file and returns a populated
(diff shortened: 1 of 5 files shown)


AdamDrewsTR:

PR #2323, opened 2026-05-12
Add dimension placeholder for accurate streaming worksheet dimensions (B3)

AdamDrewsTR:
## Summary

Adds a dimension placeholder mechanism to `StreamWriter` so that the `<dimension>` element in streaming worksheets accurately reflects the data range, without requiring a full file rewrite.

**Depends on:** #2321 (buffered-writer enhancements — for `WriteAt` and `written` tracking)

## Problem

The current `StreamWriter` writes `bulkAppendFields(&sw.rawData, sw.worksheet, 3, 4)` which includes the dimension element from the (empty) worksheet template. This means streaming worksheets always have an empty or incorrect dimension. While Excel recalculates dimensions on open, other XLSX readers (Google Sheets, LibreOffice, programmatic parsers) may rely on the `<dimension>` element for range detection.

## Solution

1. **Placeholder**: A fixed-width string `<dimension ref="A1:XFD1048576"/>` (32 bytes) is written at a known byte offset during `NewStreamWriter`, after the `<sheetPr>` element but before `<sheetViews>`.

2. **Tracking**: `SetRow` tracks `maxCol` — the maximum column index seen across all rows.

3. **Overwrite**: In `Flush`, after all XML is written and the buffer/file is complete, `updateDimension()` computes the actual dimension (e.g. `A1:D100`) and uses `WriteAt` to overwrite the placeholder in-place. The replacement is right-padded with spaces to match the placeholder length exactly, since XML ignores trailing whitespace in element content.

## Why fixed-width?

The `<dimension>` element must appear before `<sheetData>` in the XLSX schema, but we don't know the final data range until all rows are written. A fixed-width placeholder avoids shifting all subsequent byte offsets — `WriteAt` simply overwrites the exact bytes without moving anything.

The placeholder uses the maximum possible cell reference (`XFD1048576`) so any real dimension string is guaranteed to be shorter or equal in length.

## Compatibility

All existing tests pass. The dimension element is optional per ECMA-376 §18.3.1.35 — Excel recalculates it on open regardless. This change improves interoperability with readers that do honor the element.

### DIFF
--- a/excelize.go
+++ b/excelize.go
@@ -122,6 +122,20 @@ type Options struct {
 	LongDatePattern   string
 	LongTimePattern   string
 	CultureInfo       CultureName
+	// StreamingChunkSize is the number of bytes of XML data accumulated in
+	// memory before a streaming worksheet spills to a temp file. A smaller
+	// value reduces peak memory usage at the cost of more disk I/O. Zero
+	// means use the default (StreamChunkSize = 16 MiB). Set to -1 to
+	// disable temp files entirely (all data stays in memory); this
+	// eliminates disk I/O overhead and can be significantly faster when
+	// sufficient memory is available.
+	StreamingChunkSize int
+	// StreamingBufSize is the size of the bufio.Writer used for all disk
+	// writes after the StreamingChunkSize threshold is crossed. Larger values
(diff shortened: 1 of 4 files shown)


AdamDrewsTR:

PR #2324, opened 2026-05-12
Stream WriteTo directly to writer and fix ZIP64 local file header sizes (B4)

AdamDrewsTR:
## Summary

Two related changes: (1) stream `WriteTo` directly instead of buffering, and (2) fix ZIP64 Local File Header size fields for entries >4GB.

**Depends on:** #2321 (buffered-writer enhancements — for `CopyTo`)

### Problem

Go's `archive/zip` writes the Central Directory entry with `ReaderVersion=45` for entries >4GB, but the Local File Header retains `version=20` and the original (truncated) 32-bit size values. Excel strictly validates that:

1. LFH version matches CD version (45 for ZIP64) ✅ already fixed by `writeZip64LFH`
2. LFH compressed and uncompressed sizes are `0xFFFFFFFF` when ZIP64 extended info is present ❌ **not fixed until now**

Without setting sizes to `0xFFFFFFFF`, Excel sees a mismatch between the LFH size fields and the ZIP64 extra field, and reports the workbook as corrupt.

### Fix

After patching the version to 45, also set the compressed size (offset +18) and uncompressed size (offset +22) to `0xFFFFFFFF` in the LFH for all ZIP64 entries. This tells readers to use the actual sizes from the ZIP64 extended information extra field that follows the filename in the LFH.

### Problem

The current `WriteTo` calls `WriteToBuffer()`, which materializes the entire compressed ZIP archive in a `bytes.Buffer`. For large workbooks (e.g., a 4GB+ sheet compresses to 50-200 MB), this means 50-200 MB of peak memory just for the output buffer, on top of the input data.

### Solution

`WriteTo` now streams the ZIP directly to the destination `io.Writer` using a `countWriter` wrapper (to track `int64` bytes written). No intermediate `bytes.Buffer` is needed.

For encrypted files (which require post-processing the entire ZIP), a temporary file is used instead of an in-memory buffer, accessed via the new `writeToWithEncryption` helper.

### Additional changes

- `writeToZip` uses `CopyTo` instead of `Reader` + `io.Copy` for streaming worksheets, leveraging the larger read buffers from #2321
- `writeZip64LFH` is only called when `zip64Entries` is non-empty (avoids scanning the entire output for files with no >4GB entries)
- Added `writeZip64LFHFile` for file-based ZIP64 fixup (used by the encryption path)

## Compatibility

All existing tests pass including `TestZip64`.

### DIFF
--- a/excelize.go
+++ b/excelize.go
@@ -122,6 +122,20 @@ type Options struct {
 	LongDatePattern   string
 	LongTimePattern   string
 	CultureInfo       CultureName
+	// StreamingChunkSize is the number of bytes of XML data accumulated in
+	// memory before a streaming worksheet spills to a temp file. A smaller
+	// value reduces peak memory usage at the cost of more disk I/O. Zero
+	// means use the default (StreamChunkSize = 16 MiB). Set to -1 to
+	// disable temp files entirely (all data stays in memory); this
+	// eliminates disk I/O overhead and can be significantly faster when
+	// sufficient memory is available.
+	StreamingChunkSize int
+	// StreamingBufSize is the size of the bufio.Writer used for all disk
+	// writes after the StreamingChunkSize threshold is crossed. Larger values
+	// reduce write syscall counts at the cost of slightly more memory. The
+	// measured inflection point on NVMe and HDD alike is 128 KiB. Zero means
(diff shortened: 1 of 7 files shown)


AdamDrewsTR:

PR #2329, opened 2026-05-14
Fix agile encrypted package chunk boundary

### LINKED ISSUE #2328: Agile-encrypted XLSX decrypts incorrectly when EncryptedPackage payload length is a multiple of 4096
## Description

`excelize.OpenFile` fails to open a valid Agile-encrypted XLSX with the correct password when the `EncryptedPackage` encrypted payload length is exactly a multiple of 4096.

The same file can be decrypted successfully by `msoffcrypto-tool`, and the decrypted output is a valid XLSX ZIP package. After decrypting the file with `excelize.Decrypt`, the output starts with a valid ZIP local header, but the ZIP EOCD tail is corrupted, so `zip.NewReader` fails.

## Versions Tested

- `github.com/xuri/excelize/v2 v2.8.1`
- `github.com/xuri/excelize/v2 v2.9.1`
- `github.com/xuri/excelize/v2 v2.10.1`

All tested versions fail with:

```text
the supplied open workbook password is not correct
```

## Reproduction Characteristics

For the failing file:

```text
EncryptedPackage stream size: 12296
Plain package size stored in first 8 bytes: 12284
Encrypted payload size: 12288
Encrypted payload size % 4096 == 0
```

`excelize.Decrypt` returns `err == nil`, but the decrypted bytes are not a valid ZIP:

```text
decrypt err=<nil>
decrypted output len=12288
first bytes=504b0304...
zip.NewReader error: zip: not a valid zip file
```

Comparing with `msoffcrypto-tool` output:

```text
msoffcrypto decrypted len=12284
excelize decrypted len=12288
first differing offset=12272
EOCD offset is the same, but the EOCD tail is corrupted in excelize output
```

## Suspected Root Cause

The issue appears to be in `decryptPackage` in `crypt.go`:

```go
if (end + offset) < len(input) {
    inputChunk = input[start+offset : end+offset]
} else {
    inputChunk = input[start+offset : end]
}
```

When `end + offset == len(input)`, this falls into the `else` branch and slices:

```go
input[start+offset : end]
```

This drops the last 8 bytes of the encrypted payload for the final chunk. The code then pads the shortened chunk before AES decryption, so the last AES block is decrypted incorrectly.

This corrupts the ZIP EOCD tail and causes `zip.NewReader` to fail. `OpenReader` then reports `ErrWorkbookPassword`, which is misleading because the password is correct.

## Expected Behavior

`excelize.OpenFile(path, excelize.Options{Password: password})` should open the file successfully when the password is correct.

`excelize.Decrypt` should return the exact plaintext package length stored in the first 8 bytes of `EncryptedPackage`.

## Suggested Fix Direction

`decryptPackage` should probably treat `input[8:]` as the encrypted payload and chunk that payload independently. It should also truncate the decrypted output to the package size stored in the first 8 bytes of `EncryptedPackage`.

At minimum, the boundary condition should not drop bytes when `end + offset == len(input)`.

WaterRRabbit:
## Summary

- Fix Agile encrypted package chunk slicing when the encrypted payload length is a multiple of 4096 bytes.
- Decrypt chunks from the encrypted payload after the 8-byte `EncryptedPackage` size header, and trim output to the plaintext package size.
- Add a synthetic non-confidential Agile-encrypted XLSX regression fixture for issue #2328.

Fixes #2328.

## Test Plan

- `GOSUMDB=sum.golang.org GOPROXY=https://proxy.golang.org,direct go test -run TestAgileDecryptPackageBoundary`
- `GOSUMDB=sum.golang.org GOPROXY=https://proxy.golang.org,direct go test -run TestEncrypt`
- `GOSUMDB=sum.golang.org GOPROXY=https://proxy.golang.org,direct go test ./...`


xuri (approved):

PR #2331, opened 2026-05-18
This fixes security advisories GHSA-q5j5-6p94-4gwc and GHSA-fx5j-qcqg-grpf

### LINKED ISSUE #2121: Use mmap if possible (can save memory with big files)

artur-chopikian:
## Description

Replace `os.Open` with `golang.org/x/exp/mmap.Open` in `OpenFile`. The returned `*mmap.ReaderAt` implements `io.ReaderAt`, so it is passed unchanged to the existing `openReaderAt` path — no downstream changes are required, and the public `OpenFile` signature and behaviour stay the same.

With memory mapping, the operating system page-maps the `.xlsx` file into the process address space on demand. `archive/zip` then serves section reads directly from the kernel page cache instead of copying through buffered `read` syscalls into Go heap buffers. The mapping is released by `r.Close()` before `OpenFile` returns, matching the lifetime of the previous `*os.File`.

Changes:
- `excelize.go`: use `mmap.Open` and `*mmap.ReaderAt` in `OpenFile`; close the mapping on success and on error from `openReaderAt`.
- `go.mod` / `go.sum`: add `golang.org/x/exp` for the `mmap` subpackage.

## Related Issue

Closes #2121 — *Use mmap if possible (can save memory with big files)*.

## Motivation and Context

`OpenFile` is the dominant entry point for reading spreadsheets from disk. Even after recent work to read the unencrypted workbook lazily through `io.ReaderAt`, every section access still went through a buffered `read` into a Go heap buffer, so opening a large `.xlsx` produced RSS growth and GC churn proportional to the file size touched by `archive/zip`.

Memory mapping avoids both costs: the kernel page cache backs the mapping, reads are zero-copy, pages are reclaimable under memory pressure, and unrelated processes opening the same file share physical memory. The behaviour requested in #2121 is achieved without changing any caller code.

## How Has This Been Tested

- `go build ./...` — clean.
- `go test -run 'TestOpenFile|TestOpenReader' -count=1 ./...` — pass.
- `go test -count=1 ./...` — full suite passes (`ok github.com/xuri/excelize/v2 37.000s`).

No new tests are added because the change is a transparent substitution behind an existing `io.ReaderAt`: the entire existing `OpenFile` test coverage exercises the mmap-backed path unchanged.

## Types of changes

## Checklist


xuri (commented):
Thanks for your PR. Any benchmark data on the performance impact of using `golang.org/x/exp/mmap` instead of `os`? Specifically, how much memory is saved, and any speed impact? I suggest test with [excelize-benchmark](https://github.com/xuri/excelize-benchmark) for performance related changes.

artur-chopikian:
@xuri, I checked the performance with [excelize-benchmark](https://github.com/xuri/excelize-benchmark) and I do not see any visible improvements for current files and event bigger ones. So we can close it for now :)

xuri (approved):
Thanks for your feedback. I've reverted changes of use mmap to open spreadsheet file, and made some changes based on your branch to fix two security advisories.

PR #2332, opened 2026-05-27
feat: use embed for template

kenny-not-dead:
# PR Details

use package embed for template

## Description

convenient way to interact with templates

## Motivation and Context

Inspired by your library, I'm developing my own, based on your approach, but for processing pptx presentations. When working with templates, I found viewing the base template as a string rather inconvenient. Embed solved this problem. 
so I wanted to offer this to you after some time

## How Has This Been Tested

go test

## Types of changes

## Checklist


xuri (commented):
Thanks for the PR. Since the new template files use pretty-printing with four space indentation and newlines, it increases the default workbook size and effect file-hash comparisons for users, so I don't think this change is necessary. You can easily introduce new variables in template.go within your own fork.

xuri (commented):
Thanks for the PR. Since the new template files use pretty-printing with four space indentation and newlines, it increases the default workbook size and effect file-hash comparisons for users, so I don't think this change is necessary. You can easily introduce new variables in `template.go` within your own fork.

PR #2335, opened 2026-06-04
fix: compact sparse worksheet rows in linear time

rootsec1:
## What changed

This changes `trimRow` to compact worksheet rows in place with a write index instead of deleting each empty row from the middle of the backing slice.

The observable behavior is the same: empty rows are removed, cells inside retained rows are still trimmed, and rows with attributes such as `hidden` are preserved.

## Why

Some valid XLSX files have a very sparse worksheet row list, for example when the sheet dimension reaches Excel's maximum row because of a single far-down cell. In that case, many intermediate rows can be empty.

The previous implementation used:

```go
sheetData.Row = append(sheetData.Row[:k], sheetData.Row[k+1:]...)
```

inside the scan loop. When many empty rows are present, that repeatedly shifts the remaining slice contents and can make saving sparse worksheets take quadratic time.

## How this fixes it

The new implementation performs one forward pass:

- trim cells in each row
- copy retained rows to the next write position
- clear the unused tail so discarded rows can be collected
- return the compacted slice

This avoids repeated large memory moves while preserving row order and the existing trim semantics.

## Tests

Added synthetic coverage only; no external or private workbook fixtures are included.

- `TestTrimRowCompactsSparseRows` verifies blank rows and blank cells are removed while rows with attributes are retained.
- `BenchmarkTrimRowSparseRows` covers a sparse worksheet shape with 50,000 row entries and only a small number of retained rows.

Local verification:

```text
go test ./...
go test ./... -run '^$' -bench '^BenchmarkTrimRowSparseRows$' -benchmem
```

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -197,16 +197,14 @@ func trimRow(sheetData *xlsxSheetData) []xlsxRow {
 		i   int
 	)
 
-	for k := 0; k < len(sheetData.Row); k++ {
+	for k := range sheetData.Row {
 		row = sheetData.Row[k]
 		if row = trimCell(row); len(row.C) != 0 || row.hasAttr() {
 			sheetData.Row[i] = row
 			i++
-			continue
 		}
-		sheetData.Row = append(sheetData.Row[:k], sheetData.Row[k+1:]...)
-		k--
 	}
+	clear(sheetData.Row[i:])
 	return sheetData.Row[:i]
 }
 
--- a/merge_test.go
+++ b/merge_test.go
@@ -1,7 +1,10 @@
 package excelize
 
 import (
+	"encoding/xml"
+	"fmt"
 	"path/filepath"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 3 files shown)


sheet.go
```diff
@@ -197,16 +197,14 @@ func trimRow(sheetData *xlsxSheetData) []xlsxRow {
...
 		row = sheetData.Row[k]
 		if row = trimCell(row); len(row.C) != 0 || row.hasAttr() {
 			sheetData.Row[i] = row
 			i++
-			continue
 		}
-		sheetData.Row = append(sheetData.Row[:k], sheetData.Row[k+1:]...)
-		k--
 	}
+	clear(sheetData.Row[i:])
```
xuri, line 207: Here is the comparison of main branch [`875b959`](https://github.com/xuri/excelize/tree/875b959ba62fc9bf45bb0de547ae7c7f7549835c) and the PR branch:

```bash
$ go test -benchmem -bench BenchmarkTrimRowSparseRows
before: BenchmarkTrimRowSparseRows-32     1	 1847849900 ns/op  4016400 B/op  13 allocs/op
after:  BenchmarkTrimRowSparseRows-32  1252	     951312 ns/op  4006282 B/op   2 allocs/op
```

xuri (approved):

PR #2343, opened 2026-06-24
This closes #2355, improve chart and style XML element compatibility to prevent generate corrupted workbook

### LINKED ISSUE #2355: AddChart: <c:marker> and <c:invertIfNegative> serialized after <c:dLbls>, out of the CT_LineSer/CT_BarSer sequence (Excel repairs on open)
### Description

`AddChart` (and the combo overload `AddChart(sheet, cell, series, combo...)`) serializes the
series children in the wrong order. Per `dml-chart.xsd`:

- `CT_LineSer` sequence is `idx, order, tx, spPr, marker, dPt, dLbls, trendline, errBars, cat, val, smooth, extLst` — so `<c:marker>` must appear **immediately after `<c:spPr>`**, before `<c:dLbls>`.
- `CT_BarSer` sequence is `idx, order, tx, spPr, invertIfNegative, pictureOptions, dPt, dLbls, trendline, errBars, cat, val, shape, extLst` — so `<c:invertIfNegative>` must appear **immediately after `<c:spPr>`**.

excelize instead emits both elements **after `<c:dLbls>`**. The part is schema-invalid; desktop
Excel repairs it on open.

### Excelize version

v2.11.0

### Go version

go1.26.5

### Steps to reproduce

A combo column + line chart where the line series carries a marker (the column series carries the
default `invertIfNegative`):

```go
package main

import "github.com/xuri/excelize/v2"

func main() {
	f := excelize.NewFile()
	for i, row := range [][]interface{}{
		{"Q", "Revenue", "Margin"},
		{"Q1", 120, 0.30},
		{"Q2", 140, 0.32},
		{"Q3", 160, 0.35},
	} {
		cell, _ := excelize.CoordinatesToCellName(1, i+1)
		f.SetSheetRow("Sheet1", cell, &row)
	}
	col := &excelize.Chart{
		Type: excelize.Col,
		Series: []excelize.ChartSeries{
			{Name: "Sheet1!$B$1", Categories: "Sheet1!$A$2:$A$4", Values: "Sheet1!$B$2:$B$4"},
		},
	}
	line := &excelize.Chart{
		Type: excelize.Line,
		Series: []excelize.ChartSeries{
			{Name: "Sheet1!$C$1", Categories: "Sheet1!$A$2:$A$4", Values: "Sheet1!$C$2:$C$4",
				Marker: excelize.ChartMarker{Symbol: "circle"}},
		},
	}
	if err := f.AddChart("Sheet1", "E1", col, line); err != nil {
		panic(err)
	}
	if err := f.SaveAs("combo.xlsx"); err != nil {
		panic(err)
	}
}
```

Then unzip `combo.xlsx` and inspect `xl/charts/chart1.xml`.

### Observed XML (line series, `<c:marker>` after `<c:dLbls>`)

```xml
<c:ser>
  <c:idx val="1"/>
  <c:order val="1"/>
  <c:tx>...</c:tx>
  <c:spPr>...</c:spPr>
  <c:dLbls>...</c:dLbls>
  <c:marker>                <!-- out of sequence: belongs before <c:dLbls> -->
    <c:symbol val="circle"/>
  </c:marker>
  <c:cat>...</c:cat>
  <c:val>...</c:val>
  <c:smooth val="0"/>
</c:ser>
```

The column (bar) series shows the same defect with `<c:invertIfNegative>` emitted after
`<c:dLbls>` instead of immediately after `<c:spPr>`.

### Expected XML (per ECMA-376 `dml-chart.xsd`, `CT_LineSer`)

```xml
<c:ser>
  <c:idx val="1"/>
  <c:order val="1"/>
  <c:tx>...</c:tx>
  <c:spPr>...</c:spPr>
  <c:marker>                <!-- immediately after <c:spPr> -->
    <c:symbol val="circle"/>
  </c:marker>
  <c:dLbls>...</c:dLbls>
  <c:cat>...</c:cat>
  <c:val>...</c:val>
  <c:smooth val="0"/>
</c:ser>
```

and, in the bar series, `<c:invertIfNegative>` immediately after `<c:spPr>` per `CT_BarSer`.

### Impact

Every workbook with a line/marker or bar series triggers Excel's "We found a problem with some
content… Do you want us to recover as much as we can?" repair dialog on open. After repair the
chart renders, but the round-trip is lossy and the prompt is alarming to end users.

1436761858:
# Related Issue
https://github.com/qax-os/excelize/issues/2342

## Description

This change updates the save path for temp-file-backed workbook parts.

Currently, when `writeToZip` writes entries from `f.tempFiles`, it calls `f.readBytes(path)`. If the entry is backed by a temp file, `readBytes` opens the temp file, reads the entire content into memory with `io.ReadAll`, stores it back into `f.Pkg`, and then writes the resulting byte slice to the ZIP writer.

This change avoids materializing the whole temp file in heap. Instead, `writeToZip` opens the temp file with `f.readTemp(path)` and streams it directly to the ZIP writer with `io.Copy`.

The saved workbook content is intended to be functionally equivalent:

```text
Before:
temp file -> []byte in memory -> ZIP writer

After:
temp file -> ZIP writer

The same temp-file-backed entry is written to the same ZIP path with the same bytes. The change only affects how the bytes are transported during save, not the generated workbook content or workbook structure.
ZIP64 handling is preserved by using the byte count returned by io.Copy, which is equivalent to the previous Write byte count for this purpose.

A new test, TestWriteToZipStreamsTempFiles, was added to cover the temp-file-backed save path.
The test verifies that:
A temp-file-backed entry is still written into the generated ZIP output.
The written ZIP entry content matches the original temp file content.
The temp-file-backed content is not loaded back into f.Pkg during save.
The third assertion is important because readBytes(path) stores temp file content into f.Pkg after reading it. Verifying that f.Pkg does not contain the entry after save helps ensure this path remains streaming and does not regress back to full in-memory loading.

### CODE REVIEW

**Comment by xuri** [comment]
Thanks for your PR. Any benchmark data on the performance impact of the changes? Specifically, how much memory is saved, and any speed impact? I suggest test with [excelize-benchmark](https://github.com/xuri/excelize-benchmark) for performance related changes.

**Comment by xuri** [comment]
The following benchmark report compares memory usage before and after the changes in this PR. The benchmarks were conducted by generating a workbook generated by [xuri/excelize-benchmark](https://github.com/xuri/excelize-benchmark) containing 512K cells. 

**Environment:**
Windows 11 Enterprise, 13th Gen Intel&reg; Core&trade; i9-13900 (2.00 GHz), 64GB DDR5 RAM, 500GB SSD, Go version go1.25.5 windows/amd64

**Workbook generated using the following command:**

```bash
./v2.11.0-master -func=StreamWriter -rows=102400 -cols=50 -n=6
```

Base version [`09f0656`](https://github.com/xuri/excelize/tree/09f06564d8147d6e015a5b7190613280a3bac852)
 
```text
Func: StreamWriter_r102400xc50.xlsx  RSS = 182 MB  Alloc = 134 MB  TotalAlloc = 1470 MB  Sys = 189 MB  NumGC = 71  Cost = 3.1179301s
```

This PR: #2343

```text
Func: StreamWriter_r102400xc50.xlsx  RSS = 180 MB  Alloc = 134 MB  TotalAlloc = 1470 MB  Sys = 181 MB  NumGC = 71  Cost = 3.0502737s
```

xuri (approved):
According the benchmark, there are no significant memory reduced, so I not suggest accept this changes. I've made some changes based on your branch to fix another issue instead of close this directly.

PR #2346, opened 2026-06-29
This closes #2344, fixes CalcCellValue function returning incorrect calculation result when error values inside parentheses or operators

### LINKED ISSUE #2344: CalcCellValue: error values inside parentheses or operators escape IFERROR
### Description

When a formula error (e.g. `#DIV/0!`) is produced by a bare sub-expression that forms an entire function argument, `IFERROR` traps it **correctly**. But when the **same** erroring sub-expression is wrapped in parentheses, or is followed by an operator within the argument, the error is no longer trapped. It propagates all the way out of `CalcCellValue` as the returned error and `IFERROR` is never applied.

For example, `=IFERROR(1/0, 3)` correctly returns 3. But `=IFERROR((1/0), 3)` returns `""` with `err = #DIV/0!`. The error escapes instead of being caught.

The trigger is broader than parentheses. Any formula that forces the erroring operand to be evaluated before the argument's closing comma reproduces it, such as `=IFERROR(1/0+0, 3)` too.

This is also not specific to IFERROR and applies to any error-trapping function e.g. IFNA, ISERROR, ISERR, etc.

1. create an Excel file with the following formulas in cells A1, A2, and A3:
  A1. `=IFERROR(1/0, 3)`
  A2. `=IFERROR((1/0), 3)`
  A3. `=IFERROR(1/0+0, 3)`
2. in Go, open the file and call CalcCellValue on the cells:
```go
package main

import (
	"fmt"

	"github.com/xuri/excelize/v2"
)

func main() {
	f, _ := excelize.OpenFile("Book1.xlsx")

	for _, cell := range []string{"A1", "A2", "A3"} {
		val, err := f.CalcCellValue("Sheet1", cell)
		fmt.Printf("%s => value=%q, err=%v\n", cell, val, err)
	}
}
```

```
A1 => value="3", err=<nil>
A2 => value="", err=#DIV/0!
A3 => value="", err=#DIV/0!
```

```
A1 => value="3", err=<nil>
A2 => value="3", err=<nil>
A3 => value="3", err=<nil>
```

This would match Microsoft Excel's behaviour.

### Go version

1.24.4

93f0b3caed37f21ef5079e3259c6c21dcfe68453 

### Environment

```shell
OS: Ubuntu 24.04 (linux/amd64)
Go: go1.24.4
excelize: master @ 93f0b3caed37f21ef5079e3259c6c21dcfe68453
Reference application: Microsoft Excel 365 (Ver 2605 Build 16.0.20026.20182) 64-bit
```

### Validations

huijian3139:
- Update documentation for the PivotTableField data type
- Update unit tests
- Update GitHub Action packages version

# PR Details

## Types of changes

## Checklist

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -20,7 +20,7 @@ jobs:
         cache: false
 
     - name: Checkout code
-      uses: actions/checkout@v6
+      uses: actions/checkout@v7
 
     - name: Get dependencies
       run: |
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -28,7 +28,7 @@ jobs:
     - name: Build
       run: go build -v .
 
-    - name: Build on ARM
+    - name: Build on ARM and RISC-V
       if: runner.os == 'Linux'
       run: |
         GOARCH=arm GOARM=5 go build .
@@ -36,6 +36,7 @@ jobs:
         GOARCH=arm GOARM=7 go build .
         GOARCH=arm64 go build .
         GOARCH=arm64 GOOS=android go build .
+        GOARCH=riscv64 go build .
 
     - name: Test
       run: env GO111MODULE=on go test -v -timeout 60m -race ./... -coverprofile='coverage.txt' -covermode=atomic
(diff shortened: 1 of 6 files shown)


xuri (approved):

Issue #1164, opened 2022-03-01
Nested Formular Error with CalCellValue


ffleader1:
**Description**
Cannot use nested function with calculate cell value. It produces incorrect result.

Try this function: ```=IF((ROUNDUP(IF(A1=16,(A2*2+4*0.3+1+1),IF(A1=4,(A2+2*0.3+1+1),(A3*SQRT(A1/2)+(SQRT(A1/2)*0.3)+1+1))),0))<=A4,(ROUNDUP(IF(A1=16,(A2*2+4*0.3+1+1),IF(A1=4,(A2+2*0.3+1+1),(A3*SQRT(A1/2)+(SQRT(A1/2)*0.3)+1+1))),0)),-1)```

It works in excel but will not work with the plugin. It gives this error: ```IF accepts at most 3 arguments```

**Excelize version:**

```text
v2.5.0
```

xuri:
Thanks for your feedback, I have fixed this issue, please try upgrade to the master branch code, and this patch will be released in the next version.

PR #2348, opened 2026-07-09
Related to #1164: fix incorrect argument count with nested func in parenthesized sub-exporessions

Aprax14:
## Description

When a function argument contains an expression like 1+(ABS(A1)+A1)/2, the formula parser in prepareEvalInfixExp miscounts the outer function's arguments (e.g. IF or SUM), throwing errors like "IF accepts at most 3 arguments".

Root cause: when an inner function closes, the code checks whether the operand on opfdStack belongs to the current function or an outer expression by looking at the token directly below on opftStack. If a subexpression-start token "(" or another function-start token (e.g. SUM) sits between the function marker and the actual infix operator (e.g. +), this check fails and the operand gets wrongly consumed as a function argument.

## Related Issue

https://github.com/qax-os/excelize/issues/1164

## Motivation and Context

The added unit tests would fail without this fix while they are valid in MS Excel.

## How Has This Been Tested

Added unit tests

## Types of changes

## Checklist

### DIFF
--- a/calc.go
+++ b/calc.go
@@ -1213,11 +1213,11 @@ func prepareEvalInfixExp(opfStack, opftStack, opfdStack, argsStack *Stack) {
 	argument := true
 	if opftStack.Len() > 2 && opfdStack.Len() == 1 {
 		topOpt := opftStack.Pop()
-		// Look past any subexpression start and function start tokens.
+		// Look past any subexpression start, function start, and prefix operator tokens
 		var savedTokens []efp.Token
 		for opftStack.Len() > 0 {
-			tok := opftStack.Peek().(efp.Token)
-			if !isBeginParenthesesToken(tok) && !isFunctionStartToken(tok) {
+			token := opftStack.Peek().(efp.Token)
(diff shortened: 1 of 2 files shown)


calc.go
```diff
@@ -1213,9 +1213,22 @@ func prepareEvalInfixExp(opfStack, opftStack, opfdStack, argsStack *Stack) {
 	argument := true
 	if opftStack.Len() > 2 && opfdStack.Len() == 1 {
 		topOpt := opftStack.Pop()
-		if opftStack.Peek().(efp.Token).TType == efp.TokenTypeOperatorInfix {
+		// Look past any subexpression start and function start tokens.
```
xuri: The `TokenTypeOperatorPrefix` tokens can also appear between the function separator and an infix operator, when ABS closes inside `SUM(1+(-ABS(A1)))`, will caused failure.

xuri (changes requested):

xuri (changes requested):
Great work! I've made some changes based on your code branch.

xuri (approved):

PR #2350, opened 2026-07-14
Related to #1164: fix errors pushed as extra func arguments

Aprax14:
# PR Details

Fix calculation error propagation in function argument expressions

## Description

This fixes incorrect error propagation in `prepareEvalInfixExp` when an arithmetic error (e.g. `#DIV/0!`) occurs while evaluating an expression inside a function argument. 
Previously, when `calculate()` returned an error, the error was pushed to `argsStack` (the function's argument list) instead of `opfdStack` (the operand stack of the current expression). This caused each failed operation to be counted as an additional argument of the enclosing function. For example, `IF(C1=0,0,A1/C1+B1/C1)` with `C1=0` would fail with "IF accepts at most 3 arguments".

## Related Issue

Related to recent PR https://github.com/qax-os/excelize/pull/2348 and also related to https://github.com/qax-os/excelize/issues/2344

## Motivation and Context

Formulas like =IF(C1=0,0,A1/C1+B1/C1) would fail while still being a valid excel formula.

## How Has This Been Tested

Added unit test

## Types of changes

## Checklist

### DIFF
--- a/calc.go
+++ b/calc.go
@@ -1118,7 +1118,7 @@ func (f *File) evalInfixExp(ctx *calcContext, sheet, cell string, tokens []efp.T
 					// calculate trigger
 					topOpt := opftStack.Peek().(efp.Token)
 					if err := calculate(opfdStack, topOpt); err != nil {
-						argsStack.Peek().(*list.List).PushFront(newErrorFormulaArg(formulaErrorVALUE, err.Error()))
+						opfdStack.Push(newErrorFormulaArg(err.Error(), err.Error()))
 					}
 					opftStack.Pop()
 				}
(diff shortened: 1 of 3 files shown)


calc.go
```diff
@@ -1204,7 +1204,7 @@ func prepareEvalInfixExp(opfStack, opftStack, opfdStack, argsStack *Stack) {
 		// calculate trigger
 		topOpt := opftStack.Peek().(efp.Token)
 		if err := calculate(opfdStack, topOpt); err != nil {
-			argsStack.Peek().(*list.List).PushBack(newErrorFormulaArg(err.Error(), err.Error()))
+			opfdStack.Push(newErrorFormulaArg(err.Error(), err.Error()))
```
xuri, line 1207: The function `prepareEvalInfixExp` is only called for the last argument of a function (when `)` is encountered). Swapping the argument position of the bad expression still breaks, for example: `IF(A1/C1+B1/C1,0,1)`.

xuri (changes requested):

xuri (approved):
Thanks for your great work! I've made some changes based on your code branch.

Issue #1527, opened 2023-04-21
stream rows read data enhancement


fudali113:
Expect to be able to get more content when streaming reads the data:

such as:
- [ ] get cols info
- [ ] get row cells attrs
- [ ] get mergeCells info
- [ ] get dimension info
- [ ] get sheetViews info

etc.

xuri:
Thanks for your issue. The aim of creating the stream mode functions was not re-provide each function as the normal mode functions, it was just used to process the workbook with the huge amount of data, we need to evaluate if those features suit a typical scenario.

fudali113:
The problem I am facing now is that obtaining relevant information of sheet is a rigid requirement; When we encounter large files, we can reduce the memory overhead by streaming row data, but when we get other sheet information, we can only downgrade to full memory load.

Because these information can be obtained without performance loss during stream reading, it is hoped to enhance stream reading to obtain relevant information.

Issue #1100, opened 2021-12-22
Exponential grown worktime of StreamWriter 


IAkumaI:
**Description**

I use StreamWriter for generate large xlsx files. It's just text files: no formatting/styles.

```go
// about 100 columns with different data. But data size does not seriously matter in my tests
dataRow := []interface{}{"Some", "Cell", "Values", "and 100 more"}

// Do not check errors in issue
excFile := excelize.NewFile()
streamWriter, _ := excFile.NewStreamWriter("Sheet0")
for rowNum := 1; rowNum < 50000; rowNum++ {
  streamWriter.SetRow("A"+strconv.Itoa(rowNum), dataRow)
}

streamWriter.Flush()
file, _ := os.Create("some.xlsx")
excFile.Write(file)
```

In code example we create excel file with 50 000 rows with StreamWriter.
All code will consume about 1.5 minute on my server.
But if rows count increased to 100 000 in will consume not (2 * 1.5) min, but about 10 minutes.
If there are million rows, it takes about 1.5 hours.

Time increates not linear. Is there any way to make time more linear and speed up much rows creation?

**Output of `go version`:**

```text
1.17.2
```

**Excelize version or commit ID:**

```text
github.com/xuri/excelize/v2 v2.4.1
```

**Environment details (OS, Microsoft Excel™ version, physical, etc.):**
Ubuntu x86_64 with SSD drive.

xuri:
Thanks for your issue, here is some performance figures for your reference: [Performance Figures](../wiki#performance-figures).

PR #2352, opened 2026-07-15
Support absolute paths for GetFormControls and GetTables functions

### LINKED ISSUE #2386: GetTables returns no tables when the table relationship uses an absolute target path
### Description

`GetTables` silently returns an empty slice and a `nil` error for a worksheet whose table relationship target is package-absolute (`Target="/xl/tables/table1.xml"`) rather than relative (`Target="../tables/table1.xml"`). Both forms are valid OPC, and Excel does emit the absolute form.

Because the error is `nil`, callers cannot distinguish "this sheet has no tables" from "the table was not found".

This is the same class of bug as #1987, fixed for `GetPictures`/`GetPictureCells` in #1988; `table.go` was not covered by that change.

### Steps to reproduce

Take any workbook containing a table and make its relationship target absolute in `xl/worksheets/_rels/sheet1.xml.rels`:

```xml
<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
    <Relationship Id="rId1"
        Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/table"
        Target="/xl/tables/table1.xml"/>
</Relationships>
```

```go
f, err := excelize.OpenFile("TableAbsoluteTarget.xlsx")
assert.NoError(t, err)
tables, err := f.GetTables("Sheet1")
assert.NoError(t, err)
assert.Len(t, tables, 1) // fails: got 0
```

With the target written as `../tables/table1.xml` the same workbook returns the table as expected.

### Expected / actual

Expected: the table (`name="EmployeeTable" ref="A1:G9"`) is returned.
Actual: `len(tables) == 0`, `err == nil`.

### Root cause

`table.go`:

```go
target := f.getSheetRelationshipsTargetByID(sheet, tbl.RID)
tableXML := strings.ReplaceAll(target, "..", "xl")
content, ok := f.Pkg.Load(tableXML)
if !ok {
    continue
}
```

`f.Pkg` is keyed without a leading slash (`xl/tables/table1.xml`). An absolute target contains no `..`, so `ReplaceAll` leaves it as `/xl/tables/table1.xml`, `Pkg.Load` misses, and the loop `continue`s.

### Suggested fix

Trim the leading slash, matching what the codebase already does elsewhere — `adjust.go` handles drawings correctly a few lines away from the table code:

```go
drawingXML := strings.TrimPrefix(strings.ReplaceAll(target, "..", "xl"), "/")
```

so:

```go
tableXML := strings.TrimPrefix(strings.ReplaceAll(target, "..", "xl"), "/")
```

The same untrimmed line also appears in `adjustTable` (`adjust.go`), where the missed lookup means `InsertRows`/`RemoveRow`/`InsertCols` silently skip adjusting the table's `ref`, leaving a stale range in the saved workbook. Since `getTables()` fans out from the same lookup, an affected table is also invisible to `DeleteTable`, `AddTable`'s duplicate-name check, `GetSlicers` and `GetPivotTables`. A small shared helper would cover all of them.

Comparable handling already exists in `getWorksheetPath` (`sheet.go`) and `getWorkbookPath` (`workbook.go`).

Affects v2.11.0 and master.

freeeve:
## Description

This PR reduces CPU time and per-cell allocations in the three highest-volume paths -- the stream writer, the rows iterator, and per-cell value setting -- without changing any public API or serialized output. It contains four commits, each independently benchmarked:

1. **I/O buffering and fast paths** -- decode worksheet parts that were spilled to temporary files through a 1MB buffered reader instead of the XML decoder's internal 4KB buffer; copy streamed sheet data into the ZIP with a 1MB buffer; skip the XML escaper when a value contains no character requiring escape; return early from `bstrMarshal` when no `_x` escape sequence can be present (mirroring the existing guard in `bstrUnmarshal`); build cell references with a single allocation; replace the reflection-based `DecodeElement` calls for `<v>` and `<is>` in the rows iterator with direct token handling.
2. **Zero-allocation `StreamWriter.SetRow`** -- serialize plain scalar values (integers, floats, booleans, and strings needing no escaping or whitespace preservation) directly into the stream buffer with `strconv.Append*` into reused scratch buffers. Values needing escaping, truncation, formulas, rich text, or special formatting keep the existing generic path. A new test asserts byte-exact worksheet XML across both paths.
3. **Skip calculation cache clearing when nothing has been cached** -- every cell mutation cleared three `sync.Map`s, and `sync.Map.Clear` allocates even when the map is already empty, so bulk writes paid three allocations per cell for caches they never used. Cache population is now tracked with an atomic flag.
4. **Rows iterator allocation pass** -- reuse a single cell structure and character-data scratch buffer across cells, bundle the string item and text element of the common single-text inline string into one allocation, and pre-size each returned row slice from the widest row seen so far.

### Benchmark results

**Official suite** ([xuri/excelize-benchmark](https://github.com/xuri/excelize-benchmark)), master @ 1a6d13b vs this branch, both built via `go.mod` replace, runs interleaved base/branch in alternating order on a load-gated otherwise-idle machine (Apple M3 Max, Go 1.25.6). Aggregate across all 59 configs: **median -11.7% time, -25% allocations**. Per family (negative = this PR better):

| function family | time | allocations | notes |
|---|---|---|---|
| StreamWriter | -22% to -34% | -17% to -74% | allocation reduction grows with row count (-74% at 102400 rows) |
| GetRows | -11% to -29% (typ. -23%) | -27% | consistent across the size sweep |
| RowIterator | -11% to -23% | -27% | |
| SetSheetRow | -8% to -12% | -24% | consistent across the size sweep |
| MergeCell | flat to -13% | -12% to -25% | |
| AddComment | flat to -6% | 0% | |
| AddChart / SetCellHyperLink / AddPicture | flat (within a few %) | 0% | paths this PR does not touch; unchanged as expected |

Functions this PR does not touch show identical allocation counts and time within run-to-run variance, confirming the changes are isolated to the targeted paths.

**Micro-benchmarks** (`go test -bench -benchmem`; the mixed-workload benchmarks `BenchmarkStreamWriterSetRowMixed` and `BenchmarkRowsColumnsMixed` are added in this PR, `BenchmarkSetCellValue` already existed):

| benchmark | master | this PR | change |
|---|---|---|---|
| `StreamWriter.SetRow` (20-cell mixed row) | 3684 ns/op, 2648 B/op, 111 allocs/op | 950 ns/op, 59 B/op, 1 alloc/op | -74% time, -99% allocs |
| `Rows`/`Columns` (10k rows x 20 cells) | 230 ms/op, 207 MB/op, 4.43M allocs/op | 180 ms/op, 146 MB/op, 3.39M allocs/op | -22% time, -23% allocs |
| `SetCellValue` (6 string cells) | 3421 ns/op, 5036 B/op, 54 allocs/op | 1737 ns/op, 1586 B/op, 23 allocs/op | -49% time, -57% allocs |

End to end on a 1,000,000 row x 20 column mixed string/number workload (minimum of 3 runs): streaming write 11.6s -> ~8s, `GetRows` 27.1s -> ~21s, bulk `SetCellValue` write 44.5s -> ~42s.

## Related Issue

Related to #1100 (StreamWriter performance) and #1527 (stream rows read enhancement).

## Motivation and Context

Profiling large-workbook workloads (CPU and allocation profiles on 1M x 20 read/write runs) showed most of the time outside compression going to per-cell allocations and their GC pressure: reference-string and value-string allocations per streamed cell, reflection-based unmarshaling per read cell, regex evaluation per written string, and three `sync.Map.Clear` allocations per mutated cell. All of these are avoidable without changing behavior, which is what this PR does. After these changes the streaming write path is dominated by deflate, and the read path by `encoding/xml` tokenization.

## How Has This Been Tested

- The full test suite passes (`go test .`), coverage 99.8% of statements; all new functions and their fallback/error branches are covered.
- `TestStreamWriterPlainCellXML` and `TestStreamWriterPlainCellFallbacks` assert byte-exact worksheet XML for values served by the new direct serialization path and for every fallback (escaping, `xml:space` preservation, NaN, formulas, over-length strings, column overflow).
- `TestCellXMLHandler` (existing) verifies the manual `<v>`/`<is>` decoding produces structures deep-equal to `encoding/xml` unmarshaling, including error parity on malformed input; new tests cover nested elements inside `<v>`, truncated elements, and phonetic runs (whose text must not leak into values).
- New fuzz tests (`FuzzIsCanonicalNumber`, `FuzzXMLPlainText`) verify the fast-path predicates against the reference implementations (`strconv` round-trip and `xml.EscapeText` identity).
- Benchmarked before/after with the official [xuri/excelize-benchmark](https://github.com/xuri/excelize-benchmark) suite (59 configs, interleaved base/branch runs on a load-gated machine) and with the micro-benchmarks listed above; the two new mixed-workload benchmarks are included in this PR so future regressions are measurable.

## Types of changes

## Checklist


xuri (approved):
Thanks for your contribution. This pull request contains a large amount of code, and I'd prefer to have each performance optimization submitted as a separate pull request. I'll evaluate each optimization individually later, select the ones that are beneficial, and merge them into the master branch.
I reverted this change and, instead of simply closing the pull request, I used your branch as the basis for fixing another issue.

Issue #2357, opened 2026-07-25
High memory usage when saving workbooks written with StreamWriter


ValeryVerkhoturov:
### Description

The `StreamWriter` keeps memory usage flat while writing rows by spilling worksheet XML to a temporary file every 16 MB. However, that benefit is lost when the workbook is saved: `Save`, `SaveAs`, and `Write` all route through `File.WriteTo`, which calls `WriteToBuffer` and assembles the **entire compressed zip archive in an in-memory `bytes.Buffer`** before writing a byte to the destination. Peak memory therefore scales with the output file size (plus `bytes.Buffer` growth doubling), which defeats the purpose of the streaming writer for large workbooks (1 million rows and more).

Two smaller related problems in the same path:

1. `writeToZip` copies the stream temp file into the archive with `io.Copy`'s default 32 KB buffer; on a 1M-row file these small `pread` syscalls account for ~46% of total CPU time in profiles.
2. The `tempFiles` loop uses `readBytes`, which loads each on-disk part fully into memory and also stores it into `f.Pkg` permanently.

### Steps to reproduce the issue

1. Write 1,000,000 rows (10 cells each) with `NewStreamWriter`/`SetRow`/`Flush` — memory stays flat (~22 MB heap) as expected.
2. Call `SaveAs` and observe peak memory while saving.

```go
package main

import (
        "fmt"
        "runtime"
        "strconv"

        "github.com/xuri/excelize/v2"
)

func heapMB() uint64 {
        var m runtime.MemStats
        runtime.GC()
        runtime.ReadMemStats(&m)
        return m.HeapAlloc >> 20
}

func main() {
        f := excelize.NewFile()
        defer f.Close()
        sw, err := f.NewStreamWriter("Sheet1")
        if err != nil {
                fmt.Println(err)
                return
        }
        row := make([]interface{}, 10)
        for r := 1; r <= 1000000; r++ {
                for c := 0; c < 10; c++ {
                        if c%2 == 0 {
                                row[c] = r + c
                        } else {
                                row[c] = "value-" + strconv.Itoa(r+c)
                        }
                }
                cell, _ := excelize.CoordinatesToCellName(1, r)
                if err := sw.SetRow(cell, row); err != nil {
                        fmt.Println(err)
                        return
                }
        }
        if err := sw.Flush(); err != nil {
                fmt.Println(err)
                return
        }
        fmt.Println("heap after write:", heapMB(), "MB") // ~22 MB, streaming works
[...]

PR #2358, opened 2026-07-25
#2357 - Fix high memory usage when saving workbooks written with StreamWriter

ValeryVerkhoturov:
# PR Details

Fix high memory usage when saving workbooks written with StreamWriter

## Description

This PR makes `Save`, `SaveAs`, and `Write` stream the workbook archive directly to the destination with constant memory, instead of assembling the entire compressed zip in an in-memory `bytes.Buffer`:

- `File.WriteTo` now streams the archive directly to the writer. The buffered `WriteToBuffer` path is kept only when `Options.Password` is set, because encryption requires the complete archive bytes.
- When any archive entry exceeds the ZIP64 4 GB threshold (detected up front by the new `zip64Needed` from actual stream/part sizes), the archive is spooled to a temporary file, its local file headers are patched in place by the new `patchZip64LFH` (chunked `ReadAt` scan + `WriteAt`, constant memory, same semantics as `writeZip64LFH`), and the result is copied to the writer.
- `writeToZip` copies stream temp files with a 1 MB `io.CopyBuffer` instead of `io.Copy`'s 32 KB buffer, and streams `tempFiles` parts from disk via `readTemp` instead of loading them whole into memory with `readBytes`.
- Added `bufferedWriter.Size()` used for the up-front ZIP64 check, and a `benchmark/main.go` program for measuring write/save time, throughput, and peak memory.

`WriteToBuffer` keeps its existing public contract and behavior.

## Related Issue

#1100 #2357

## Motivation and Context

The `StreamWriter` keeps memory flat during writing by spilling worksheet XML to a temporary file every 16 MB, but all of that was undone at save time:

1. **The whole compressed archive was built in memory.** `Save`, `SaveAs`, and `Write` all route through `File.WriteTo`, which called `WriteToBuffer` — so peak memory scaled with the output file size (plus `bytes.Buffer` growth doubling), defeating the purpose of the streaming writer. Full buffering existed for two reasons only: `writeZip64LFH` patches ZIP64 local file header versions in the buffer for entries over 4 GB, and password encryption needs the complete archive bytes.
2. **The stream temp file was read back inefficiently.** `writeToZip` copied it into the zip with `io.Copy`'s default 32 KB buffer; on a 1M-row file these small `pread` syscalls accounted for ~46% of total CPU time in profiles.
3. **On-disk workbook parts were loaded whole into memory.** The `tempFiles` loop used `readBytes`, which reads the entire part into memory and also stores it into `f.Pkg` permanently.

With this change, saving a workbook written with the stream writer uses constant memory regardless of the output size — for example, saving a 10-million-row workbook drops peak heap from ~1.1 GB to ~393 MB, where the remainder is write-phase stream buffers, not the save.

[excelize-benchmark](https://github.com/xuri/excelize-benchmark) compare:
excelize-benchmark compare

## How Has This Been Tested

Environment: macOS (Darwin, Apple silicon), Go 1.25.

**Unit tests** (added in this PR): `TestWriteDirect` (direct streaming round-trip through a plain `io.Writer`, password-encrypted `WriteTo` round-trip, temp-file part streamed from disk and not cached into `Pkg`, unreadable part skipped without error), `TestWriteSpooled` (spooled round-trip and error paths), `TestZip64Needed` (stream/temp-file/in-memory parts over the 4 GB threshold, using sparse files), `TestPatchZip64LFH` (patching across the chunk scanning boundary, truncated headers, filename length overflow), and `TestBufferedWriterSize`. The full test suite passes with 99.7% statement coverage, including the existing `TestZip64`, which saves a workbook containing a >4 GB worksheet entry through the new spooled path; that output was inspected byte-level — the >4 GB entry's local file header version is patched to 45, all other entries remain 20, and `unzip -t` reports a valid archive.

**Benchmark** (each row has 10 cells — 5 integers, 5 short strings; row counts above the 1,048,576 worksheet limit are split across multiple worksheets):

```go
package main

import (
	"fmt"
	"os"
	"runtime"
	"strconv"
	"sync/atomic"
	"time"

	"github.com/xuri/excelize/v2"
)

const output = "benchmark.xlsx"

func main() {
	rows := 1000000
	if len(os.Args) > 1 {
		var err error
		if rows, err = strconv.Atoi(os.Args[1]); err != nil {
			fmt.Println("invalid rows count:", os.Args[1])
			os.Exit(1)
		}
	}
	var peakHeap, peakSys atomic.Uint64
	done := make(chan struct{})
	go func() {
		var m runtime.MemStats
		t := time.NewTicker(10 * time.Millisecond)
		defer t.Stop()
		for {
			select {
			case <-done:
				return
			case <-t.C:
				runtime.ReadMemStats(&m)
				if m.HeapAlloc > peakHeap.Load() {
					peakHeap.Store(m.HeapAlloc)
				}
				if m.Sys > peakSys.Load() {
					peakSys.Store(m.Sys)
				}
			}
		}
	}()

	f := excelize.NewFile()
	defer func() {
			fmt.Println(err)
		}
	}()
	writeStart := time.Now()
	row := make([]interface{}, 10)
	remaining, sheetIdx := rows, 0
	for remaining > 0 {
		sheetIdx++
		sheet := "Sheet" + strconv.Itoa(sheetIdx)
		if sheetIdx > 1 {
			if _, err := f.NewSheet(sheet); err != nil {
				fmt.Println(err)
				return
			}
		}
		sw, err := f.NewStreamWriter(sheet)
		if err != nil {
			fmt.Println(err)
			return
		}
		n := min(remaining, excelize.TotalRows)
		for r := 1; r <= n; r++ {
			for c := 0; c < 10; c++ {
				if c%2 == 0 {
					row[c] = r + c
				} else {
					row[c] = "value-" + strconv.Itoa(r+c)
				}
			}
			cell, err := excelize.CoordinatesToCellName(1, r)
			if err != nil {
				fmt.Println(err)
				return
			}
			if err := sw.SetRow(cell, row); err != nil {
				fmt.Println(err)
				return
			}
		}
			fmt.Println(err)
			return
		}
		remaining -= n
	}
	writeDur := time.Since(writeStart)

	saveStart := time.Now()
	if err := f.SaveAs(output); err != nil {
		fmt.Println(err)
		return
	}
	saveDur := time.Since(saveStart)
	close(done)

	fi, err := os.Stat(output)
	if err != nil {
		fmt.Println(err)
		return
	}
	writeSpeed := float64(rows) / writeDur.Seconds()
	saveSpeed := float64(fi.Size()) / (1 << 20) / saveDur.Seconds()
	fmt.Printf("rows=%d write=%v (%.0f rows/s) save=%v (%.1f MB/s) peakHeap=%dMB peakSys=%dMB size=%dMB\n",
		rows, writeDur.Round(time.Millisecond), writeSpeed,
		saveDur.Round(time.Millisecond), saveSpeed,
		peakHeap.Load()>>20, peakSys.Load()>>20, fi.Size()>>20)
}
```

```bash
cd benchmark
go run main.go 1000
go run main.go 1000000
go run main.go 10000000
```

Timing and throughput (before → after):

| Rows | Sheets | Output size | Write time | Write speed | Save time | Save speed |
|---|---|---|---|---|---|---|
| 1,000 | 1 | 0.3 MB | 6ms → 6ms | ~170K rows/s → ~170K rows/s | 4ms → 4ms | — |
| 1,000,000 | 1 | 41 MB | 2.4s → 2.8s | 414K rows/s → 354K rows/s | 2.8s → 2.8s | 14.7 MB/s → 14.8 MB/s |
| 10,000,000 | 10 | 419 MB | 24.8s → 26.1s | 404K rows/s → 383K rows/s | 29.9s → 28.3s | 14.0 MB/s → 14.8 MB/s |

The write path (`SetRow`/`Flush`) is untouched by this PR, so the write-speed differences are run-to-run variance; save throughput is disk- and compression-bound and stays the same or slightly improves. Sub-second numbers for the 1,000-row case are dominated by process startup noise.

Memory, sampled every 10 ms (before → after):

| Rows | Peak heap | Peak process memory |
|---|---|---|
| 1,000 | 3 MB → 3 MB | 12 MB → 12 MB |
| 1,000,000 | **133 MB → 56 MB** | **183 MB → 70 MB** |
| 10,000,000 | **1,098 MB → 393 MB** | **1,431 MB → 421 MB** |

Benchmark outputs at every size were validated with `unzip -t` and by re-opening them with `excelize.OpenFile` and reading back first/last-row cell values. For the 10-million-row workbook specifically: all 10 worksheets parse as well-formed XML with exact row counts (9 × 1,048,576 + 562,816 = 10,000,000), and `Sheet1!A1`, `Sheet1!B1048576`, `Sheet10!A1`, and `Sheet10!B562816` all read back the expected values.

Note that the 10-million-row output is a valid workbook, but exceeds the capabilities of most desktop spreadsheet applications: it expands to ~4.85 GB of XML across ~100 million cells. Apple Numbers is limited to 1,000,000 rows per sheet, so it cannot import the 1,048,576-row worksheets and will hang or crash; desktop Excel needs many gigabytes of memory to load it. Verify large benchmark outputs programmatically rather than by opening them in a GUI application.

## Types of changes

## Checklist

- [x] All new and existing tests passed

### DIFF
--- a/file.go
+++ b/file.go
@@ -113,7 +113,10 @@ func (f *File) Write(w io.Writer, opts ...Options) error {
 	return err
 }
 
-// WriteTo implements io.WriterTo to write the file.
+// WriteTo implements io.WriterTo to write the file. The workbook will be
+// streamed to the writer without buffering the whole archive in memory,
+// except when saving with password protection, which requires the complete
+// archive for encryption.
 func (f *File) WriteTo(w io.Writer, opts ...Options) (int64, error) {
 	for i := range opts {
 		f.options = &opts[i]
@@ -127,11 +130,119 @@ func (f *File) WriteTo(w io.Writer, opts ...Options) (int64, error) {
 			return 0, err
 		}
 	}
-	buf, err := f.WriteToBuffer()
+	if f.options != nil && f.options.Password != "" {
+		buf, err := f.WriteToBuffer()
+		if err != nil {
+			return 0, err
+		}
+		return buf.WriteTo(w)
+	}
+	return f.writeDirect(w)
+}
+
+// countWriter wraps an io.Writer and counts the bytes written through it.
+type countWriter struct {
+	w       io.Writer
+	written int64
+}
+
+// Write writes to the underlying writer and accumulates the written bytes.
+func (cw *countWriter) Write(p []byte) (int, error) {
+	n, err := cw.w.Write(p)
+	cw.written += int64(n)
+	return n, err
+}
+
+// writeDirect provides a function to write the workbook to w, streaming
+// archive entries from their sources instead of buffering the whole archive
+// in memory. When any archive entry exceeds the ZIP64 size threshold, the
+// local file headers must be patched after writing, which requires random
+// access to the output, so the archive will be spooled to a temporary file
+// in that case.
+func (f *File) writeDirect(w io.Writer) (int64, error) {
+	if f.zip64Needed() {
+		return f.writeSpooled(w)
+	}
+	cw := countWriter{w: w}
+	zw := f.ZipWriter(&cw)
+	if err := f.writeToZip(zw); err != nil {
+		_ = zw.Close()
+		return cw.written, err
+	}
+	return cw.written, zw.Close()
+}
+
+// writeSpooled provides a function to write the workbook to a temporary file,
+// patch the ZIP64 local file headers there, and copy the result to w.
+func (f *File) writeSpooled(w io.Writer) (int64, error) {
+	var tmpDir string
+	if f.options != nil {
+		tmpDir = f.options.TmpDir
+	}
+	tmp, err := os.CreateTemp(tmpDir, "excelize-")
 	if err != nil {
 		return 0, err
 	}
-	return buf.WriteTo(w)
+	defer func() {
+		_ = tmp.Close()
+		_ = os.Remove(tmp.Name())
+	}()
+	zw := f.ZipWriter(tmp)
+	if err := f.writeToZip(zw); err != nil {
+		_ = zw.Close()
+		return 0, err
+	}
+	if err := zw.Close(); err != nil {
+		return 0, err
+	}
+	size, err := tmp.Seek(0, io.SeekEnd)
+	if err != nil {
+		return 0, err
+	}
+	if err := f.patchZip64LFH(tmp, size); err != nil {
+		return 0, err
+	}
+	if _, err := tmp.Seek(0, io.SeekStart); err != nil {
+		return 0, err
+	}
+	return io.CopyBuffer(w, tmp, make([]byte, 1<<20))
+}
+
+// zip64Needed returns whether any archive entry will exceed the ZIP64 size
+// threshold, which requires patching local file headers after writing the
+// archive, see writeZip64LFH.
+func (f *File) zip64Needed() bool {
+	needed := false
+	for _, stream := range f.streams {
+		if stream.rawData.Size() > math.MaxUint32 {
+			needed = true
+		}
+	}
+	if !needed {
+		f.Pkg.Range(func(path, content interface{}) bool {
+			if _, ok := f.streams[path.(string)]; ok {
+				return true
+			}
+			if b, ok := content.([]byte); ok && int64(len(b)) > math.MaxUint32 {
+				needed = true
+				return false
+			}
+			return true
+		})
+	}
+	if !needed {
+		f.tempFiles.Range(func(path, tmpPath interface{}) bool {
+			if _, ok := f.Pkg.Load(path); ok {
+				return true
+			}
+			if fi, err := os.Stat(tmpPath.(string)); err == nil && fi.Size() > math.MaxUint32 {
+				needed = true
+				return false
+			}
+			return true
+		})
+	}
+	return needed
 }
 
 // WriteToBuffer provides a function to get bytes.Buffer from the saved file,
@@ -175,6 +286,7 @@ func (f *File) writeToZip(zw ZipWriter) error {
 	f.styleSheetWriter()
 	f.themeWriter()
 
+	var copyBuf []byte
 	for path, stream := range f.streams {
 		fi, err := zw.Create(path)
 		if err != nil {
@@ -185,7 +297,10 @@ func (f *File) writeToZip(zw ZipWriter) error {
 			_ = stream.rawData.Close()
 			return err
 		}
-		written, err := io.Copy(fi, from)
+		if copyBuf == nil {
+			copyBuf = make([]byte, 1<<20)
+		}
+		written, err := io.CopyBuffer(fi, from, copyBuf)
 		if err != nil {
 			return err
 		}
@@ -229,8 +344,21 @@ func (f *File) writeToZip(zw ZipWriter) error {
 		if fi, err = zw.Create(path); err != nil {
 			break
 		}
-		if n, err = fi.Write(f.readBytes(path)); int64(n) > math.MaxUint32 {
+		file, readErr := f.readTemp(path)
+		if readErr != nil || file == nil {
+			continue
+		}
+		if copyBuf == nil {
+			copyBuf = make([]byte, 1<<20)
+		}
+		var written int64
+		written, err = io.CopyBuffer(fi, file, copyBuf)
+		_ = file.Close()
+		if written > math.MaxUint32 {
 			f.zip64Entries = append(f.zip64Entries, path)
+		}
+		if err != nil {
+			break
 		}
 	}
 	return err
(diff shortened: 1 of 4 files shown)


ValeryVerkhoturov:
Also [excelize-benchmark](https://github.com/xuri/excelize-benchmark).

At 102,400 rows that's a 2.5× reduction in RSS and ~5× reduction in live heap, with total allocations down ~9%. The memory curve went from linear-in-rows to O(1), which is the property users actually want from a streaming writer — it would keep paying off at even larger sheet sizes.                                                                                                                                     

Before changes
```
Func: StreamWriter_r200xc50.xlsx        RSS = 13 MB     Alloc = 2 MB    TotalAlloc = 6 MB       Sys = 13 MB     NumGC = 3       Cost = 6.868542ms
Func: StreamWriter_r400xc50.xlsx        RSS = 15 MB     Alloc = 3 MB    TotalAlloc = 10 MB      Sys = 17 MB     NumGC = 4       Cost = 12.925125ms
Func: StreamWriter_r800xc50.xlsx        RSS = 18 MB     Alloc = 5 MB    TotalAlloc = 18 MB      Sys = 16 MB     NumGC = 5       Cost = 24.831667ms
Func: StreamWriter_r1600xc50.xlsx       RSS = 26 MB     Alloc = 8 MB    TotalAlloc = 34 MB      Sys = 25 MB     NumGC = 7       Cost = 48.7965ms
Func: StreamWriter_r3200xc50.xlsx       RSS = 41 MB     Alloc = 16 MB   TotalAlloc = 66 MB      Sys = 41 MB     NumGC = 9       Cost = 96.489459ms
Func: StreamWriter_r6400xc50.xlsx       RSS = 71 MB     Alloc = 32 MB   TotalAlloc = 131 MB     Sys = 70 MB     NumGC = 11      Cost = 217.8025ms
Func: StreamWriter_r12800xc50.xlsx      RSS = 71 MB     Alloc = 40 MB   TotalAlloc = 220 MB     Sys = 70 MB     NumGC = 15      Cost = 423.312709ms
Func: StreamWriter_r25600xc50.xlsx      RSS = 79 MB     Alloc = 45 MB   TotalAlloc = 399 MB     Sys = 87 MB     NumGC = 24      Cost = 918.794333ms
Func: StreamWriter_r51200xc50.xlsx      RSS = 111 MB    Alloc = 77 MB   TotalAlloc = 756 MB     Sys = 119 MB    NumGC = 40      Cost = 1.822896833s
Func: StreamWriter_r102400xc50.xlsx     RSS = 174 MB    Alloc = 133 MB  TotalAlloc = 1470 MB    Sys = 182 MB    NumGC = 73      Cost = 3.677953042s
```

After changes
```
Func: StreamWriter_r200xc50.xlsx        RSS = 13 MB     Alloc = 3 MB    TotalAlloc = 7 MB       Sys = 13 MB     NumGC = 3       Cost = 8.117666ms
Func: StreamWriter_r400xc50.xlsx        RSS = 15 MB     Alloc = 4 MB    TotalAlloc = 11 MB      Sys = 17 MB     NumGC = 4       Cost = 12.639583ms
Func: StreamWriter_r800xc50.xlsx        RSS = 18 MB     Alloc = 5 MB    TotalAlloc = 18 MB      Sys = 17 MB     NumGC = 5       Cost = 24.442125ms
Func: StreamWriter_r1600xc50.xlsx       RSS = 26 MB     Alloc = 7 MB    TotalAlloc = 33 MB      Sys = 29 MB     NumGC = 7       Cost = 48.525542ms
Func: StreamWriter_r3200xc50.xlsx       RSS = 41 MB     Alloc = 13 MB   TotalAlloc = 64 MB      Sys = 41 MB     NumGC = 9       Cost = 101.738209ms
Func: StreamWriter_r6400xc50.xlsx       RSS = 71 MB     Alloc = 24 MB   TotalAlloc = 124 MB     Sys = 70 MB     NumGC = 11      Cost = 317.085083ms
Func: StreamWriter_r12800xc50.xlsx      RSS = 71 MB     Alloc = 25 MB   TotalAlloc = 205 MB     Sys = 70 MB     NumGC = 15      Cost = 572.722625ms
Func: StreamWriter_r25600xc50.xlsx      RSS = 71 MB     Alloc = 25 MB   TotalAlloc = 368 MB     Sys = 70 MB     NumGC = 23      Cost = 904.355166ms
Func: StreamWriter_r51200xc50.xlsx      RSS = 71 MB     Alloc = 26 MB   TotalAlloc = 693 MB     Sys = 75 MB     NumGC = 39      Cost = 1.836587042s
Func: StreamWriter_r102400xc50.xlsx     RSS = 71 MB     Alloc = 27 MB   TotalAlloc = 1343 MB    Sys = 70 MB     NumGC = 72      Cost = 3.585185041s
```

RAM alloc diagram:
ram alloc diagram

I expect, that 25-27 MB contains XML sheet with data 16 MB size + meta files. Something to consider: move the 16MB constant to the options.

file.go
```diff
@@ -127,11 +131,120 @@ func (f *File) WriteTo(w io.Writer, opts ...Options) (int64, error) {
...
 	}
-	buf, err := f.WriteToBuffer()
+	if f.options != nil && f.options.Password != "" {
+		buf, err := f.WriteToBuffer()
+		if err != nil {
+			return 0, err
+		}
+		return buf.WriteTo(w)
+	}
+	return f.writeDirect(w)
```
rayhankinan, line 141: Is there any plans on supporting password encryption on `f.writeDirect` function?

ValeryVerkhoturov: I do not have such plan. Also I am not sure what it is possible.

PR #2366, opened 2026-07-28
This closes #GHSA-48hm-4h8j-58fg, fix panic on negative shared string index

arpitjain099:
Fix for [GHSA-48hm-4h8j-58fg](https://github.com/qax-os/excelize/security/advisories/GHSA-48hm-4h8j-58fg).

> **Correction:** the original title and this line cited `GHSA-wcg2-648h-mhxq`, which belongs to a different project. The excelize advisory this fixes is `GHSA-48hm-4h8j-58fg`, which is what the link above always pointed to. My mistake; corrected here for anyone tracing the fix. The merge commit message still carries the wrong id and I would not suggest rewriting history over it.

`getFromStringItem` bounds the shared string index only from above:

```go
if len(f.sharedStringItem) <= index {
    return strconv.Itoa(index)
}
offsetRange := f.sharedStringItem[index]
```

so a negative index reaches the slice lookup and panics. The in-memory path already rejects this in `xlsxC.getValueFrom`:

```go
if xlsxSI < 0 || xlsxSI >= len(d.SI) {
    return "", newInvalidSharedStringIndex(xlsxSI)
}
```

but the streaming path, taken when the shared strings part exceeds `UnzipXMLSizeLimit`, had no equivalent lower bound. This adds it, so a negative index falls through to the same handling as one that is too large rather than panicking.

I kept the existing `strconv.Itoa(index)` return rather than surfacing `newInvalidSharedStringIndex`, since `getFromStringItem` returns only a string and changing that would ripple through its callers. Happy to take the other approach if you would prefer the two paths to report identically.

The test drives `getFromStringItem` directly rather than building a >16MB workbook, so it stays fast. It panics without the change and passes with it. Full suite is green (579 tests), `go vet` and `gofmt` clean.

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -619,7 +619,11 @@ func (c *xlsxC) getValueFrom(f *File, d *xlsxSST, raw bool) (string, error) {
 		if c.V != "" {
 			xlsxSI, _ := strconv.Atoi(strings.TrimSpace(c.V))
 			if _, ok := f.tempFiles.Load(defaultXMLPathSharedStrings); ok {
-				return f.formattedValue(&xlsxC{S: c.S, V: f.getFromStringItem(xlsxSI)}, raw, CellTypeSharedString)
+				val, err := f.getFromStringItem(xlsxSI)
+				if err != nil {
+					return "", err
+				}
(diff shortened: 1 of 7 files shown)


xuri (approved):

PR #2367, opened 2026-07-30
This closes #GHSA-5h23-36rv-pm65, fix panic on negative style index

arpitjain099:
The `fill`, `border` and `font` conditions in `extractStyleCondFuncs` bounded only the top of the range:

```go
*xf.FillID < len(s.Fills.Fill)
```

Any negative value satisfies that, so a `cellXfs` entry carrying `fillId="-1"`, `borderId="-1"` or `fontId="-1"` passed the guard and `GetStyle` then indexed the fill, border or font table with `-1`. Reading cell styling out of an untrusted workbook panicked with `index out of range [-1]`.

`GetStyle` already checks the low end for the style index itself a few lines above, so this just brings the three table lookups in line with it:

```go
if idx < 0 || s.CellXfs == nil || len(s.CellXfs.Xf) <= idx {
```

Behaviour for valid workbooks is unchanged. A negative id now makes the condition false and the style keeps its default, which is what already happened for an out-of-range positive id.

The regression test writes the styles part directly so the negative attribute goes through the real XML unmarshal rather than being set on the struct, and covers all three ids. Reverting the `styles.go` change makes it panic at `styles.go:1683`, and the full suite passes with the change in.

Same shape as the shared string index fix in #2366, in a different file.

### DIFF
--- a/styles.go
+++ b/styles.go
@@ -1171,17 +1171,17 @@ var (
 		"fill": func(xf xlsxXf, s *xlsxStyleSheet) bool {
 			return (xf.ApplyFill == nil || (xf.ApplyFill != nil && *xf.ApplyFill)) &&
 				xf.FillID != nil && s.Fills != nil &&
-				*xf.FillID >= 0 && *xf.FillID < len(s.Fills.Fill)
+				0 <= *xf.FillID && *xf.FillID < len(s.Fills.Fill)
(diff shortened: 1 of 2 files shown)


xuri (approved):
Thanks for your contribution.

PR #2368, opened 2026-07-31
fix: support matrix arguments in AND formula function

yashisrani:
# PR Details

## Description

Fix the `AND` formula function to properly handle matrix/cell range arguments. Previously, `=AND(A1:B10)` would return a `#VALUE!` error instead of evaluating all cells in the range.

The change replaces the TODO placeholder in `calc.go` with proper matrix iteration using `ToList()`, following the same pattern already implemented in the `OR` function. Each element in the matrix is evaluated:
- Returns `FALSE` immediately if any cell is `0` or `"FALSE"`
- Continues for `TRUE` values and non-zero numbers
- Returns error only for truly invalid string values

Also added 6 new test cases in `calc_test.go` covering:
- All TRUE matrix
- Mixed TRUE/FALSE matrix
- All numbers with one zero
- Mixed types (string, number, bool)
- Invalid string values

## Related Issue

Resolves TODO comment at `calc.go:12092`.

## Motivation and Context

The `AND` function is a core logical function in Excel. Users expect to be able to pass cell ranges like `=AND(A1:B5)` to evaluate multiple conditions at once. Without this fix, any formula using a range argument would fail with `#VALUE!`, forcing users to list each cell individually (e.g., `=AND(A1,A2,A3,B1,B2,B3)`), which is impractical and inconsistent with Excel's behavior.

## How Has This Been Tested

Unit tests and integration tests were run:

Screenshot 2026-07-31 at 5 04 15 PM

Screenshot 2026-07-31 at 5 10 17 PM

## Types of changes

Logical function tests (`TestCalcAND`, `TestCalcOR`, `TestCalcNOT`) all pass without regression.

## Checklist

### DIFF
--- a/calc.go
+++ b/calc.go
@@ -12089,8 +12089,23 @@ func (fn *formulaFuncs) AND(argsList *list.List) formulaArg {
 		case ArgNumber:
 			and = and && token.Number != 0
 		case ArgMatrix:
-			// TODO
-			return newErrorFormulaArg(formulaErrorVALUE, formulaErrorVALUE)
+			for _, item := range token.ToList() {
+				switch item.Type {
+				case ArgUnknown:
+					continue
+				case ArgString:
+					if item.String == "FALSE" {
+						return newStringFormulaArg(item.String)
+					}
+					if item.String != "TRUE" {
+						return newErrorFormulaArg(formulaErrorVALUE, formulaErrorVALUE)
+					}
+				case ArgNumber:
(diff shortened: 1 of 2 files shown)


xuri (approved):
Thanks for your contribution. I've made some changes based on your branch, for simplify unit tests and split code for AND function into the `and` function to reduce cyclomatic complexity.

PR #2369, opened 2026-08-03
Reject a Zip64 uncompressed size that does not fit a signed int64

arpitjain099:
Follows up on the advisory discussion, where you asked whether I wanted to send a patch.

`zip.File.FileInfo().Size()` returns `int64(UncompressedSize64)`, and `UncompressedSize64` comes from the Zip64 extended information extra field in the central directory. Any declared value in `[2^63, 2^64)` turns negative on that conversion, and the guard in `ReadZipReader` is signed arithmetic, so:

- `unzipSize += fileSize` makes the running total negative, and `unzipSize > f.options.UnzipSizeLimit` stays false, so the decompression cap never fires
- the two `UnzipXMLSizeLimit` comparisons are false as well, so a large part is never routed to a temporary file
- `readFile` reaches `make([]byte, 0, negative)`, which panics with `makeslice: cap out of range`

A 159 byte archive is enough to reach it through `OpenFile`, `OpenReader` or `ReadZipReader`, and nothing in the library recovers, so the panic propagates to the caller.

## The change

Compare the declared size against the limit as a `uint64` before it is narrowed. An entry that cannot fit is rejected on its own, so the signed arithmetic below only ever sees a value already within the limit, and `fileSize` can no longer be negative. The `UnzipSizeLimit < 0` clause keeps the conversion honest if someone configures a negative limit.

`readFile` also bounds its capacity hint, so the allocation is safe regardless of how that function is reached. It takes the limit as a parameter since it is package level and has no access to `f.options`; there is one caller.

## Testing

`TestZip64UncompressedSizeOverflow` builds the archive in memory rather than committing a fixture. It asserts the declared size really is negative after conversion, then that `OpenReader` returns the unzip size limit error instead of panicking. It also runs a control at `2^63-1`, one below the sign flip, which already returned the limit error before this change. That is what pins the bug to the conversion rather than to the size being large.

Verified fail before and pass after: on master the new test panics with `makeslice: cap out of range`, and with the fix it passes. Full suite is green locally on go1.26.5 darwin/arm64.

### DIFF
--- a/lib.go
+++ b/lib.go
@@ -28,6 +28,18 @@ import (
 	"unicode/utf16"
 )
 
+// checkFileSize checks if the file size and unzip size exceed the limit set in
+// options.
+func (f *File) checkFileSize(fileSize, unzipSize int64) error {
+	if f.options.UnzipSizeLimit < 0 || uint64(f.options.UnzipSizeLimit) < uint64(fileSize) || fileSize < 0 {
+		return newUnzipSizeLimitError(f.options.UnzipSizeLimit)
+	}
+	if unzipSize > f.options.UnzipSizeLimit {
+		return newUnzipSizeLimitError(f.options.UnzipSizeLimit)
+	}
+	return nil
+}
+
 // ReadZipReader extract spreadsheet with given options.
 func (f *File) ReadZipReader(r *zip.Reader) (map[string][]byte, int, error) {
 	var (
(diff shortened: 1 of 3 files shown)


lib.go
```diff
@@ -41,6 +41,13 @@ func (f *File) ReadZipReader(r *zip.Reader) (map[string][]byte, int, error) {
 		unzipSize  int64
 	)
 	for _, v := range r.File {
+		// The declared uncompressed size is a 64-bit field in the Zip64 extra
+		// record, so compare it before it gets narrowed to a signed int64.
+		// A value of 2^63 or above would otherwise turn negative and slip past
+		// the checks below.
+		if f.options.UnzipSizeLimit < 0 || v.UncompressedSize64 > uint64(f.options.UnzipSizeLimit) {
```
xuri: `File.UnzipSizeLimit` is same as `v.FileInfo().Size()`. We need avoid using `UncompressedSize` for files larger than 4 GB, use `UncompressedSize64` instead of it.

xuri (changes requested):

xuri (approved):

arpitjain099:
Thakn you @xuri Yesterday and today have been busy, will look at your feedback items later in the evening.

xuri:
> Thakn you @xuri Yesterday and today have been busy, will look at your feedback items later in the evening.

No worries, I've made some changes based on your branch. Please take a look for my changes. If your have any additional suggestions, please let me know.

PR #2370, opened 2026-08-06
Bound flatCols to MaxColumns when flattening worksheet columns

arpitjain099:
`flatCols` materializes one `xlsxCol` per column in the range a `<col>` element declares. Those `min` and `max` attributes come straight from the worksheet XML and are not validated anywhere on the way in, so the size of that loop is whatever the file says it is.

A worksheet cannot hold more than `MaxColumns` columns, so flattening past that bound cannot produce anything usable. This clamps both loops to `MinColumns..MaxColumns`.

For any range a worksheet can genuinely contain the result is unchanged, since the clamp only takes effect outside the representable range. The four column mutators that call `flatCols` (`setColVisible`, `setColOutlineLevel`, `setColStyle`, `setColWidth`) keep their signatures, so nothing changes for callers.

I went with clamping rather than returning `ErrColumnNumber` to keep this to `flatCols`. Returning an error would mean threading it through those four helpers and their nine call sites across `col.go` and `stream.go`. Happy to redo it that way if you would rather reject such a file outright, and equally happy to move the check to parse time instead.

Verified:

- `go build ./...` clean, `gofmt` clean
- full `go test ./...` passes (71s)
- added `TestFlatColsBounded`, which sets a `<col>` range of `1..2147483647` on a sheet, calls `SetColWidth`, and asserts the flattened set stays within `MaxColumns` while a real column is still flattened and readable

Filed at your request from the report on the advisory.

### DIFF
--- a/col.go
+++ b/col.go
@@ -545,24 +545,8 @@ func (ws *xlsxWorksheet) setColWidth(minVal, maxVal int, width float64) {
 // flatCols provides a method for the column's operation functions to flatten
 // and check the worksheet columns.
 func flatCols(col xlsxCol, cols []xlsxCol, replacer func(fc, c xlsxCol) xlsxCol) []xlsxCol {
-	// The min and max attributes of a col element are read straight from the
-	// worksheet XML and are not otherwise bounded. A worksheet cannot hold more
-	// than MaxColumns columns, so flattening past that is meaningless, while a
-	// crafted file declaring max="2147483647" would otherwise have this function
-	// allocate one xlsxCol per declared column.
-	clamp := func(c xlsxCol) (int, int) {
-		minVal, maxVal := c.Min, c.Max
-		if minVal < MinColumns {
-			minVal = MinColumns
-		}
-		if maxVal > MaxColumns {
-			maxVal = MaxColumns
-		}
-		return minVal, maxVal
-	}
 	var fc []xlsxCol
-	colMin, colMax := clamp(col)
-	for i := colMin; i <= colMax; i++ {
+	for i := max(MinColumns, col.Min); i <= min(MaxColumns, col.Max); i++ {
 		var c xlsxCol
 		_ = deepcopy.Copy(&c, col)
 		c.Min, c.Max = i, i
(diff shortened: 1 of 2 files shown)


col_test.go
```diff
@@ -585,3 +585,23 @@ func TestAutoFitColWidth(t *testing.T) {
...
+	// crafted file can declare a range far beyond the MaxColumns a worksheet can
+	// actually hold, which previously made flatCols allocate one entry per
+	// declared column.
+	f := NewFile()
+	ws, err := f.workSheetReader("Sheet1")
+	assert.NoError(t, err)
+	ws.Cols = &xlsxCols{Col: []xlsxCol{{Min: 1, Max: 2147483647, Width: float64Ptr(9)}}}
+
+	assert.NoError(t, f.SetColWidth("Sheet1", "A", "A", 12))
+	assert.LessOrEqual(t, len(ws.Cols.Col), MaxColumns)
```
xuri: Using `Equal` instead of `LessOrEqual` for accuracy assertion.

col.go
```diff
@@ -545,8 +545,24 @@ func (ws *xlsxWorksheet) setColWidth(minVal, maxVal int, width float64) {
 // flatCols provides a method for the column's operation functions to flatten
 // and check the worksheet columns.
 func flatCols(col xlsxCol, cols []xlsxCol, replacer func(fc, c xlsxCol) xlsxCol) []xlsxCol {
+	// The min and max attributes of a col element are read straight from the
+	// worksheet XML and are not otherwise bounded. A worksheet cannot hold more
+	// than MaxColumns columns, so flattening past that is meaningless, while a
+	// crafted file declaring max="2147483647" would otherwise have this function
+	// allocate one xlsxCol per declared column.
+	clamp := func(c xlsxCol) (int, int) {
```
xuri: Using:

```go
max(MinColumns, col.Min)
min(MaxColumns, col.Max)
```

instead of introduce new `clamp` would be better approach.

xuri (changes requested):

xuri (approved):

PR #2373, opened 2026-08-08
This closes #2372, fix precision loss for time.Duration cell values

### LINKED ISSUE #2372: time.Duration loses precision due to 32-bit float formatting
### Description

`time.Duration` values lose precision because `setCellDuration` formats the Excel serial value using `bitSize=32`:

```go
func setCellDuration(value time.Duration) (t string, v string) {
	v = strconv.FormatFloat(value.Seconds()/86400, 'f', -1, 32)
	return
}
```

The result of `value.Seconds()/86400` is a `float64`. However, passing `bitSize=32` to `FormatFloat` causes the formatted result to be rounded as if the original value had `float32` precision.

```go
package main

import (
	"fmt"
	"time"

	"github.com/xuri/excelize/v2"
)

func main() {
	f := excelize.NewFile()
	defer f.Close()

	value := 100*24*time.Hour + 123456789*time.Nanosecond

	_ = f.SetCellValue("Sheet1", "A1", value)

	raw, _ := f.GetCellValue(
		"Sheet1",
		"A1",
		excelize.Options{RawCellValue: true},
	)

	fmt.Println(raw)
}
```

```text
100
```

The fractional part of the duration is lost.

The duration value should be formatted with `bitSize=64`, since the value being formatted is a `float64`.

### Go version

1.26.0

v2.11.0

### Environment

```shell
There is nothing to specify.
```

### Validations

alliasgher:
#### Description

`setCellDuration` formats the Excel serial number with `bitSize` 32:

```go
v = strconv.FormatFloat(value.Seconds()/86400, 'f', -1, 32)
```

`value.Seconds()` returns a `float64` and the division is done in `float64`, but `bitSize` 32 tells `FormatFloat` to emit the shortest decimal that round-trips as a `float32`, so the extra precision is thrown away on write.

It is the only float write path in the package that does this — `cell.go:278`, `cell.go:603`, `cell.go:646/648`, `lib.go:815`, `numfmt.go:5290/5544/5676`, `stream.go:358/722` and the `calc.go` sites all use 64. (`cell.go:420` takes a caller-supplied `bitSize` from `SetCellFloat`, which is by design.)

Reproduced on master before the change:

| duration | stored | with bitSize 64 |
| --- | --- | --- |
| `100*24*time.Hour + 123456789ns` | `100` | `100.00000142889802` |
| `1m30s` | `0.0010416667` | `0.0010416666666666667` |

For the first one the entire fractional part is gone. Durations whose serial is exactly representable in a few decimal digits (`25h30m`, `3h15m45s`) were already unaffected.

#### Related Issue

This closes #2372

#### Motivation and Context

Excel stores serials as IEEE doubles, so writing the full `float64` expansion is what the format expects. Note this does change the raw value written into the XML for durations that are not exactly representable — that is the point of the fix, but it is an output change worth stating.

`StreamWriter.SetRow` (`stream.go`) also goes through `setCellDuration`, so it is fixed by the same one-line change.

#### How Has This Been Tested

The existing duration table in `cell_test.go` only asserts formatted output (`"21:51:44"` and friends), which passes both before and after — those are whole-second values that the number format rounds anyway, so they could not guard this. Added a raw-value assertion using `Options{RawCellValue: true}` for two durations that are not exactly representable; it fails on master with `expected: "100.00000142889802" / actual: "100"`.

`go test -race -timeout 60m ./...` passes in full (839s — `TestZip64` needs the longer timeout), and `gofmt -s -l` is clean on both changed files.

#### Types of changes

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -285,7 +285,7 @@ func (c *xlsxC) setCellTime(value time.Time, date1904 bool) (isNum bool, err err
 // setCellDuration prepares cell type and value by given Go time.Duration type
 // time duration.
 func setCellDuration(value time.Duration) (t string, v string) {
-	v = strconv.FormatFloat(value.Seconds()/86400, 'f', -1, 32)
+	v = strconv.FormatFloat(value.Seconds()/86400, 'f', -1, 64)
 	return
 }
 
--- a/cell_test.go
+++ b/cell_test.go
@@ -317,8 +317,7 @@ func TestSetCellValue(t *testing.T) {
 		assert.NoError(t, err)
 		assert.Equal(t, expected, val)
 	}
-	// Test set cell value with a time duration whose serial number is not
-	// exactly representable, the stored value should keep float64 precision
+	// Test set cell value with time
 	for val, expected := range map[time.Duration]string{
 		time.Hour*24*100 + 123456789: "100.00000142889802",
 		time.Minute + time.Second*30: "0.0010416666666666667",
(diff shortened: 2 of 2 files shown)


xuri (approved):
Thanks for your contribution! I've made some modification based on your branch.

PR #2375, opened 2026-08-12
Fix panics in GetConditionalFormats on malformed conditional formatting rules

arpitjain099:
Fix for `GHSA-rxcj-4pj5-74gr`. Thanks for asking for the PR.

## What panics

`GetConditionalFormats` reaches three extractors that index `cfRule` sub-elements straight out of the worksheet XML, without checking length and in one case without checking for nil. A workbook whose rule is missing a child that Excel would always write panics the call, so a service that opens an untrusted spreadsheet and reads its conditional formats can be crashed by a small file.

Four sinks, all in `styles.go`:

| where | guard before | what reaches it |
|---|---|---|
| `extractCondFmtCellIs` | none | `cellIs` rule with no `<formula>`, indexing `Formula[0]` on an empty slice |
| `extractCondFmtColorScale` | none | `colorScale` rule with no `<colorScale>` child, so `c.ColorScale` is nil |
| same, three-colour branch | `colors == 3` | three `<color>` elements and one `<cfvo>`, indexing `Cfvo[1]` and `Cfvo[2]` |
| `extractCondFmtDataBar` | `c.DataBar != nil` | empty `<dataBar></dataBar>`, indexing `Cfvo[0]`, `Cfvo[1]` and `Color[0]` |

**The third one is not in the advisory.** It turned up while writing the tests: that branch is gated on the colour count and then indexes the cfvo slice, so the two counts can disagree. Worth fixing in the same pass.

## The change

Four guards, eleven lines in `styles.go`. Each one leaves the surrounding behaviour alone: a rule missing its children now yields the zero value for those fields rather than stopping the caller. No signature changes and no new error returns, so this is not a breaking change for anyone whose files are well formed.

## Tests

Four cases appended to `styles_test.go`. Each builds a minimal workbook in memory with `archive/zip`, opens it with the public `OpenReader` and calls `GetConditionalFormats`, so they drive the same path a caller does rather than reaching into internals.

Verified individually, since a panic takes the test binary down and would otherwise mask the later cases:

```
cellIs_without_formula         panic: index out of range [0] with length 0
colorScale_element_absent      panic: invalid memory address or nil pointer dereference
dataBar_element_empty          panic: index out of range [0] with length 0
three_colors_but_one_cfvo      panic: index out of range [1] with length 1
```

With the change, all four pass and the full suite is green:

```
ok  github.com/xuri/excelize/v2  72.413s
```

`gofmt` and `go vet` are clean.

Happy to split the third fix into its own commit or PR if you would rather keep the advisory scope exact.

### DIFF
--- a/styles.go
+++ b/styles.go
@@ -3000,7 +3000,9 @@ func (f *File) extractCondFmtCellIs(c *xlsxCfRule, extLst *xlsxExtLst) Condition
 		format.MinValue, format.MaxValue = c.Formula[0], c.Formula[1]
 		return format
 	}
-	format.Value = c.Formula[0]
+	if len(c.Formula) > 0 {
+		format.Value = c.Formula[0]
+	}
 	return format
 }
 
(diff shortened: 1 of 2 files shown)


xuri (approved):
Thanks for your contribution. I've made some changes based on your branch for simplify unit tests code.

PR #2376, opened 2026-08-14
Fix panic on a row whose cells are not in ascending column order

arpitjain099:
Fixes the panic reported in GHSA-8mcq-6wmr-jrjv.

`checkRow` reads `lastCol` from the last element of the row and sizes the rebuilt slice from it, then scatters every cell into that slice by the cell's own column reference. Cells are not guaranteed to be stored in ascending column order, so when an earlier cell names a higher column than the final element, the write goes past the end.

A row holding `Z1` followed by `C1` has `colCount` 2 and `lastCol` 3, so the branch is entered, three placeholders are created, and `Z1` is then written to index 25:

```
panic: runtime error: index out of range [25] with length 3
```

This is reachable from the ordinary read APIs, since `checkRow` runs on worksheet load, so opening a crafted file and calling `GetCellValue` is enough.

The change takes the maximum column across the row rather than reading the last element. The existing loop above already tracks a running maximum for filling in missing `r` attributes, but that value is not used here.

Added `TestCheckRowOutOfOrderColumns` to `rows_test.go`. It covers the direct `checkRow` call and a read through `GetCellValue`. It panics with the message above on the current code and passes with the change. `go test ./...` is green.

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -921,11 +921,8 @@ func (ws *xlsxWorksheet) checkRow() error {
 		if colCount == 0 {
 			continue
 		}
-		// check and fill the cell without r attribute in a row element. Cells are
-		// not guaranteed to be in ascending column order, so track the highest
-		// column seen instead of reading it off the last element, which would
-		// leave a cell sorting after that element without a slot below.
-		rCount, lastCol := 0, 0
+		// check and fill the cell without r attribute in a row element.
+		var rCount, lastCol int
 		for idx, cell := range rowData.C {
 			rCount++
 			if cell.R != "" {
--- a/rows.go
+++ b/rows.go
@@ -937,9 +937,19 @@ func (ws *xlsxWorksheet) checkRow() error {
 			}
 			rowData.C[idx].R, _ = CoordinatesToCellName(rCount, rowIdx+1)
 		}
-		lastCol, _, err := CellNameToCoordinates(rowData.C[colCount-1].R)
-		if err != nil {
-			return err
+		// Cells are not guaranteed to be in ascending column order, so the
+		// highest column in the row is not necessarily the last element. Take
+		// the maximum, otherwise a cell that sorts later than the final element
+		// is scattered past the end of the rebuilt slice below.
+		var lastCol int
+		for cellIdx := range rowData.C {
+			colNum, _, err := CellNameToCoordinates(rowData.C[cellIdx].R)
+			if err != nil {
+				return err
+			}
+			if colNum > lastCol {
+				lastCol = colNum
+			}
 		}
 
 		if colCount < lastCol {
--- a/rows.go
+++ b/rows.go
@@ -921,8 +921,11 @@ func (ws *xlsxWorksheet) checkRow() error {
 		if colCount == 0 {
 			continue
 		}
-		// check and fill the cell without r attribute in a row element
-		rCount := 0
+		// check and fill the cell without r attribute in a row element. Cells are
+		// not guaranteed to be in ascending column order, so track the highest
+		// column seen instead of reading it off the last element, which would
+		// leave a cell sorting after that element without a slot below.
+		rCount, lastCol := 0, 0
 		for idx, cell := range rowData.C {
 			rCount++
 			if cell.R != "" {
@@ -933,22 +936,14 @@ func (ws *xlsxWorksheet) checkRow() error {
 				if lastR > rCount {
 					rCount = lastR
 				}
+				if lastR > lastCol {
+					lastCol = lastR
+				}
 				continue
 			}
 			rowData.C[idx].R, _ = CoordinatesToCellName(rCount, rowIdx+1)
-		}
-		// Cells are not guaranteed to be in ascending column order, so the
-		// highest column in the row is not necessarily the last element. Take
-		// the maximum, otherwise a cell that sorts later than the final element
-		// is scattered past the end of the rebuilt slice below.
-		var lastCol int
-		for cellIdx := range rowData.C {
-			colNum, _, err := CellNameToCoordinates(rowData.C[cellIdx].R)
-			if err != nil {
-				return err
-			}
-			if colNum > lastCol {
-				lastCol = colNum
+			if rCount > lastCol {
+				lastCol = rCount
 			}
 		}
 
--- a/rows.go
+++ b/rows.go
@@ -917,34 +917,34 @@ func (ws *xlsxWorksheet) checkRow() error {
 	for rowIdx := range ws.SheetData.Row {
 		rowData := &ws.SheetData.Row[rowIdx]
 
-		colCount := len(rowData.C)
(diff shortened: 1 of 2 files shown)


rows_test.go
```diff
@@ -1267,3 +1267,31 @@ func trimSliceSpace(s []string) []string {
...
+		C: []xlsxC{{R: "Z1", V: "z"}, {R: "C1", V: "c"}},
+	}}
+	assert.NoError(t, ws.checkRow())
+	assert.Len(t, ws.SheetData.Row[0].C, 26)
+	assert.Equal(t, "z", ws.SheetData.Row[0].C[25].V)
+	assert.Equal(t, "c", ws.SheetData.Row[0].C[2].V)
+
+	// Reading such a sheet through the public API must not panic.
+	f := NewFile()
+	f.Sheet.Store("xl/worksheets/sheet1.xml", &xlsxWorksheet{
```
xuri: `f.Sheet` not used for storage deserialized data, need to using XML instead of `xlsxWorksheet`.

```go
f.Pkg.Store("xl/worksheets/sheet1.xml", fmt.Appendf(nil, `<worksheet xmlns="%s"><sheetData><row r="1"><c r="Z1"><v>z</v></c><c r="C1"><v>c</v></c></row></sheetData></worksheet>`, NameSpaceSpreadSheet.Value))
```

rows.go
```diff
@@ -937,9 +937,19 @@ func (ws *xlsxWorksheet) checkRow() error {
...
 			rowData.C[idx].R, _ = CoordinatesToCellName(rCount, rowIdx+1)
 		}
-		lastCol, _, err := CellNameToCoordinates(rowData.C[colCount-1].R)
-		if err != nil {
-			return err
+		// Cells are not guaranteed to be in ascending column order, so the
+		// highest column in the row is not necessarily the last element. Take
+		// the maximum, otherwise a cell that sorts later than the final element
+		// is scattered past the end of the rebuilt slice below.
+		var lastCol int
```
xuri: We need to check performance impact of this changes, need to run [excelize-benchmark](https://github.com/xuri/excelize-benchmark) to compare matrix for this.

rows.go
```diff
@@ -937,9 +937,19 @@ func (ws *xlsxWorksheet) checkRow() error {
...
+		for cellIdx := range rowData.C {
+			colNum, _, err := CellNameToCoordinates(rowData.C[cellIdx].R)
+			if err != nil {
+				return err
```
xuri: After this changes, this error no longer covered, need to check the root cause or add unit tests cover this cases.

xuri (changes requested):

arpitjain099:
Thanks, all three addressed.

**Test setup.** Switched to `f.Pkg.Store` with raw sheet XML as you suggested. It needed `f.Sheet.Delete` and a `f.checked` reset alongside it, otherwise `NewFile`'s already-deserialized worksheet wins and `GetCellValue` never reads the stored bytes. I also had to add `t="str"` to the two cells, since a bare `<v>` is treated as numeric and both values came back empty. Matches the idiom in `TestRows`.

**Performance.** Rather than defend the second pass I removed it. The loop above already walks every cell to fill in missing `r` attributes and already calls `CellNameToCoordinates` on each one, so the maximum can be tracked there. The added work is two integer comparisons per cell and no extra parsing.

In-repo benchmarks, 5 runs each, medians:

```
bench               master ns/op  branch ns/op    delta   allocs
BenchmarkOpenFile        982,035     1,007,110     2.6%   4272 -> 4272
BenchmarkRows            135,946       137,263     1.0%   1419 -> 1419
```

Both deltas sit inside the run-to-run spread I saw (master OpenFile ranged 955k to 991k, branch 978k to 1063k) and allocations are identical. I have not run xuri/excelize-benchmark itself, only the benchmarks in this repo, so if you want the full matrix from that one please say and I will set it up.

**The uncovered error.** That branch was already dead before this PR. The loop above returns for any cell whose `r` fails to parse, and the names it generates for cells without `r` are always valid, so by the time control reached `CellNameToCoordinates(rowData.C[colCount-1].R)` the value could not fail. A coverage profile on master with the full suite confirms it:

```
$ go test -coverprofile=cov.out .
ok  github.com/xuri/excelize/v2  78.500s  coverage: 99.8% of statements
$ grep 'rows.go:94' cov.out
rows.go:941.17,943.4 1 0
```

That `0` is the `return err`. Folding the search into the existing loop removes the call and the dead branch with it, so there is no longer an uncovered statement there rather than a new one.

Full suite passes, `gofmt` and `go vet` clean.

rows_test.go
arpitjain099: Done, switched to storing the raw sheet XML.

Two things it needed beyond the snippet. `NewFile` leaves an already-deserialized worksheet in `f.Sheet`, which wins over `f.Pkg`, so it also needs `f.Sheet.Delete` and a `f.checked` reset before the store or `GetCellValue` never reads the bytes. And the two cells need `t="str"`, since a bare `<v>` is read as numeric and both values came back empty. That matches how `TestRows` sets up its input.

rows.go
arpitjain099: Rather than measure the second pass I removed it. The loop just above already walks every cell to fill in missing `r` attributes and already calls `CellNameToCoordinates` on each one, so the maximum can be tracked there. The added work is now two integer comparisons per cell and no extra parsing.

In-repo benchmarks, 5 runs each, medians:

```
bench               master ns/op  branch ns/op    delta   allocs
BenchmarkOpenFile        982,035     1,007,110     2.6%   4272 -> 4272
BenchmarkRows            135,946       137,263     1.0%   1419 -> 1419
```

Both deltas sit inside the run-to-run spread I saw (master OpenFile ranged 955k to 991k, branch 978k to 1063k), and allocations are identical.

To be clear about what I did not do: those are the benchmarks in this repo, not xuri/excelize-benchmark. Happy to set that one up and post the full matrix if you would rather see it.

rows.go
arpitjain099: That branch was already dead before this PR, so nothing regressed in coverage.

The loop above returns for any cell whose `r` attribute fails to parse, and the names it generates for cells without `r` come from `CoordinatesToCellName`, so they always parse. By the time control reached `CellNameToCoordinates(rowData.C[colCount-1].R)` the value could not fail, which a coverage profile on master confirms:

```
$ go test -coverprofile=cov.out .
ok  github.com/xuri/excelize/v2  78.500s  coverage: 99.8% of statements
$ grep 'rows.go:94' cov.out
rows.go:941.17,943.4 1 0
```

The trailing `0` is the execution count for that `return err`, with the whole suite running.

Folding the search into the loop above removes the call and the dead branch with it, so there is one fewer uncovered statement here rather than a new one. If you would still prefer an explicit test for a malformed `r`, the reachable path is the check inside that loop and I can add one there.

xuri (approved):

PR #2379, opened 2026-08-18
This closes a panic on a worksheet with an empty merged cell reference

arpitjain099:
Fix for the merged-cell panic reported privately.

`mergeCellsParser` only populates the cached rectangle when `ref != ""`, so a `<mergeCell ref=""/>` leaves `rect` nil, and the `cellInRange` call immediately after is unconditional and indexes it at four positions.

This skips the entry when the rectangle is empty. `checkCellInRangeRef` handled the same shape cleanly through its `len(rng) != 2` guard before a34c81e replaced it with the cached-rect path.

`TestMergeCellsParser` gains the empty-ref case alongside the existing nil-cell one. It panics without the change and passes with it. Full `go test ./...` is green, and `gofmt` and `go vet` are clean.

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -1657,6 +1657,9 @@ func (ws *xlsxWorksheet) mergeCellsParser(cell string) (string, error) {
 				_ = sortCoordinates(rect)
 				ws.MergeCells.Cells[i].rect = rect
 			}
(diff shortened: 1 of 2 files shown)


xuri (approved):
LGTM, thanks for your contribution!

PR #2388, opened 2026-08-31
This closes #2387, fix GetPanes not reporting split panes

### LINKED ISSUE #2387: GetPanes does not report split panes, so GetPanes/SetPanes drops the split
### Description

`GetPanes` never reports `Split`. `getPanes` derives `Freeze` from `state == "frozen"` but leaves `Split` at its zero value for every sheet, including sheets that genuinely carry a split pane.

Because `SetPanes` writes no `<pane>` element when both `Freeze` and `Split` are false, `GetPanes`/`SetPanes` does not round-trip: re-applying the panes excelize itself just reported silently removes the split. Any code that reads the current panes, adjusts one field (a selection, say) and writes them back loses it.

```go
package main

import (
	"fmt"

	"github.com/xuri/excelize/v2"
)

func main() {
	f := excelize.NewFile()
	defer f.Close()

	if err := f.SetPanes("Sheet1", &excelize.Panes{
		Split: true, XSplit: 3270, YSplit: 1800,
		TopLeftCell: "N57", ActivePane: "bottomLeft",
	}); err != nil {
		fmt.Println(err)
		return
	}

	panes, err := f.GetPanes("Sheet1")
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Printf("GetPanes   -> Split=%v XSplit=%d YSplit=%d\n", panes.Split, panes.XSplit, panes.YSplit)

	// Re-apply exactly what excelize just reported.
	if err := f.SetPanes("Sheet1", &panes); err != nil {
		fmt.Println(err)
		return
	}
	again, _ := f.GetPanes("Sheet1")
	fmt.Printf("round trip -> Split=%v XSplit=%d YSplit=%d\n", again.Split, again.XSplit, again.YSplit)
}
```

```
GetPanes   -> Split=false XSplit=3270 YSplit=1800
round trip -> Split=false XSplit=0 YSplit=0
```

`Split` is reported as `false` for a sheet that was just given a split pane, and the round trip drops the pane element entirely.

```
GetPanes   -> Split=true XSplit=3270 YSplit=1800
round trip -> Split=true XSplit=3270 YSplit=1800
```

### Additional context

`ST_PaneState` has four cases and `getPanes` only handles one:

| state | meaning |
|---|---|
| `frozen` | frozen, no split before freezing |
| `frozenSplit` | frozen, and split before freezing |
| `split` | split, not frozen |
| *(absent)* | what `setPanes` writes for a split |

`frozenSplit` is worth calling out: it is a *frozen* state, so inferring `Split` from a non-zero `XSplit`/`YSplit` alone would report such a sheet as unfrozen and convert it to a plain split on the next write. It also cannot currently be produced by `setPanes`, which only ever writes `frozen`.

Reproduced with excelize v2.11.0 on go1.26.7, darwin/arm64. Happy to send a PR — I have the fix and tests ready.

### Output of `go version`

```
go version go1.26.7 darwin/arm64
```

```
v2.11.0
```

### Environment details (OS, Microsoft Excel&trade; version, physical, etc.)

macOS (darwin/arm64), reproduced with the Go library alone — no Excel required.

saifulferoz:
## Description

`getPanes` only ever assigned `Freeze`, derived from `state == "frozen"`. `Split` was never set, so `GetPanes` reported it as `false` for every sheet — including sheets that genuinely carry a split pane.

Because `SetPanes` writes no `<pane>` element when both `Freeze` and `Split` are false, `GetPanes`/`SetPanes` does not round-trip: re-applying the panes excelize itself just reported removes the split entirely.

`ST_PaneState` has four cases and only one was handled:

| state | meaning |
|---|---|
| `frozen` | frozen, no split before freezing |
| `frozenSplit` | frozen, and split before freezing |
| `split` | split, not frozen |
| *(absent)* | what `setPanes` writes for a split |

`getPanes` now maps all four. `frozenSplit` reports both `Freeze` and `Split`, since it is a *frozen* state — inferring `Split` from a non-zero `XSplit`/`YSplit` alone would have reported such a sheet as unfrozen and converted it to a plain split on the next write. `setPanes` gains the matching direction so the state survives a round trip rather than degrading to `frozen`.

## Related Issue

Closes #2387.

## Motivation and Context

Any code that reads the current panes, adjusts one field and writes them back silently loses the split. I hit this applying a cell selection to a sheet that already had panes: selection is carried inside the pane record, so the read-modify-write is the only way to add one without clobbering what is there.

## How Has This Been Tested

Two new tests, both of which fail without the change (15 failing assertions between them):

- `TestGetPanesPaneState` — every `ST_PaneState` value maps to the right `Freeze`/`Split` pair, and a pane with neither a state nor a split offset reports neither.
- `TestPaneStateRoundTrip` — the pane element and its split offsets survive `GetPanes` → `SetPanes` for `frozen`, `frozenSplit` and `split`.

Full suite passes locally (`go test ./...`, ~55s) on go1.26.7, darwin/arm64. `go vet` and `gofmt` clean.

## Types of changes

Worth flagging on the last box: `GetPanes` now returns `Split=true` where it previously always returned `false`. That is the bug being fixed, but it is a visible behaviour change for anyone who had worked around it. `setPanes` writing `frozenSplit` is likewise new output, reachable only when a caller passes both `Freeze` and `Split` — which previously produced `frozen` and silently dropped the split half.

## Checklist

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -860,6 +860,10 @@ func (ws *xlsxWorksheet) setPanes(panes *Panes) error {
 	}
 	if panes.Freeze {
 		p.State = "frozen"
+		if panes.Split {
+			// ST_PaneState: panes frozen after having been split.
+			p.State = "frozenSplit"
+		}
 	}
 	if ws.SheetViews == nil {
 		ws.SheetViews = &xlsxSheetViews{SheetView: []xlsxSheetView{{}}}
@@ -1031,8 +1035,20 @@ func (ws *xlsxWorksheet) getPanes() Panes {
 		return panes
 	}
 	panes.ActivePane = sw.Pane.ActivePane
-	if sw.Pane.State == "frozen" {
+	// ST_PaneState: "frozen" and "frozenSplit" are both frozen; "split" is
+	// split. A pane element with no state attribute is what SetPanes writes
+	// for a split, so a split offset without a state is treated as one too.
+	// Split was previously never assigned, so GetPanes reported it as false
+	// for every sheet and SetPanes(GetPanes(sheet)) silently dropped a split.
+	switch sw.Pane.State {
+	case "frozen":
 		panes.Freeze = true
+	case "frozenSplit":
+		panes.Freeze, panes.Split = true, true
+	case "split":
+		panes.Split = true
+	default:
+		panes.Split = sw.Pane.XSplit > 0 || sw.Pane.YSplit > 0
 	}
 	panes.TopLeftCell = sw.Pane.TopLeftCell
 	panes.XSplit = int(sw.Pane.XSplit)
--- a/sheet.go
+++ b/sheet.go
@@ -861,7 +861,6 @@ func (ws *xlsxWorksheet) setPanes(panes *Panes) error {
 	if panes.Freeze {
 		p.State = "frozen"
 		if panes.Split {
-			// ST_PaneState: panes frozen after having been split.
 			p.State = "frozenSplit"
 		}
 	}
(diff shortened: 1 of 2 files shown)


xuri (approved):
Cool! Thanks for your contribution, I've made some changes based on your code branch.

PR #2389, opened 2026-09-05
Limit the spin count when deriving the agile decryption key

arpitjain099:
As discussed in the advisory, this bounds the key derivation loop instead of letting the file pick the iteration count.

`convertPasswdToKey` now reads `SpinCount` once and returns `ErrMaxSpinCount` if it is negative or above `maxSpinCount`, which I set to 1e7 next to the existing `sheetProtectionSpinCount` and `workbookProtectionSpinCount`. Excel and LibreOffice both write 100000, so that leaves two orders of magnitude of headroom over anything a real producer emits.

The test takes the real `encryptSHA1.xlsx`, rewrites the spin count in its EncryptionInfo stream and calls `agileDecrypt`, so it goes through the same path `OpenFile` uses rather than calling the helper directly. It also covers the negative case. With the check removed the test does not fail so much as hang, and `go test` kills it at the timeout; with the check in place it finishes in about half a second.

Full suite passes locally on go1.27.1.

Happy to change the constant or move the check to parse time if you would rather reject it when the XML is read.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -48,6 +48,7 @@ var (
 	packageOffset               = 8 // First 8 bytes are the size of the stream
 	sheetProtectionSpinCount    = 1e5
 	workbookProtectionSpinCount = 1e5
+	maxSpinCount                = 1e7
 )
 
 // Encryption specifies the encryption structure, streams, and storages are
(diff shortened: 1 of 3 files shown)


xuri (approved):
Cool! Thanks for your contribution.

PR #2390, opened 2026-09-07
Fix Unicode handling and count overflow in LEFT, LEFTB, RIGHT, and RIGHTB

arpitjain099:
The guard on line 14236 measures the text in UTF-16 code units while the RIGHT branch below it slices `[]rune(text)` by rune index. For text holding supplementary-plane characters the two measures differ, so a `numChars` between them passes the guard and produces a negative index.

Rather than change the guard, this makes the RIGHT branch measure the same way the LEFT branch already does. `takeRightUTF16Units` is the counterpart of `truncateUTF16Units`: it returns the trailing part of the string that fits in the requested number of UTF-16 units, and like its counterpart it never splits a surrogate pair, so `RIGHT("𝟙𝟙",3)` gives one character rather than half of one.

Test cases cover both sides of the boundary and the mixed case, `RIGHT("𝟙𝟙",n)` for n from 1 to 5 and `RIGHT("a𝟙",n)` for 2 and 3. I checked they fail before the change rather than assuming it: with the old line restored the suite panics with `slice bounds out of range [-1:]`, and with the change it passes.

Full suite passes locally on go1.27.1.

RIGHTB is untouched. It reaches the same function but takes the byte branch, which is internally consistent, and I probed MID and MIDB across start positions 1 to 5 with no panic.


calc_test.go
```diff
@@ -1898,6 +1898,16 @@ func TestCalcCellValue(t *testing.T) {
 		"RIGHT(\"オリジナルテキスト\",4)":      "テキスト",
 		"RIGHT(\"オリジナルテキスト\",7)":      "ジナルテキスト",
 		"RIGHT(\"オリジナルテキスト\",20)":     "オリジナルテキスト",
+		// RIGHT with supplementary-plane characters, which count as 2 UTF-16
+		// code units each, so a count between the rune count and the UTF-16
+		// count must not walk off the front of the string
+		"RIGHT(\"\U0001D7D9\U0001D7D9\",1)": "",
+		"RIGHT(\"\U0001D7D9\U0001D7D9\",2)": "\U0001D7D9",
```
xuri: The expected value should be `\U0001D7D9\U0001D7D9` if `number_bytes` argument value greater or equal than 2 for this case. The calculate result with `\U0001D7D9\U0001D7D9`of LEFT, LEFTB, RIGHTB formula functions also incorrect.

xuri (changes requested):

xuri (approved):
Thanks for your pull request. I've made some changes based on your code branch.

arpitjain099:
@xuri Thank you, I just got back online. Thanks

PR #2392, opened 2026-09-07
This closes #2391, preserve empty cells in IF functions

### LINKED ISSUE #2391: CalcCellValue returns #VALUE! for arithmetic on an empty cell reference returned by IF
### Description

When IF returns a reference to an empty cell, the result is converted into an empty string rather than kept as a blank. Any arithmetic performed on that result then fails with #VALUE!, while Excel evaluates a blank as 0 in a value context.

I'm opening the issue to link it to the PR i'm sending soon.

Set a cell value to  =IF(FALSE,1,B1)+1 with B1 empty. 
This is valid in excel and evaluates to 1 but breaks in excelize because is considered as =""+1.

```
import (
      "fmt"

      "github.com/xuri/excelize/v2"
)

func main() {
      f := excelize.NewFile()
      // B1 is left empty
      if err := f.SetCellFormula("Sheet1", "A1", "=IF(FALSE,1,B1)+1"); err != nil {
              fmt.Println(err)
              return
      }
      result, err := f.CalcCellValue("Sheet1", "A1")
      fmt.Printf("result: %q, err: %v\n", result, err)
}
```

result: "#VALUE!", err: #VALUE!

result: "1", err: <nil>

### Go version

go version go1.27.0-X:nodwarf5 linux/amd64

2badfcd5841d0d88b0ffea119f5a2f3c53b4910b

### Environment

```shell
Arch Linux
```

### Validations

Aprax14:
# PR Details

Make IF function preserve empty cells as empty and not cast them as empty strings.

## Description

This change adds formulaArg.ToResult, which returns an argument as the result of a formula function while preserving ArgEmpty instead of flattening it into a string. The coercion is then applied by the consumer of the result. 

## Related Issue

https://github.com/qax-os/excelize/issues/2391

## Motivation and Context

=IF(FALSE,1,B1)+1 with B1 empty is valid in excel and evaluates to 1. This change makes excelize consistent with excel behaviour.

## How Has This Been Tested

Added a unit test

## Types of changes

## Checklist

### DIFF
--- a/calc.go
+++ b/calc.go
@@ -329,19 +329,6 @@ func (fa formulaArg) Value() (value string) {
 	return
 }
 
-// ToResult returns a formula argument as the result of a formula function.
-// A reference to a blank cell is preserved as blank so that the consumer
-// of the result applies the required coercion
-func (fa formulaArg) ToResult() formulaArg {
-	switch fa.Type {
-	case ArgNumber:
-		return fa.ToNumber()
-	case ArgEmpty:
-		return fa
-	}
-	return newStringFormulaArg(fa.Value())
-}
-
 // ToNumber returns a formula argument with number data type.
 func (fa formulaArg) ToNumber() formulaArg {
 	var n float64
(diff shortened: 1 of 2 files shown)


calc_test.go
```diff
@@ -4880,6 +4880,24 @@ func TestCalcAND(t *testing.T) {
 	assert.Equal(t, newBoolFormulaArg(true), fn.AND(argsList))
 }
 
+func TestCalcEmptyCellRef(t *testing.T) {
+	f := NewFile()
+	assert.NoError(t, f.SetCellValue("Sheet1", "A1", 1))
+
+	formulaList := map[string]string{
+		"=A1-B1":                "1",
```
xuri: This change only affects the result of the `=A1-IF(FALSE,1,B1)` formula calculation, so we can remove the other cases and move this case into the `IF` test block.

calc.go
```diff
@@ -329,6 +329,19 @@ func (fa formulaArg) Value() (value string) {
 	return
 }
 
+// ToResult returns a formula argument as the result of a formula function.
+// A reference to a blank cell is preserved as blank so that the consumer
+// of the result applies the required coercion
+func (fa formulaArg) ToResult() formulaArg {
```
xuri: It would be better to handle `ArgEmpty` type tokens rather than introduce a `ToResult` function.

xuri (changes requested):

xuri (approved):

PR #2394, opened 2026-09-09
fix: bound ColumnNameToNumber accumulation inside the digit loop

lmonstergg:
Fixes the column-side counterpart of the worksheet normalization panic tracked in draft security advisory GHSA-c85p-xxjj-2r75.

## Description

`ColumnNameToNumber` (lib.go) accumulated the bijective base-26 value of the column name into an int64 and only rejected `col > MaxColumns` after the loop. A crafted cell reference such as `r="VGWQHXLSDVIKWV1"` (column value 3*2^64) wraps the accumulator to 0 and passes the gate with a nil error, so worksheet normalization (`checkSheet` -> `checkSheetR0`) indexes the row's cell slice with `col-1 = -1` and panics with `runtime error: index out of range [-1]`, escaping the public `OpenReader` / `GetCellValue` API and terminating processes parsing untrusted workbooks.

This PR bounds the accumulated value against `MaxColumns` after every digit. Each digit contributes a positive amount, so the partial sums grow strictly and must cross `MaxColumns` before any int64 wrap can occur — this provably rejects every out-of-domain name, including wraps that would land back inside the valid domain.

## Related Issue

Draft security advisory GHSA-c85p-xxjj-2r75 (reported privately). This is the column-side sibling of the row-index allocation issue fixed for CVE-2026-54063 / GHSA-h69g-9hx6-f3v4, but with a distinct root cause (int64 wrap-around → negative index, not oversized allocation) and a distinct fix.

## Motivation and Context

The row side of the same coordinate gate is enforced (`checkRowNum` bounds `r` before allocation); the column side was not. `err == nil` from `CellNameToCoordinates` must imply an in-domain coordinate for every consumer (`checkSheet`, `lastRowNum`, `checkSheetR0`, and `checkRow` at rows.go:969).

## How Has This Been Tested

- Added `f19_regression_test.go`: builds a workbook with the library itself, swaps the worksheet part for a crafted one containing `<row r="0"><c r="VGWQHXLSDVIKWV1" t="inlineStr">...`, and requires `OpenReader` + `GetCellValue` to complete without panicking. On master this fails with `runtime error: index out of range [-1]`; with this patch it passes.
- `go test -run "Column|CellName|CheckSheet|SheetName" -v .` — all pass.
- A standalone public-API reproducer prints `PANIC_REPRODUCED` on master and `NO_PANIC_BLOCKED` with this patch (provided to the maintainer privately together with the draft advisory).

## Types of changes

## Checklist


f19_regression_test.go
```diff
@@ -0,0 +1,89 @@
...
+	base := NewFile()
+	defer func() { _ = base.Close() }()
+	buf, err := base.WriteToBuffer()
+	if err != nil {
+		t.Fatalf("build base workbook: %v", err)
+	}
+
+	craftedSheet := strings.Replace(templateSheet,
+		"<sheetData/>",
+		`<sheetData><row r="0"><c r="VGWQHXLSDVIKWV1" t="inlineStr"><is><t>pwn</t></is></c></row></sheetData>`,
```
xuri: Define invalid column name in the [`invalidColumns`](https://github.com/xuri/excelize/blob/c6971ba9668c15292de4be82233abcf943d73f80/lib_test.go#L36) instead of add new test file to keep code clear.

f19_regression_test.go
```diff
@@ -0,0 +1,89 @@
...
+			t.Fatalf("worksheet normalization panicked on attacker-controlled cell reference: %v", r)
+		}
+	}()
+	_, _ = f.GetCellValue("Sheet1", "A1")
+
+	// The coordinate gate itself must reject the overflowing name in every
+	// case: the in-loop bound check catches the overflow before the int64
+	// accumulator can wrap (including wraps that would land back inside the
+	// valid domain).
+	if col, err := ColumnNameToNumber("VGWQHXLSDVIKWV"); err == nil {
```
xuri: Use the column name `CRPXNLSKVLJFHH`, which corresponds to `math.MaxInt64 + 1`, instead of `VGWQHXLSDVIKWV`, which corresponds to `3 * (math.MaxUint64 + 1)`. That's the minimum value that can trigger an overflow on a 64-bit system.

xuri (changes requested):
Good catch! I've left some comments.

lib.go
```diff
@@ -228,11 +228,11 @@ func ColumnNameToNumber(name string) (int, error) {
 		} else {
 			return -1, newInvalidColumnNameError(name)
 		}
+		if col > MaxColumns {
```
xuri, line 222: We need also adding guard for `JoinCellName` function, and make sure `TestJoinCellName_Error` test passed.

xuri (changes requested):

xuri (approved):

PR #2396, opened 2026-09-09
crypt: bound extractPart allocation by declared CFB stream size

lmonstergg:
Fixes GHSA-x2q3-8cjh-766f

## Summary
extractPart allocates make([]byte, entry.Size) directly from the raw CFB directory-entry streamSize with no clamping (mscfb only detects short sector chains after the allocation). A version-4 entry declaring streamSize 0xFFFFFFFFFFFFFFFF yields entry.Size == -1 → guaranteed makeslice panic escaping Decrypt; any large declared size forces an unbounded allocation (4 GiB from a ~12 KB file).

## Fix
Reject negative sizes and sizes above a 1 GiB policy cap with ErrWorkbookFileFormat before allocating. Regression test included: builds the malformed container in memory and drives the public Decrypt API — panic before the fix, clean error after.

## Verification
Reproduced on master; blocked by this change. The Encrypt()-based legitimate-input regression guard passes.


crypt.go
```diff
@@ -190,17 +190,28 @@ func Encrypt(raw []byte, opts *Options) ([]byte, error) {
 }
 
 // extractPart extract data from storage by specified part name.
+// maxEncryptedPartSize bounds the buffer extractPart allocates from a CFB
+// directory entry's declared stream size (1 GiB is far above any real
+// encrypted workbook part).
+const maxEncryptedPartSize = 1 << 30
```
xuri: According to [\[MS-CFB\]: Compound File Binary File Format](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cfb/53989ce4-7b05-4f8d-829b-d08d6148375b), the compound file binary file format versions 3 and 4 specify different maximum stream sizes. So, we shouldn't define size limit as a fixed value.

dig_s07_f1_poc_test.go
```diff
@@ -0,0 +1,101 @@
+// dig_s07_f1_poc_test.go — regression test for the extractPart allocation
```
xuri: Please updating unit test case in the crypt_test.go, and remove this file.

xuri (changes requested):

xuri (approved):

PR #2397, opened 2026-09-09
calc: evaluate ANCHORARRAY spill cells in the current calculation con…

lmonstergg:
Fixes GHSA-wp2g-vpjj-g53r

## Summary
ANCHORARRAY evaluated each spill-range cell through the exported CalcCellValue, which builds a fresh calcContext per cell (fresh entry marker, fresh iterations budget). Two ordinary, Excel-legal dynamic array formulas referencing each other via ANCHORARRAY therefore recurse unboundedly and abort the process with a fatal, unrecoverable stack overflow when a service evaluates an attacker-supplied workbook.

## Fix
Route spill-cell evaluation through cellResolver on the current calculation context so the entry check and iterations gate apply. An ArgEmpty → "" shim preserves the previous ToNumber behavior for empty spill cells; a fresh-context fallback covers the legacy nil-ctx test paths. Regression test included: two mutually-referencing ANCHORARRAY formulas now terminate normally instead of aborting the process.

## Verification
The mutual-cycle PoC aborts with "fatal error: stack overflow" on master; with this change it terminates normally, and the existing CalcCellValue/cache tests pass unchanged.


xuri (changes requested):

audit_s58_anchor_test.go
```diff
@@ -0,0 +1,48 @@
...
+// FRESH calcContext with a fresh entry marker and a fresh iterations budget
+// (calc.go:896-900). During a pure cycle no evaluation ever completes, so
+// nothing is ever written to formulaArgCache / calcRawCache either, and each
+// nested call re-enters with a full budget -> unbounded recursion -> fatal
+// "stack overflow" (process crash, unrecoverable in Go).
+//
+// Before the fix the process dies with a runtime fatal error; after the fix
+// the calculation terminates and this test prints
+// S58_ANCHORARRAY_CYCLE_TERMINATED.
+func TestS58AnchorArrayMutualCycleUnboundedRecursion(t *testing.T) {
```
xuri: Update existing test case `TestCalcANCHORARRAY` instead of add new test file would be better approach.

xuri (changes requested):

xuri (approved):

PR #2399, opened 2026-09-12
Return every conditional formatting rule on a repeated range reference

### LINKED ISSUE #2398: GetConditionalFormats drops rules when a worksheet has several conditional formatting blocks with the same range reference
### Description

A worksheet may contain more than one `<conditionalFormatting>` element sharing the same `sqref`. Excel writes this routinely — adding a rule to a range that already carries one produces a second block rather than extending the first. `SetConditionalFormat` does the same on a second call for the same range.

`GetConditionalFormats` collects each block's rules into a `map[string][]ConditionalFormatOptions` keyed by `sqref`, assigning rather than appending:

```go
// styles.go
conditionalFormats[cf.SQRef] = opts
```

Every block but the last is therefore discarded. The rules are present and valid in the file, and Excel honours them, but the API never reports them. There is no error and no indication that anything was omitted, so a caller reading a sheet to inspect or preserve its formatting silently works from an incomplete picture.

1. Create a workbook and add a conditional format rule to `A1:A10`.
2. Add a second rule to the same range in a separate `SetConditionalFormat` call.
3. Read the rules back with `GetConditionalFormats`.

```go
package main

import (
	"fmt"

	"github.com/xuri/excelize/v2"
)

func main() {
	f := excelize.NewFile()
	defer f.Close()

	styleID, err := f.NewConditionalStyle(&excelize.Style{
		Fill: excelize.Fill{Type: "pattern", Pattern: 1, Color: []string{"FFC7CE"}},
	})
	if err != nil {
		fmt.Println(err)
		return
	}
	for _, opts := range [][]excelize.ConditionalFormatOptions{
		{{Type: "cell", Format: &styleID, Criteria: "less than", Value: "5"}},
		{{Type: "cell", Format: &styleID, Criteria: "greater than", Value: "10"}},
	} {
		if err := f.SetConditionalFormat("Sheet1", "A1:A10", opts); err != nil {
			fmt.Println(err)
			return
		}
	}

	formats, err := f.GetConditionalFormats("Sheet1")
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Printf("rules returned for A1:A10: %d\n", len(formats["A1:A10"]))
}
```

```
rules returned for A1:A10: 1
```

Only the rule from the last block is returned. The saved `xl/worksheets/sheet1.xml` contains two `<conditionalFormatting>` elements with `sqref="A1:A10"` and two `<cfRule>` elements.

This also occurs when reading files produced by Excel. On a workbook written by Excel containing 5 blocks and 12 rules across 4 distinct ranges — two of the blocks sharing one `sqref` — `GetConditionalFormats` returns 4 keys and 11 rules.

```
rules returned for A1:A10: 2
```

Both rules reported, since both are in the file and both are applied by Excel.

### Go version

1.27.1

ea12859 (master), also reproduces on v2.11.0

### Environment

```shell
macOS 27.0, arm64
go1.27.1 darwin/arm64
```

### Validations

pcvelz:
# PR Details

## Description

`GetConditionalFormats` built its result by assigning each conditional formatting block's rules into a map keyed by the block's range reference:

```go
conditionalFormats[cf.SQRef] = opts
```

A worksheet may hold several `<conditionalFormatting>` elements sharing one `sqref`, so every block but the last was dropped from the returned map.

This changes the assignment to an append, which is how the x14 extension list rules are already collected further down the same function.

## Related Issue

Closes #2398

## Motivation and Context

Excel writes a separate `<conditionalFormatting>` block whenever a rule is added to a range that already carries one, and `SetConditionalFormat` does the same on a second call for the same range. Both are ordinary, so the rules lost this way are ordinary too.

The loss is silent: no error is returned and the omitted rules are valid and honoured by Excel. A caller reading a sheet in order to inspect or preserve its conditional formatting has no way to tell that anything is missing. On a workbook written by Excel with 5 blocks and 12 rules across 4 distinct ranges, two blocks sharing one `sqref`, the function returned 4 keys and 11 rules.

## How Has This Been Tested

Added a case to `TestGetConditionalFormats` that writes two rules to `A1:A10` in separate `SetConditionalFormat` calls and asserts both are returned.

Confirmed the test fails without the change:

```
Error: "[{cell false false 0x400a0a23ac48 greater than 10 ...}]" should have 2 item(s), but has 1
```

and passes with it. The full test suite passes (`go test .`), together with `go build ./...`, `go vet ./...` and `gofmt -l .`.

Tested on go1.27.1 darwin/arm64, macOS 27.0.

## Types of changes

## Checklist

### DIFF
--- a/styles.go
+++ b/styles.go
@@ -3278,10 +3278,6 @@ func (f *File) GetConditionalFormats(sheet string) (map[string][]ConditionalForm
 				opts = append(opts, extractFunc(f, mastCell, cr, ws.ExtLst))
 			}
 		}
-		// A worksheet may hold more than one conditional formatting block for
-		// the same range reference, which Excel writes whenever a rule is
-		// added to a range that already has one. Append so that those rules
-		// are all reported instead of the last block replacing the others.
 		conditionalFormats[cf.SQRef] = append(conditionalFormats[cf.SQRef], opts...)
 	}
 	if ws.ExtLst != nil {
--- a/styles.go
+++ b/styles.go
@@ -3278,7 +3278,11 @@ func (f *File) GetConditionalFormats(sheet string) (map[string][]ConditionalForm
 				opts = append(opts, extractFunc(f, mastCell, cr, ws.ExtLst))
 			}
 		}
-		conditionalFormats[cf.SQRef] = opts
+		// A worksheet may hold more than one conditional formatting block for
+		// the same range reference, which Excel writes whenever a rule is
+		// added to a range that already has one. Append so that those rules
+		// are all reported instead of the last block replacing the others.
+		conditionalFormats[cf.SQRef] = append(conditionalFormats[cf.SQRef], opts...)
 	}
 	if ws.ExtLst != nil {
 		decodeExtLst := new(decodeExtLst)
--- a/styles_test.go
+++ b/styles_test.go
@@ -290,22 +290,24 @@ func TestGetConditionalFormats(t *testing.T) {
 		assert.Equal(t, format, opts[fmt.Sprintf("%s1:%s10", col, col)])
 	}
 	assert.NoError(t, f.SaveAs(filepath.Join("test", "TestGetConditionalFormats.xlsx")))
-	// Test get conditional formats for a range reference carrying more than
-	// one conditional formatting block, which Excel writes whenever a rule is
-	// added to a range that already has one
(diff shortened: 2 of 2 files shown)


xuri (approved):
Thanks for your contribution! I've made some changes based on your code branch.

PR #2401, opened 2026-09-13
Reject a conditional format style index that is not a differential style

### LINKED ISSUE #2400: SetConditionalFormat accepts a style index that is not a differential style, writing a rule with a nonexistent dxfId
### Description

`SetConditionalFormat` writes the `Format` field of `ConditionalFormatOptions` straight into the rule's `dxfId` attribute without checking it. `Format` must be a differential style index, as returned by `NewConditionalStyle`. An index returned by `NewStyle` is a cell style index into a different table, but it is accepted without an error.

The saved workbook then contains a rule whose `dxfId` points at a differential style that does not exist. Excel applies no formatting for that rule, so the mistake is silent in both the library and the resulting file.

1. Create a style with `NewStyle` instead of `NewConditionalStyle`.
2. Use that index as the `Format` of a conditional format rule.
3. Save the workbook and inspect the rule and the differential styles.

```go
package main

import (
	"fmt"

	"github.com/xuri/excelize/v2"
)

func main() {
	f := excelize.NewFile()
	defer f.Close()
	for row := 1; row <= 10; row++ {
		f.SetCellValue("Sheet1", fmt.Sprintf("A%d", row), row)
	}
		Fill: excelize.Fill{Type: "pattern", Pattern: 1, Color: []string{"FFC7CE"}},
	})
	if err != nil {
		fmt.Println(err)
		return
	}
	err = f.SetConditionalFormat("Sheet1", "A1:A10", []excelize.ConditionalFormatOptions{
		{Type: "cell", Format: &styleID, Criteria: "greater than", Value: "5"},
	})
	fmt.Printf("SetConditionalFormat error: %v\n", err)
		fmt.Println(err)
	}
}
```

```
SetConditionalFormat error: <nil>
```

`xl/worksheets/sheet1.xml` contains:

```xml
<cfRule type="cellIs" dxfId="1" priority="1" operator="greaterThan">
```

while `xl/styles.xml` contains:

```xml
<dxfs count="0">
```

The rule references differential style 1, but the workbook defines no differential styles at all.

`SetConditionalFormat` returns an invalid style ID error, the same error `GetConditionalStyle` already returns for an index outside the differential styles. Rule types that never write a `dxfId` (color scales, data bars, icon sets) keep accepting any `Format`, as they do today.

### Go version

1.27.1

4459059 (master), also reproduces on v2.11.0

### Environment

```shell
macOS 27.0, arm64
go1.27.1 darwin/arm64
```

### Validations

pcvelz:
# PR Details

## Description

`SetConditionalFormat` wrote `ConditionalFormatOptions.Format` into the rule's `dxfId` attribute without checking it against the differential styles. This validates the index before any rule is written and returns the same invalid style ID error that `GetConditionalStyle` already returns for such an index.

Color scales, data bars and icon sets never write a `dxfId`, so they are skipped and keep accepting any `Format`, as the existing data bar test relies on.

The `NewConditionalStyle` and `GetConditionalStyle` doc comments now note that the index belongs to the differential styles, not the cell styles `NewStyle` returns.

## Related Issue

Closes #2400

## Motivation and Context

Passing an index from `NewStyle` is an easy mistake, and it was accepted without an error. The saved workbook then held a rule pointing at a differential style that does not exist, with nothing in the library or the file to flag it.

## How Has This Been Tested

Added a `with_style_index_that_is_not_a_differential_style` subtest to `TestSetConditionalFormat`. It asserts an invalid style ID error for an index from `NewStyle` and for a negative index.

The full test suite passes (`go test .`), along with `go build ./...`, `go vet ./...` and `gofmt -l .`.

Tested on go1.27.1 darwin/arm64, macOS 27.0.

## Types of changes

## Checklist

### DIFF
--- a/styles.go
+++ b/styles.go
@@ -1731,7 +1731,8 @@ func (f *File) getStyleID(ss *xlsxStyleSheet, style *Style) (int, error) {
 
 // NewConditionalStyle provides a function to create style for conditional
 // format by given style format. The parameters are the same with the NewStyle
-// function.
+// function. Use the returned index, not one from NewStyle, as the Format of
+// ConditionalFormatOptions.
 func (f *File) NewConditionalStyle(style *Style) (int, error) {
 	f.mu.Lock()
 	s, err := f.stylesReader()
@@ -1772,7 +1773,8 @@ func (f *File) NewConditionalStyle(style *Style) (int, error) {
 }
 
 // GetConditionalStyle returns conditional format style definition by specified
-// style index.
+// style index, as returned by NewConditionalStyle. The GetStyle function does
+// not accept this index.
 func (f *File) GetConditionalStyle(idx int) (*Style, error) {
 	var style *Style
 	f.mu.Lock()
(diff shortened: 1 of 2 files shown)


styles.go
```diff
@@ -2845,6 +2873,9 @@ func (f *File) SetConditionalFormat(sheet, rangeRef string, opts []ConditionalFo
 	if err != nil {
 		return err
 	}
+	if err := f.validateConditionalFormatStyleID(opts); err != nil {
```
xuri: Keep the cyclomatic complexity of the `SetConditionalFormat` function < 16.

styles.go
```diff
@@ -2836,6 +2838,32 @@ func (f *File) SetCellStyle(sheet, topLeftCell, bottomRightCell string, styleID
...
+// generated rule carries no dxfId attribute.
+var condFmtTypesWithoutDxf = []string{"2_color_scale", "3_color_scale", "data_bar", "icon_set"}
+
+// validateConditionalFormatStyleID checks that every style index given in the
+// conditional format options refers to an existing differential style.
+func (f *File) validateConditionalFormatStyleID(opts []ConditionalFormatOptions) error {
+	f.mu.Lock()
+	s, err := f.stylesReader()
+	if err != nil {
+		f.mu.Unlock()
```
xuri: Add unit test code to cover this line.

xuri (changes requested):

styles.go
```diff
@@ -2836,6 +2838,32 @@ func (f *File) SetCellStyle(sheet, topLeftCell, bottomRightCell string, styleID
 // formatting rule when more than one rule is applied to a cell or a range of
 // cells. When this parameter is set then subsequent rules are not evaluated
 // if the current rule is true.
+// condFmtTypesWithoutDxf lists the conditional formatting rule types whose
```
xuri: This interrupts the documentation comment for the `SetConditionalFormat` function. Please declare the variable either before or after the `SetConditionalFormat` function comment. The same issue also appears in Pull Request #2399. Please manually review any code generated by generative AI.

xuri (changes requested):

xuri (approved):

PR #2402, opened 2026-09-15
Fix conditional formatting and data validation range shifted on removing its first row

askalf:
## Summary

- `adjustCellRef` (`adjust.go:243`) shifted the **start** of a range whenever it was `>= num`. On a
  deletion the row/column equal to `num` is the one being removed, so a range whose first row *is*
  the removed row must shrink in place — not move up.
- `RemoveRow("Sheet1", 5)` on a conditional format covering `D5:D8` produced `D4:D7`: the range
  silently moved onto row 4, which was never part of it.
- `adjustCellRef` is the shared adjustment path for conditional formatting (`adjust.go:954`), data
  validations (`adjust.go:989`) and shared-formula `F.Ref` (`adjust.go:296`), for both rows and
  columns, so six feature × direction combinations were affected.
- When the range started at row 1 / column A, base did not merely mis-adjust it —
  `RemoveRow(sheet, 1)` drove the start to 0 and returned `invalid cell reference [1, 0]` to the
  caller, leaving the rule unadjusted. The same predicate fixes that.
- Fix is one predicate: on deletion the start moves only when it is strictly after the removed
  row/column; insertions keep the existing `>= num` behaviour. `adjust.go` +3/-1.
- Regression tests: 19 cases across `TestAdjustConditionalFormats`, `TestAdjustDataValidations` and
  `TestAdjustFormula`. 12 fail on base; 7 are named `_control` and pass on both arms by design.

```console
$ # BASE — adjust.go at master, all tests on the branch applied
$ go test -run 'TestAdjustConditionalFormats|TestAdjustDataValidations|TestAdjustFormula' -v .
--- FAIL: TestAdjustFormula (0.06s)
    --- FAIL: TestAdjustFormula/for_shared_formula_ref_on_first_row_of_range (0.00s)
--- FAIL: TestAdjustConditionalFormats (0.03s)
    --- FAIL: TestAdjustConditionalFormats/for_adjust_conditional_formats_on_first_row_of_range (0.02s)
        --- FAIL: .../remove_first_row_of_range (0.00s)
        --- FAIL: .../remove_first_row_of_multiple_range (0.00s)
        --- FAIL: .../remove_first_column_of_range (0.00s)
        --- FAIL: .../remove_first_row_of_range_at_first_row (0.00s)
        --- FAIL: .../remove_first_column_of_range_at_first_column (0.00s)
        --- FAIL: .../remove_first_row_of_absolute_range (0.00s)
        --- FAIL: .../remove_first_row_of_multiple_range_leading (0.00s)
        --- FAIL: .../remove_first_row_of_adjacent_multiple_range (0.00s)
    --- FAIL: TestAdjustConditionalFormats/for_remove_first_row_of_conditional_formats_range (0.00s)
    --- FAIL: TestAdjustConditionalFormats/for_remove_first_row_of_conditional_formats_range_at_first_row (0.00s)
    --- FAIL: TestAdjustConditionalFormats/for_remove_first_column_of_conditional_formats_range (0.00s)
--- FAIL: TestAdjustDataValidations (0.01s)
    --- FAIL: .../for_remove_first_row_of_data_validations_range (0.00s)
    --- FAIL: .../for_remove_first_column_of_data_validations_range (0.00s)
    --- FAIL: .../for_remove_first_row_of_data_validations_range_at_first_row (0.00s)
FAIL
FAIL	github.com/xuri/excelize/v2	0.121s

$ # HEAD — with the fix, same tests: 42 PASS, 0 FAIL
$ go test -run 'TestAdjustConditionalFormats|TestAdjustDataValidations|TestAdjustFormula' -v .
PASS
ok  	github.com/xuri/excelize/v2	0.136s

$ # HEAD — the whole TestAdjust* family in adjust_test.go: 54 PASS, 0 FAIL
$ go test -run 'TestAdjust' -v .
PASS
ok  	github.com/xuri/excelize/v2	0.209s
```

## Decisions

```diff
--- a/adjust.go
+++ b/adjust.go
@@ -241,7 +241,9 @@ func (f *File) adjustSingleRowFormulas(sheet, sheetN string, r *xlsxRow, num, of
 func (f *File) adjustCellRef(cellRef string, dir adjustDirection, num, offset int) (string, error) {
 	var SQRef []string
 	applyOffset := func(coordinates []int, idx1, idx2, maxVal int) []int {
-		if coordinates[idx1] >= num {
+		// On deleting, the start of the range moves only when it is after the
+		// removed row or column, a range starting on it shrinks in place
+		if start := coordinates[idx1]; (offset < 0 && start > num) || (offset >= 0 && start >= num) {
 			coordinates[idx1] += offset
 		}
 		if coordinates[idx2] >= num {
```

The end-of-range branch is deliberately left alone: for the end, `>= num` is correct in both
directions (deleting a row inside or at the end of a range must pull the end up; inserting must
push it down). This is minimal and local — one predicate, no new function, no new dependency, no
signature change.

Rejected alternatives:
- Guarding at the call sites (`adjustConditionalFormats`, `adjustDataValidations`,
  `adjustFormula`) instead of in `adjustCellRef` — would duplicate the same condition in all three
  callers (confirmed there are exactly three non-test callers) and leave `adjustCellRef` wrong for
  any future caller.
- Reusing `adjustMergeCellsHelper` — it takes a `(p1, p2)` pair and has different deletion
  semantics (deletes single-row merges upstream of the call, normalises reversed pairs), so reusing
  it here would be a behaviour change beyond this bug.
- Fixing only the rows branch — the same defect is demonstrably present for columns
  (`remove_first_column_of_range`: `E1:H1` → `D1:G1` on base), so a rows-only fix would leave half
  the bug.
- Skipping the range entirely when `start == num` — wrong: only a single-row range collapses, and
  that case is already handled earlier in the function before `applyOffset` is reached.

Not run: the fork has no CI configured, so there is no CI signal on this branch beyond the local
`go test`, `gofmt -l` and `go vet` runs quoted above.

AI assistance: this bug was found and the fix and tests were drafted with AI tooling in my
workflow; the tests and checks above were executed as pasted. I'm responsible for the change and
will handle review feedback.


xuri (approved):
Thanks for your pull request. I've made some changes based on your code branch, instead of close this.

PR #2405, opened 2026-09-19
This closes #2404, use explicit defaults for `CalcPropsOptions`

### LINKED ISSUE #2404: SetCalcProps cannot set fullPrecision to false
### Description

There is an Excel file config "Set precision as displayed". By default, this checkbox is NOT ticked in Excel.
In the XML schema, this field is `fullPrecision` which defaults to true (the opposite of the checkbox).
I want to create an Excel workbook with "Set precision as displayed" turned on -> `fullPrecision` false.

```
f := excelize.NewFile()
fullPrecision := false
if err := f.SetCalcProps(&excelize.CalcPropsOptions{FullPrecision: &fullPrecision}); err != nil {
      log.Fatal(err)
}
opts, err := f.GetCalcProps()
if err != nil {
      log.Fatal(err)
}
fmt.Println(opts.FullPrecision) // <nil>, and xl/workbook.xml has no fullPrecision
```

fullPrecision="false" or fullPrecision="0" XML attribute is NOT written. So the Excel workbook has "Set precision as displayed" turned OFF.

fullPrecision="false" or fullPrecision="0" XML attribute written. So the Excel workbook has "Set precision as displayed" turned ON.

### Go version

1.27.1

2.11.0

### Environment

```shell
MacOS darwin/arm64
```

### Validations

zac-scheiwe:
Closes #2404

# PR Details

Motivation: Allow `SetCalcProps` to set `fullPrecision` to false. Fix all instances of this type of bug in `CalcPropsOptions` while we're here.

## Description

The `omitempty` struct tag can't tell explicitly set to false from unset. This PR removes the `omitempty` struct tags and uses pointers and explicit defaults instead. The defaults come from [ECMA-376](https://ecma-international.org/publications-and-standards/standards/ecma-376/).

This is a breaking change because `GetCalcProps` now returns a schema default where it used to return nil, and `SetCalcProps` now omits an attributes it used to write. However, the Excel behaviour is identical.

## Related Issue

[#2404](https://github.com/qax-os/excelize/issues/2404)

## How Has This Been Tested

Bug resolved: I produced an Excel file with "Set precision as displayed" ticked on.
Breaking change: No change in Excel behaviour when using the explicit defaults introduced in this PR.

## Types of changes

## Checklist


xmlWorkbook.go
```diff
@@ -314,19 +314,19 @@ type xlsxDefinedName struct {
...
 // displaying the results as values in the cells that contain the formulas.
 type xlsxCalcPr struct {
-	CalcCompleted         bool    `xml:"calcCompleted,attr,omitempty"`
-	CalcID                int     `xml:"calcId,attr,omitempty"`
-	CalcMode              string  `xml:"calcMode,attr,omitempty"`
-	CalcOnSave            bool    `xml:"calcOnSave,attr,omitempty"`
-	ConcurrentCalc        *bool   `xml:"concurrentCalc,attr"`
-	ConcurrentManualCount int     `xml:"concurrentManualCount,attr,omitempty"`
-	ForceFullCalc         bool    `xml:"forceFullCalc,attr,omitempty"`
-	FullCalcOnLoad        bool    `xml:"fullCalcOnLoad,attr,omitempty"`
```
xuri: Default value of `fullCalcOnLoad` and `iterate` field was false, so we can keep `omitempty` tag.

workbook_test.go
```diff
@@ -36,41 +37,151 @@ func TestWorkbookProps(t *testing.T) {
 	assert.EqualError(t, err, "XML syntax error on line 1: invalid UTF-8")
 }
 
+func effectiveCalcProps() CalcPropsOptions {
```
xuri: Using [`setPtrFieldsVal`](https://github.com/xuri/excelize/blob/0434413565bf2eff25c307e84b2525087065c5e2/lib.go#L974-L992) assigns values from the pointer or no-pointer structs.

xmlWorkbook.go
```diff
@@ -314,19 +314,19 @@ type xlsxDefinedName struct {
...
-	FullPrecision         bool    `xml:"fullPrecision,attr,omitempty"`
-	Iterate               bool    `xml:"iterate,attr,omitempty"`
-	IterateCount          int     `xml:"iterateCount,attr,omitempty"`
-	IterateDelta          float64 `xml:"iterateDelta,attr,omitempty"`
-	RefMode               string  `xml:"refMode,attr,omitempty"`
+	CalcCompleted         *bool    `xml:"calcCompleted,attr"`
```
xuri, line 325: Align fields order with `CT_CalcPr` schema.

xmlWorkbook.go
```diff
@@ -314,19 +314,19 @@ type xlsxDefinedName struct {
...
+	CalcID                *int     `xml:"calcId,attr"`
+	CalcMode              *string  `xml:"calcMode,attr"`
+	CalcOnSave            *bool    `xml:"calcOnSave,attr"`
+	ConcurrentCalc        *bool    `xml:"concurrentCalc,attr"`
+	ConcurrentManualCount *int     `xml:"concurrentManualCount,attr"`
+	ForceFullCalc         *bool    `xml:"forceFullCalc,attr"`
+	FullCalcOnLoad        *bool    `xml:"fullCalcOnLoad,attr"`
```
xuri: Default value of `fullCalcOnLoad` and `iterate` field was false, so we can keep `omitempty` tag.

workbook.go
```diff
@@ -73,43 +157,50 @@ func (f *File) SetCalcProps(opts *CalcPropsOptions) error {
...
-		wb.CalcPr.CalcID = int(*opts.CalcID)
-	}
-	if opts.ConcurrentManualCount != nil {
-		wb.CalcPr.ConcurrentManualCount = int(*opts.ConcurrentManualCount)
-	}
-	if opts.IterateCount != nil {
-		wb.CalcPr.IterateCount = int(*opts.IterateCount)
-	}
-	wb.CalcPr.ConcurrentCalc = opts.ConcurrentCalc
+	cp := wb.CalcPr
```
xuri: Using [`setPtrFields`](https://github.com/xuri/excelize/blob/0434413565bf2eff25c307e84b2525087065c5e2/lib.go#L946-L957) assigns values from the pointer or no-pointer structs.

workbook_test.go
```diff
@@ -36,41 +37,151 @@ func TestWorkbookProps(t *testing.T) {
...
 }
 
+func TestCalcPropsOmitsSpecDefaults(t *testing.T) {
+	for _, tc := range []struct {
+		name    string
+		opts    CalcPropsOptions
+		attr    string
+		written bool
+	}{
+		{"FullPrecision default", CalcPropsOptions{FullPrecision: boolPtr(true)}, "fullPrecision", false},
```
xuri: We expect the behavior to be as follows:

- When the `FullPrecision` property is explicitly set to `true`, the XML should contain `fullPrecision="true"`. When the property is read using the `GetCalcProps` function, `FullPrecision` should be `true`
- When the `FullPrecision` property is explicitly set to `false`, the XML should contain `fullPrecision="false"`. When the property is read using the `GetCalcProps` function, `FullPrecision` should be `false`
- When the user does not set the `FullPrecision` property, the `fullPrecision` attribute should not be written to the XML. When the property is read using the `GetCalcProps` function, `FullPrecision` should be `nil`, indicating that the property was not set. In this case, the default value is `true`

xuri (approved):

xuri (approved):

zac-scheiwe:
Thanks @xuri

PR #2407, opened 2026-09-19
Fix CalcCellValue mis-grouping args of nested calls used as operands

### LINKED ISSUE #2406: CalcCellValue returns wrong result for deeply nested IF used as arithmetic operand
## Description

`CalcCellValue` silently returns a wrong result when a deeply nested `IF` is used as an arithmetic operand inside another `IF`. The same nested `IF` evaluates correctly standalone, and a shallower nesting evaluates correctly as an operand — so this looks like a parser/evaluator bug with nested calls as operands (same class as #2084), not a documented limitation.

## Steps to reproduce

```go
package main

import (
	"fmt"

	"github.com/xuri/excelize/v2"
)

func main() {
	f := excelize.NewFile()
	defer f.Close()
	for c, v := range map[string]float64{"B6": 30, "D6": 100, "E6": 200, "B14": 0.3, "B15": 0.5, "B16": 0.7, "B17": 0.9} {
		if err := f.SetCellValue("Sheet1", c, v); err != nil {
			panic(err)
		}
	}
	check := func(cell, formula, want string) {
		if err := f.SetCellFormula("Sheet1", cell, formula); err != nil {
			panic(err)
		}
		got, err := f.CalcCellValue("Sheet1", cell)
		fmt.Printf("got=%-8q want=%-8q err=%v   %s\n", got, want, err, formula)
	}
	// baseline: standalone 4-level nested IF -> correct
	check("Z8", `IF($B6<=30,$B$14,IF($B6<=90,$B$15,IF($B6<=180,$B$16,$B$17)))`, "0.3")
	// baseline: 2-level nested IF as operand -> correct
	check("Z6", `IF($B6>=1,-(D6+E6)*IF($B6<=30,$B$14,0.99),0)`, "-90")
	// BUG: 4-level nested IF as operand -> wrong, silent
	check("F6", `IF($B6>=1,-(D6+E6)*IF($B6<=30,$B$14,IF($B6<=90,$B$15,IF($B6<=180,$B$16,$B$17))),0)`, "-90")
}
```

Output on v2.11.0:

```
got="0.3"      want="0.3"      err=<nil>   IF($B6<=30,...)
got="-90"      want="-90"      err=<nil>   IF($B6>=1,-(D6+E6)*IF($B6<=30,$B$14,0.99),0)
got="0.15"     want="-90"      err=<nil>   IF($B6>=1,-(D6+E6)*IF($B6<=30,$B$14,IF(...)),0)
```

With `B6=30`, the inner `IF` picks `0.3`, so `-(100+200)*0.3 = -90`.
Excel and LibreOffice both evaluate it to `-90`. No error is returned —
the wrong value is silent, which is the dangerous part.

## Environment

- excelize v2.11.0 (`go get` fresh)
- Go 1.27.1 linux/amd64

lrweck:
Fixes #2406.

When a function call nested 2+ deep is used as an arithmetic operand inside another call, `prepareEvalInfixExp` pushed a pending operand of the enclosing expression into the closing call's argument list. Depending on shape this either errored spuriously (`IF accepts at most 3 arguments`) or silently mis-evaluated: `IF(B6>=1,-(D6+E6)*IF(...),0)` with B6=30 returned `0.15` instead of `-90` (Excel and LibreOffice agree on `-90`).

The fix tracks `opfdStack.Len()` at each function start (watermark) and only flushes operands produced inside the call itself; a pending outer operand stays on `opfd`. Push/pop sites are 1:1 with the existing stacks.

Validation:
- New regression test `TestCalcCellValueNestedOperandAsArgument` (fails before, passes after).
- Full package suite green.


xuri (approved):
Thanks for your pull request. The similar issue has been fixed by #2348. I've made some changes based on your code branch instead of closes this.

PR #2410, opened 2026-09-20
This closes #2409, fix panic on unbalanced closing parentheses in formula

### LINKED ISSUE #2409: CalcCellValue panics on a formula with unbalanced closing parentheses inside a function argument
### Description

`CalcCellValue` panics with `interface conversion: interface {} is nil, not efp.Token` instead of returning an error when a formula closes more parentheses than it opened inside a function argument list.

In `parseToken` (calc.go:1571) the closing parenthesis branch peeks the operator stack without checking whether it is empty:

```go
if isEndParenthesesToken(token) { // )
	for !isBeginParenthesesToken(optStack.Peek().(efp.Token)) { // != (
```

`Stack.Peek` returns `nil` on an empty stack (lib.go:1042), so the type assertion panics. Inside a function argument list the operator stack starts empty, which is why the top level form `1)` is fine but `0(((;)))` is not.

This is reachable from an opened workbook, not just from `SetCellFormula`, so a malformed file crashes the calling process rather than returning an error.

Note that the same formula without the leading token, `(((;)))`, already returns `formula not valid`, so the panicking case looks like a missing guard rather than intended behaviour.

I found this by fuzzing `CalcCellValue`.

```go
package main

import (
	"fmt"

	"github.com/xuri/excelize/v2"
)

func main() {
	f := excelize.NewFile()
	if err := f.SetCellFormula("Sheet1", "A1", "0(((;)))"); err != nil {
		panic(err)
	}
	fmt.Println(f.CalcCellValue("Sheet1", "A1"))
}
```

The same happens after a round trip through a file, which is the case that matters for untrusted input:

```go
buf, _ := src.WriteToBuffer()
reopened, _ := excelize.OpenReader(bytes.NewReader(buf.Bytes()))
reopened.CalcCellValue("Sheet1", "A1") // panics
```

Depth 1 and 2 are fine, depth 3 and above panic:

```
0(;)      -> #VALUE!, not support 0 function
0((;))    -> #VALUE!, not support 0 function
0(((;)))  -> panic
```

```
panic: interface conversion: interface {} is nil, not efp.Token

github.com/xuri/excelize/v2.(*File).parseToken(...)
	calc.go:1571
github.com/xuri/excelize/v2.(*File).evalInfixExp(...)
	calc.go:1111
github.com/xuri/excelize/v2.(*File).calcCellValue(...)
	calc.go:961
github.com/xuri/excelize/v2.(*File).CalcCellValue(...)
	calc.go:896
```

An error rather than a panic. `ErrInvalidFormula` looks like the right one, since it is what `(((;)))` already returns and what the existing `SUM((` test case expects.

### Go version

1.26.2

master at 57de107, the same code path is present in v2.11.0

### Environment

```shell
macOS 15.1, darwin/arm64
```

### Validations

- [x] The provided reproduction is a minimal reproducible example.

n0liu:
Closes #2409.

The closing parenthesis branch of `parseToken` peeked the operator stack without checking whether it was empty:

```go
if isEndParenthesesToken(token) { // )
	for !isBeginParenthesesToken(optStack.Peek().(efp.Token)) { // != (
```

`Stack.Peek` returns `nil` on an empty stack, so the `efp.Token` type assertion panicked. Inside a function argument list the operator stack starts empty, which is why `1)` at the top level is fine but `0(((;)))` panics. Since formulas also come from the workbook, a malformed file crashed the calling process rather than returning an error.

The fix guards the loop with `Stack.Empty`, the same way the postfix operator branch a few lines below already does, and returns `ErrInvalidFormula` once the stack runs out. I picked that error because the same formula without the leading token, `(((;)))`, already returns it, and because the existing `SUM((` test case expects it, so the two forms now behave alike.

Verified locally: the two new cases panic before the change and pass after it, the full suite is green with no other changes, and `-race`, `gofmt` and `go vet` are clean.

One thing worth flagging: I found this by fuzzing `CalcCellValue`, and the same nil-peek pattern exists elsewhere in `calc.go`, for example at line 1076 where `opftStack.Peek().(efp.Token)` is compared against `opfStack.Peek().(efp.Token)`. The input `0((;(0()))A` reaches it. I kept this PR to the one site so the change stays small and easy to review. Happy to open a separate issue for the others, or to look at a broader guard if you would rather handle them together.


calc_test.go
```diff
@@ -2404,6 +2404,8 @@ func TestCalcCellValue(t *testing.T) {
 		"1*\"\"":      {"#VALUE!", "#VALUE!"},
 		"\"\"^1":      {"#VALUE!", "#VALUE!"},
 		"1/\"\"":      {"#VALUE!", "#VALUE!"},
+		"0(((;)))":    {"", ErrInvalidFormula.Error()},
+		"0((((;))))":  {"", ErrInvalidFormula.Error()},
```
xuri: Invalid formula `0((;(0()))A` still cause panic.

xuri (changes requested):

xuri (approved):
The root cause is in efp library, it unconditionally treats `;` as an instruction to pop the top of the stack and start an ARRAYROW, regardless of whether it appears inside `{}`. As a result, the token stream becomes unbalanced. I'll update efp at first, and then update this pull request.

xuri (approved):

xuri (approved):

